PR #44, opened 2018-10-03
docs(readme): fix directory name

visr:


langevin-usgs:
Thanks.  And nice job on the commit message format!

PR #45, opened 2018-10-04
test(test000_setup): fallback for when os.path.relpath throws

visr:
Happens on Windows only. It is only used for logging, so no functional change.

Example of issue:
```python
>>> import os
>>> os.getcwd()[:2]
'D:'
>>> os.path.relpath("C:\\tmp")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\bin\Miniconda3\lib\ntpath.py", line 586, in relpath
    path_drive, start_drive))
ValueError: path is on mount 'C:', start on mount 'D:'
>>> os.path.relpath("D:\\tmp")
'..\\..\\tmp'
```

And the new behavior:
```python
>>> relpath_fallback("D:\\tmp")
'..\\..\\tmp'
>>> relpath_fallback("C:\\tmp")
'C:\\tmp'
```

### DIFF
--- a/autotest/test000_setup.py
+++ b/autotest/test000_setup.py
@@ -26,6 +26,14 @@ download_version = '2.0'
 mfexe_pth = 'temp/mfexes'
 
 
+def relpath_fallback(pth):
+    try:
+        # throws ValueError on Windows if pth is on a different drive
+        return os.path.relpath(pth)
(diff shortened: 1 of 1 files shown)


visr:
Rebased. Test are passing now.

PR #82, opened 2019-01-30
Fix typos

mwtoews:
This is not comprehensive, just aligning with a few typos found in https://github.com/modflowpy/flopy/pull/455


langevin-usgs:
@mwtoews not sure we want to version typos.txt.  Can you remove that?

mwtoews:
> @mwtoews not sure we want to version typos.txt. Can you remove that?

Whoops! It's gone now.

PR #85, opened 2019-02-08
fix(src_cc): Add SFR convergence check

jdhughes-usgs:
Add `OUTER_RCLOSEBND` variable to IMS that is used when performing final
convergence checks on model packages that solve a separate equation
not solved by the IMS linear solver. Add stage and residual convergence
checks to the SFR package to make sure that stage and upstream flow
changes between successive outer iterations are less than `OUTER_HCLOSE`
and `OUTER_RCLOSEBND`, respectively.

Modify the final convergence check for the LAK package to use
`OUTER_HCLOSE` when evaluating lake stage changes between successive
outer iterations. Modify the final convergence check for the UZF package
to use `OUTER_RCLOSEBND` when evaluating rejected infiltration,
groundwater recharge, and groundwater seepage changes between successive
outer iterations.


src/Solution/NumericalSolution.f90
```diff
@@ -794,7 +783,42 @@ subroutine sln_ar(this)
...
+    &      /1X,'NONLINEAR ITERATION METHOD            (NONLINMETH) = ', I9,    &
+    &      /1X,'LINEAR SOLUTION METHOD                   (LINMETH) = ', I9)
+    !
+    IF(this%nonmeth /= 0)THEN
+      WRITE(IOUT,9003) this%theta, this%akappa, this%gamma, this%amomentum,    &
+                       this%numtrack
+      IF(this%numtrack /= 0) WRITE(IOUT,9004) this%btol,this%breduc,this%res_lim
+    END IF
+    !
+    ! -- under-relacation formats
```
langevin-usgs: "under-relaxation"

jdhughes-usgs: fixed in latest commit

PR #109, opened 2019-03-12
feat(obs/ts/tas): dfn support for new flopy obs/ts/tas interface

spaulins-usgs:
Syncing dfn files between flopy and mf6.

### DIFF
--- a/doc/mf6io/mf6ivar/dfn/exg-gwfgwf.dfn
+++ b/doc/mf6io/mf6ivar/dfn/exg-gwfgwf.dfn
@@ -153,6 +153,9 @@ description is the file name of the water mover input file to apply to this exch
 block options
 name obs_filerecord
 type record obs6 filein obs6_filename
+construct_package obs
+construct_data continuous
+parameter_name observations
 shape
 reader urword
 tagged true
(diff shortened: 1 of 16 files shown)


langevin-usgs:
Scott, you're gonna also need to modify this file to say what these new things are (https://github.com/MODFLOW-USGS/modflow6/tree/develop/doc/mf6io/mf6ivar).  And keep in mind that the descriptions should be MODFLOW descriptions, not necessarily flopy descriptions.  Hopefully these new attributes are broadly useful.

spaulins-usgs:
The readme has been modified. I tried to describe these attributes generally, but also included how flopy specifically uses them as an example of their use. I'm not sure if the new attributes will be broadly useful, though they might be useful to another program that interfaces with MF6. The attributes are designed to explicitly define the differences between the flopy interface and the modflow6 package file format. If you don't like these new attributes I can move them to separate flopy-specific definition files.

langevin-usgs:
@spaulins-usgs, my preference would be to burn these things into flopy for now.  The ts, tas, and obs are low-level functionality that can be custom coded in flopy.   In MODFLOW speak, we don't refer to these as "child packages."  Instead we describe them as low level utilities.

spaulins-usgs:
@langevin-usgs, No problem. I will put these things into a "flopy.dfn" file that will not be part of MF6.

PR #161, opened 2019-06-10
First version of "callable MF6"

pya:
The wrapping concept with `f2py` works. This version allows to access all variables exposed by the memory manager. As in the proof-of-concept `Fython`, this version works with the callback principle and data exchange via a shared Fortran `module`.  There is no direct access to the data structures the pointer from the memory manger provides. Therefore, a read and writing `subroutine` for each datatype and dimension is necessary.  This ends up in 12 `subroutines` ( 2 (int, float) x 3 (0d, 1d, 2d) x 2 (get, set)). After splitting up in Fortran, it will come back together in Python.

This is the resulting Python interface: 

```
value = get_value(name, origin))
set_value(name, origin, value))
```
Here `value` can be an `int` or `float` as well as a 1d or 2d arrays of these types. Lists and NumPy arrays both work.

The code needs such a datatype table:

```
MF6_DATA_TYPE_TABLE = {
    ('NPER', 'TDIS'): {'data_type': 'int_scalar'},
    ('DELT', 'TDIS'): {'data_type': 'float_scalar'},
    ('NSTP', 'TDIS'): {'data_type': 'int_1d',
                       'dim': ('NPER', 'TDIS')},
    ('PERLEN', 'TDIS'): {'data_type': 'float_1d',
                         'dim': ('NPER', 'TDIS')},
    ('LRCH', 'SLN_1'): {'data_type': 'int_2d',
                        'dim': (3, ('MXITER', 'SLN_1'))},
    ('MXITER', 'SLN_1'): {'data_type': 'int_scalar'},
    ('AUXVAR', 'GWF_1 WEL'): {'data_type': 'float_2d',
                              'dim': (('NAUX', 'GWF_1 WEL'),
                                      ('MAXBOUND', 'GWF_1 WEL'))},
    ('NAUX', 'GWF_1 WEL'): {'data_type': 'int_scalar'},
    ('MAXBOUND', 'GWF_1 WEL'): {'data_type': 'int_scalar'},
}
```

Its is still very incomplete. Maybe it can be automatically populated. Also, it should probably move to an external JSON file.

Make works with `make/make_f2py/compile_f2py`. It assumes that the object files are already in `make/obj_temp/`. Also `gfortran.o`must exist there. The generated Python extension ends up in the folder `python`. Run from there by specifying the absolute path to the name file as argument. 

Let me know if this works for you.


langevin-usgs:
At the moment, this PR is for discussion purposes only and will not be merged into develop any time soon.  @pya has experience coupling fortran and python programs and with this PR has shown a possible path forward in terms of driving MODFLOW using Python.  The idea is that a Python program could get and set all variables in the MODFLOW memory manager and allow the user to control execution using a callback function.

pya:
The impact on the "standalone" MF6 is minimal. The only changes are:

1. The main program moved in the `subroutine mf6_sub(cback)`.
2. The main program looks like this:

```
!
program mf6

  use :: cbackmod

  call mf6_sub(cback)

end program mf6

```

3. The `subroutine cback()` is empty:

```
module cbackmod
contains
  subroutine cback()
  !empty
  end subroutine
end module
```

No other changes here.

The "callable" version is a Python extension that is compiled totally independently  from the "standalone" version. Therefore, the files `shared_data.f90` and `access_memory.f90` are not need for the "standalone" version.

pya:
Maybe it is a good idea to put the "callable" part into its own repo. This means only the files:

```
backmod.f90
mf6.f90
mf6sub.f90
```

and the slightly modified Makefile (added the new source files) need to be added / merged. This way, the development of the "callable" version can be done independently from the "standalone" one.

langevin-usgs:
Yeah, it might be better to make this a separate repository.  And right now this pull request fails our test, and I wasn't able to determine why that is.  It looks like the script we use to compile isn't working at the moment.  If you have time, and want to make this a separate repo, it might be good to write a test for it that runs on Travis so that we can verify it is working.  I'll think about whether or not we want to take the main MODFLOW 6 routine, and make it a subroutine so that it is easier to call.

jdhughes-usgs:
Sounds like a bad idea to have a separate repo because of the issues it will create for syncing both repos. Its seems it would be better to isolate the new code and have two different build processes. This may require modifications to pymake (exclude specific directories, etc.).

langevin-usgs:
Yeah, I agree with you, and I see that.  But I guess we have a long way to go before we'd want to accept this particular functionality into our MODFLOW 6 supported code base.  The current wrapper written by @pya is close to being almost completely independent, as long as our main program is written as a subroutine instead of as main().  I'm hesitant to create a branch in our repo until we've had time to discuss and determine if this is the path that we want to take and support.  I think the concept is promising, but there are a lot of details to work out.

pya:
I don't think we need to sync repos. Once the callback is in, it would be one-way, i.e. the "callable" would go with whatever is in the "standalone" repo. I don't see a two-way syncing here.

@langevin-usgs The "standalone" compiles for me by just typing `make`. Does the CI work for the  for the commit a I forked from?

langevin-usgs:
@pya, you need to be able to run nosetests from within the autotest folder.  Specifically, the test_build_modflow6() function in test000_setup.py appears to be broken.  This build chain has a dependency on pymake (https://github.com/modflowpy/pymake).  We do NOT use the makefile for building.  We use pymake so that we can build across operating systems and with different compilers.

pya:
@langevin-usgs  OK. I did not used `pymake` but the makefile. I will fix this. I will be traveling today, starting in less than an hour. Therefore, it might take a day or two until I can look at this.

pya:
I get the tests to pass. But sometimes get:

```
   use :: cbackmod
        1
Fatal Error: Can't open module file ‘cbackmod.mod’ for reading at (1): No such file or directory
compilation terminated.
```

re-running all test even though `cbackmod.f90` is there. This is not deterministic. Running the same test serval times some run pass and others fail.
Looks like it depends on the order `pymake` uses for the Fortran source code files.
Judging from the produced makefile this order is arbitrary and different for each each `pymake` run. Is there a away to force `pymake` to compile `cbackmod.f90` first before any other file?

langevin-usgs:
pymake builds a dependency tree, and should determine when cbackmod is needed.  The compile order might change from run to run, but that should be inconsequential.

langevin-usgs:
Do you have a circular reference?

langevin-usgs:
In mf6sub.f90, add a "use cbackmod".  pymake looks for "use" statements to determine the dependency order.

pya:
@langevin-usgs  I use this code now:


```
program mf6

  use :: KindModule, only: cback

  call mf6_sub(cback)

end program mf6
```
and moved  `cback` into the kind module:

```
module KindModule

...

  subroutine cback()
    ! empty

  end subroutine cback

end module KindModule 
```
This gives this error message:

```
use :: KindModule, only: cback
        1
Fatal Error: Can't open module file 'kindmodule.mod' for reading at (1): No such file or directory
compilation terminated.
```

The problem is that I need `cback` right at the beginning of the program. As soon as we are in subroutine it is too late.

pya:
Running pymake a second time works. Somehow `kindmodule.mod` seems to exits know.

pya:
I added another cover subroutine. Looks like it works now. :)

pya:
This gets way too big for a discussion in a RP. Therefore, I created a new project for experimenting: https://github.com/hydrocomputing/pymf6

There is a short explaination about the design principles: https://github.com/hydrocomputing/pymf6/blob/master/doc/principle.md 

You can conda-install it (see README). Currently this is Windows-only. But I will add Mac and Linux versions soon. 

There is also a short description on how to compile: https://github.com/hydrocomputing/pymf6/blob/master/DEVELOPER.md (There is certainly room for improvement here. ;) )

Let me know what you think.

pya:
Added a `conda` package for Mac OS X.

langevin-usgs:
Deleting this PR for now.  @pya has a separate repo for this functionality at: https://github.com/hydrocomputing/pymf6

pya:
Good. But in the long-term I would like to get a few changes merged. Otherwise I would nee to keep modifying my Fortran sources if anything changes in the main program. I guess I will create a new PR with only the relevant changes. Would this work?

langevin-usgs:
How about creating an issue instead?  That will give us a chance to think about the changes and make any other adjustments that we need to make for our next release.  In the issue, please be specific about the restructuring that you'd like to see.

pya:
Sounds good. I will create an issue that explains what I did and what kind of changes I would like go into the  MF6 source. :) We can discussion from there.

PR #194, opened 2019-08-30
refactor(sfr): move SfrDataType members into flat arrays

langevin-usgs:


langevin-usgs:
@jdhughes-usgs, I flattened most of the SfrDataType arrays, but left a few of the more troublesome ones in the SfrDataType derived type for now.  I may need to deal with them as the budget refactoring moves forward.  You want to push the squash-and-merge button if it all looks okay?

PR #197, opened 2019-09-17
refactor(griddata): refactor grid data to isolate read and constructor

jdhughes-usgs:


jdhughes-usgs:
I added idomain to disu but @langevin-usgs  should look at the changes prior to accepting pull request. Will need a usernode version of iac, ja, ihc, cl12, hwva, and angldegx(??) in addition to reduced node versions for grb. Maybe these can be ditched after writing the grb.

langevin-usgs:
Did you figure out a clean way to handle pass through cells with negative IDOMAIN values?  I think that was the main complication in getting DISU working with IDOMAIN.

jdhughes-usgs:
I didn't do anything besides adding `IDOMAIN` and refactoring so disu could be initialized from memory. The ia and ja is going to need to be processed to come up with a reduced ia and ja.

Issue #180, opened 2019-07-13
Make MF6 callable, i.e. live modifications of MF6 values with Python


pya:
# Motivation

The objective is to make MODFLOW 6 (MF6) callable. This means the model user is able to automatically interact with a running model, using a user-written Python program.

# Proposed Solution

The package [pymf6](https://github.com/hydrocomputing/pymf6) is a proof-of-concept implementation of a Python solution. The [documentation](https://pymf6.readthedocs.io/en/latest/index.html) explains how it works. The [principle](https://pymf6.readthedocs.io/en/latest/principle.html) shows how the user can change MF6 values at run time. The [current way of building](https://pymf6.readthedocs.io/en/latest/developer.html) depends on this [fork](https://github.com/hydrocomputing/modflow6) of MF6. 

# Needed Changes to the MF6 source

In order to make `pymf6` work with the source of this MF6 repo the following changes are necessary.

## 1. Move the code of the main program into a subroutine

The main program `mf6.f90` will call the subroutine `mf6_sub_outer`:

```fortran
program mf6

  call mf6_sub_outer()

end program mf6
```

## 2. Add a callback

The subroutine `mf6_sub_outer` in the file `mf6sub.f90`:

```fortran
subroutine mf6_sub_outer()

  use :: KindModule, only: cback

  call mf6_sub(cback)

end subroutine mf6_sub_outer
```

It calls the subroutine `mf6_sub` with an callback subroutine that comes from the module `KindModule`.

The subroutine `mf6_sub` is also located in `mf6sub.f90`:

```
subroutine mf6_sub(cback)

... 

tsloop: do
    !
    ! -- TIME UPDATE (TU)
    call tdis_tu()
    call cback()


....

enddo
```

It contains the original code from the main program. The only difference to the original code is the call to `cback()` at the beginning of of each time step loop. In `pymf6` this callback is a Python callable. In the standalone version `cback` is an external Fortran subroutine.

## Add an empty callback subroutine

The file `kind.f90` contains this code:

```
module KindModule

  implicit none

  public

...

  contains
....

  subroutine cback()
    ! empty

  end subroutine cback

end module KindModule
```

The subroutine `cback` is empty. This subroutine can be located in a different module as long this module is compiled at the very beginning of the build process. The new call to this empty subroutine should not add any run time, as the compiler will like inline this subroutine and therefore no additional instructions will be executed.

pya:
[...]

PR #205, opened 2019-10-01
refactor(mf6sub): convert main to a subroutine for external driver programs

langevin-usgs:


langevin-usgs:
This is in reference to https://github.com/MODFLOW-USGS/modflow6/issues/180, which allows MODFLOW 6 to be callable from other programs.

src/mf6sub.f90
```diff
@@ -0,0 +1,239 @@
...
+  ! -- TIME STEP LOOP
+  tsloop: do
+    !
+    ! -- TIME UPDATE (TU)
+    call tdis_tu()
+    !
+    ! -- MODFLOW 6 callback function allows an external program
+    !    to modify program during execution
+    call mf6cback()
+    !
```
pya: Please move `call mf6cback()` to line 143, i.e. after:

        do is=1,basesolutionlist%Count()
          sp => GetBaseSolutionFromList(basesolutionlist, is)
          call sp%sln_rp()
        enddo

src/mf6sub.f90
```diff
@@ -0,0 +1,239 @@
...
+      call ep%exg_rp()
+    enddo
+    !
+    ! -- Read and prepare each solution
+    do is=1,basesolutionlist%Count()
+      sp => GetBaseSolutionFromList(basesolutionlist, is)
+      call sp%sln_rp()
+    enddo
+    !
+    !
```
pya: Move  call mf6cback() here.

pymake/makefile
```diff
@@ -43,108 +43,114 @@ ${SOURCEDIR14}
 
 # Define the Fortran compile flags
 FC = gfortran
-FFLAGS = -O2 -fbacktrace -ffpe-summary=overflow -ffpe-trap=overflow,zero,invalid 
+FFLAGS = -O2 -ffpe-summary=overflow -ffpe-trap=overflow,zero,invalid -fbacktrace --coverage
```
mjr-deltares: --coverage was added on purpose?

pymake/makefile
```diff
@@ -43,108 +43,114 @@ ${SOURCEDIR14}
...
 
 # Define the C compile flags
 CC = gcc
-CFLAGS = -O3 -D_UF 
+CFLAGS = -O2 -D_UF
```
mjr-deltares: not sure if it matters anywhere, but here we go from -O3 to -O2

pymake/makefile
langevin-usgs: Good catch!  I was testing with the fortran code coverage and forgot to remove this.

pymake/makefile
langevin-usgs: Switched back to 3.  No reason for the change.

langevin-usgs:
Good suggestions.  The one thing that deserves further discussion is how to handle additional callbacks in a more logical manner.  Should we have callbacks after each of our main procedures, like the following?  Or maybe since we are moving toward a refactoring of initialize/advance/finalize we should wait for a more robust solution until then?  Maybe the mf6cback_rp() is the only one that we implement in this first phase?

```
call simulation_cr()
call mf6cback_cr()

model/exg/sln_df()
call mf6cback_df()

model/exg/sln_ar()
call mf6cback_ar()

call tdis_tu()
call mf6cback_tu()

call model/exg/sln_rp()
call mf6cback_rp()

call calculate()
call mf6cback_ca()

call model/exg/sln_ot()
call mf6cback_ot()

call model/exg/sln_fp()
call mf6cback_fp()

call model/exg/sln_da()
call mf6cback_da()
```

@mjr-deltares, I like the idea of trying to get around passing all these subroutines as arguments to mf6sub.  Seems like there would be a more elegant way to do this, but @pya would know better. 

@pya, can we just "use" these procedures in mf6sub instead of passing them in?  Would that work for the python implementation?

langevin-usgs:
For BMI compliance, I think we are ultimately moving toward something like what is shown below for the main program, although we have been discussing a more advanced approach that allows for a convergence loop within the time loop so that we could support tight coupling with other models. 

This approach below complicates the current callback objective because one or more callback routines should reside within `sim.update()`.  With this in mind, it might be nice to pursue what @mjr-deltares has been suggesting about inserting the callback routines into the program in some other way, such as with defined interfaces.

```
call sim.initialize()

time = sim.get_current_time()
end_time = sim.get_end_time()

do while (time < end_time)
  call sim.update()
  time = sim.get_current_time()
end do

call sim.finalize()
```

pya:
I need to hand in the callback as an external function, i.e. as an argument. I don't think it can be put into a module and used by `use`. I can work with one callback that is called at different places in Fortran and get additional information which call is the current one. For example, there could be a string `callback_name` in `shared_data.f90` that can be changed to `mf6cback_cr` or `mf6cback_df` etc. All data will be exchanged via `shared_data.f90`. Would this work for now?

mjr-deltares:
What do you think about the following:

add a module variable like this:

`procedure(c_callbackiface), pointer :: theCallbackFct => null()`

and an externally exposed setter:

```
subroutine set_c_callback(c_callback) bind(C, name="set_logger_c_callback")
   !DEC$ ATTRIBUTES DLLEXPORT :: set_c_callback
   use iso_c_binding
   type(c_funptr), value    :: c_callback
  
   ! assign c_callback to theCallback here

end subroutine set_logger_c_callback
```
Then have `c_callbackiface` such that the callback passes an identifying string from where in the program it is called, e.g:

`call theCallbackFct("afterInit")`

You could then dispatch on the exact value of that string. Would that do the trick? I think it would be nicer then to have the callback passed as the argument of the main routine. Maybe, in the future, there will be multiple parties subscribing with their own callback routines.

langevin-usgs:
@mjr-deltares I'm not sure I fully understand what you are suggesting and how it will be used.  Is the idea that in MODFLOW we would call the routine with the location of where it was called from?  I don't want to sidetrack you from the other things you have going on, but maybe it would be best to see a very small example that demonstrates what you are proposing.

mjr-deltares:
@langevin-usgs Yes, I think so. This is a tiny msvs project to illustrate the idea:

[callback.zip](https://github.com/MODFLOW-USGS/modflow6/files/3685222/callback.zip)

You should be able to build it and run caller.exe

mjr-deltares:
@langevin-usgs Just thinking out load here, and contemplating whether a callback mechanism would be required. If we would restructure the top level of the program for use with BMI anyway and with that subroutines will become available like 
`initialize()
prepareUpdate()
update()
processUpdate()
finalize()`
etc. Could you then not call them from outside and manipulate the simulation in between, any way you like? @pya Could that work as well for your application?

langevin-usgs:
Thats a good question, @mjr-deltares, and I think it's why we are treading slowly here to explore all the possibilities.  If we make MODFLOW BMI-compliant, it seems like it would be an alternative to what @pya has already implemented.  Would be good to understand the benefits and limitations of each approach.  The nice thing about a callback that I see at the moment, is that we could place them anywhere in the code (provided we can "use" it and not have to pass it around).  This would allow us to monitor and even visualize convergence behavior, for example, or make adjustments in the code while it's running.

pya:
I use `f2py` (part of NumPy, https://docs.scipy.org/doc/numpy/f2py/python-usage.html). So far I provide only one external function, i.e. the callback function. I can try to  use five callbacks:

1. `initialize()`
2. `prepareUpdate()`
3. `update()` 
4. `processUpdate()`
5. `finalize()`

Would this always be 5 functions? Any reason to have more? "Turning off" a callback is easy, just give it a function that does nothing. Adding one more callback would lead to code changes.

I am at PyCon DE in Berlin. There will be sprints over the weekend. I will work on `pymf6` and will test if  such a multi-callback approach works in principle.

mjr-deltares:
@pya My previous point was not to use 5 callbacks instead of 1, but whether you would need the callback at all, if we were going to structure MF that way. And to answer your questions, 5 is just an arbitrarily chosen number here, as are the names of those subroutines. I guess I am trying to explain how a different code structure would easily allow you to control the program from outside with a certain granularity, such that you can straightforwardly interact (or intercept) at certain spots in the control flow. To get a feeling for how such an interface could look like, see: https://bmi.readthedocs.io/en/latest/

This will take some time to design and implement though, and to have something available on a shorter term, the callback mechanism might be the way to go. However, in that case, I would still argue that it is better not to pass the callback as the argument of the main routine, but use the approach that I illustrated in the minimal example `callback.zip` above.

pya:
I am using `f2py`. This comes with two import points:

1. `f2py` takes care of all low-level and error-prone details, interfacing Python with Fortran via C. Therefore, I can access Fortran arrays as NumPy arrays. Furthermore, the "callback trick" allows me to get into the time step loop with a "least invasive method".
2. On the other hand, I have to play by the `f2py`rules. This means I can only supply a callback as an argument to a Fortran subroutine. I am not aware of a different way of doing this.

@mjr-deltares: Your example `callback.zip` works with a "normal" DLL. This is different from the *.pyd` file `f2py` produces. This is a special DLL, holding a Python extension. I don't think you can access this `*.pyd` file from C++ in the standard way as you do with a "normal" DLL.

Why not separate both (or more) approaches by using one source code for the different backends:

1. Standalone executable
2. BMI-interfaced library for all kinds of coupling
3. Callback-based Python extension

plus as little as possible backend-specific code. Based on this idea, I see a code structure like this:

```
src
    common
          common_code_1.f90
          common_code_2.f90
          ...
          common_code_n.f90
    backends
         standalone
             standalone_1.f90
             ....
         bmi
             bmi_1.f90
             ....
         callback
             callback_1.f90
             ...
```

Objectives:

* keep as much code in `common` as possible
* don*t sprinkle lots of `if <case>`(where `case` stand for 'standalone`, 'bmi`, `callback`)
   into the code
* use empty mock subroutines for a case if a specific subroutine is not used (just like
   the empty `cback` subroutine for the `standalone` case)

I already do this with `pymf6`. This is my current top-level directory layout:

* `doc` - Sphinx-based documentation 
* `mfsrc`- MF6 source as clone from the USG repo
* `fsrc` - Fortran source code specific to `pymf6`
* `make` - `pymake` scripts
* `pymf6`- Python source files

This structure may be improved but it currently works for me.

Would this approach work?  My objective here is to minimize coupling and dependencies between the different backends. So, if in the future somebody would like to add yet another backend, it should be straight forward following this approach. I am open to better ideas, offering simpler solutions, but I need to keep my callback approach for now because I already use `pymf6` for  my research project. I need to create considerable output over the next month. I cannot wait for the BMI interface. If it turns that the BMI backend is better than the callback backend, I would consider migrating if the effort-gain ratio is appropriate.

mjr-deltares:
Hi @pya, thanks for your clear explanation. I understand your point not being able to wait for a bmi. I rewrote the callback solution a bit to see if this could work for you:

[callback_v2.zip](https://github.com/MODFLOW-USGS/modflow6/files/3723274/callback_v2.zip)

So I took the DLL stuff out for clarity, as it is not relevant to your intended application and it also should not interfere. Now in this example, there is main modflow:

- runmf6.f90
- callback.f90

and there is a new backend

- mf6runwithcback.f90

The latter has the subroutine:

```
subroutine mf6cback(cback)
    procedure(cback_iface), pointer, intent(in) :: cback
    
    ! set the function pointer
    call set_cback(cback)
    ! run modflow
    call runmf6()
    
  end subroutine
```
Can you test your f2py on this? (In case you are not working with Visual Studio, it is really on these three files mentioned above, I guess.) This way you could have your own, very small, wrapper project in the solution. Similarly, we can have a bmi wrapper project and maybe other frontends in a later stage as well. It keeps modflow clean and has the callback mechanism in a dedicated module.

pya:
`f2py` does **not** work with pointers (yet). Furthermore, as far as I know the `external`  is required to make it work. I will traveling this week and therefore I can test it in more detail only next week.

PR #215, opened 2019-10-11
feat(cmask): add mask for calculation of amat terms

mjr-deltares:
(to prepare for the upcoming interface model) the mask allows to skip insertion and calculation of amat coefficients for certain connections.

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3npf8.f90
+++ b/src/Model/GroundWaterFlow/gwf3npf8.f90
@@ -83,6 +83,7 @@ module GwfNpfModule
     procedure                               :: npf_ac
     procedure                               :: npf_mc
     procedure                               :: npf_ar
+    procedure                               :: npf_init_mem
     procedure                               :: npf_ad
     procedure                               :: npf_cf
     procedure                               :: npf_fc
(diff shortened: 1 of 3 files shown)


langevin-usgs:
Yes, things would go very wrong.  But I was thinking that we would have a connections%set_mask() method, that would be the way someone would set a mask.  If someone changes connections%mask on their own, well, they will suffer the consequences.  Because these connection-length arrays can be pretty big, we should carefully consider whether we want to allocate that additional memory.

mjr-deltares:
ok, I see. So, we leave the mask public, add a setter, and put a comment in place to urge the programmer to actually use that setter. I will commit those changes.

mjr-deltares:
@langevin-usgs maybe one more thing, when you say set_mask, is that to replace the entire array, or do you want to set a single mask: `set_mask(ipos, maskValue)`

langevin-usgs:
Good question.  Maybe it should be `set_mask(ipos, maskValue)` and it could be programmed to be smart.

```
subroutine set_mask(ipos, maskValue)
  ...
  if (associated(this%mask, this%ja) then
    allocate this%mask with size nja (perhaps with mem_allocate()?)
    do i = 1, this%nja
      this%mask(i) = 1
    end do
  end if
  this%mask(ipos) = maskValue
  
end subroutine set_mask
```

mjr-deltares:
Yes, I will program it like that.

PR #226, opened 2019-10-16
ci(.travis.yml): upgrade to xenial, replace miniconda with apt-get

mwtoews:
* reduce repetition in test matrix
* remove inactive features like sudo and cache: apt
* install packages with apt-get rather than miniconda

Supersedes #225

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,138 +1,40 @@
 language: c
 os: linux
-dist: trusty
-sudo: required
+dist: xenial
 compiler: gcc
 cache:
-  apt: true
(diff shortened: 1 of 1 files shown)


langevin-usgs:
Well that sure looks a lot cleaner!

langevin-usgs:
@mwtoews, thanks so much for putting this together.  You clearly have the linux skills to get this setup properly.  This looks like a much cleaner approach.  Not sure how you figured this all out, but it looks to be running perfectly.  Thanks.

PR #237, opened 2019-10-30
Refactor src

mjr-deltares:
Let's have a draft PR first to test what this splitup does to the build. From there continue with next steps (bmi/dll etc)


mjr-deltares:
OK, great. Next step for me is to implement a first, basic BMI, where we can steer modflow from outside. I will continue to do that on the refactor-src branch, if we all agree.

PR #239, opened 2019-10-31
 Bugfix: cellxy in DIS

mjr-deltares:
- fixed how celly is determined from column numbers


mjr-deltares:
This is the fix in determining the y coordinate from the column number, that I was talking about. It should have been part of PR228.

PR #240, opened 2019-10-31
Refactor src

mjr-deltares:
Next step towards bmi: regrouping logic in the main routine to come to something that will look like

- init
- loop: update
- finalize

and finer grained variations of it, such that we can call it from outside (python, C, fortran).


jdhughes-usgs:
Closed by #255

PR #259, opened 2019-11-19
feat(csub): merge CSUB feature branch into develop

jdhughes-usgs:


jdhughes-usgs:
@langevin-usgs I will let you do the pull request for the csub package

doc/version.py
```diff
@@ -1,11 +1,11 @@
 # MODFLOW 6 version file automatically created using...pre-commit.py
-# created on...February 22, 2018 10:50:04
+# created on...May 01, 2018 13:46:06
 
 major = 6
 minor = 0
-micro = 3
```
langevin-usgs: These version numbers aren't correct.  Probably should leave these as is.

PR #260, opened 2019-11-25
Refactor src

mjr-deltares:
A first working version of bmi (exposing c-style pointer to our internal arrays). We still need to discuss how we can use the CSDMS interface explicitly, and how they have wrapped it into Python.

### DIFF
--- a/dllsrc/bmi.f90
+++ b/dllsrc/bmi.f90
@@ -12,56 +12,56 @@ module bmif
 
   type, abstract :: bmi
      contains
-       !procedure (bmif_get_component_name), deferred :: get_component_name
-       !procedure (bmif_get_input_var_names), deferred :: get_input_var_names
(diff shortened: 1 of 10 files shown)


mjr-deltares:
apparently the floating point model on Travis is not equal to the one on my Windows machine, nor to the one I have on my Linux virtual box. Currently I tried making the implementation safe by explicitly assigning `totalsimtime` to `totim` when the simulation is finished.

langevin-usgs:
Working on Travis is like feeling around in the dark...

mjr-deltares:
Yes, it is :-) 

The floating point issue is solved I guess, but would you have an idea why that last test (testmnw2) is still failing?

mjr-deltares:
O, wait, it is quite recent, let me pull develop in first...

Issue #33, opened 2018-08-10
IDOMAIN capability for DISU


langevin-usgs:
Need to keep this one on the table.  Strictly speaking, it's not necessary for DISU, but it might make life a lot easier for MODFLOW-USG users.

jdhughes-usgs:
Also requested by Alaistair Black and Sorab Panday

langevin-usgs:
Thinking about how to implement this.  Let's say we have the following case:

![image](https://user-images.githubusercontent.com/3372403/59273030-ea562d80-8c1c-11e9-8d45-602466c44817.png)

Now, what if cell 3 is assigned an IDOMAIN value of -1 (pass through cell)?  We know that 3 is connected upward to 1 and downward to 6, which means we can bypass 3 by connecting 1 to 6 and 6 - 1 using the flow areas of the 3-1 connection and 3-6 connection.  This is straightforward, and we end up with this:

![image](https://user-images.githubusercontent.com/3372403/59273144-212c4380-8c1d-11e9-8f20-0cb80755451c.png)

The new user JA array, as written to the GRB file is:
1 2 4 5 6
2 1 4 6
3
4 1 2 5 6
5 1 4 6
6 1 2 4 5

Note that this connectivity is different from what the user enters (and we have three different connectivity arrays!).  I think we want this behavior as it clearly shows that cell 3 has no connections and that cell 1 is connected to cell 6.  Ideally, we'd just want to use the JA array specified by the user, but we have to add new connections so this isn't really possible.  And this writing of a new JA to the GRB file is how we handle IDOMAIN for DIS and DISV.  The difference here is that the user enters JA for DISU, and the resulting JA array will not be the same, although it will be written to the GRB file.

So now for a complication.  What about this case?

![image](https://user-images.githubusercontent.com/3372403/59273329-95ff7d80-8c1d-11e9-9e18-9ab5199d23df.png)

The user specified JA array is
1 2 3
2 1 3
3 1 2

The challenge here is that if we turn 2 into a pass through cell, we don’t just make a new connection between 1 and 3, because they are already connected.  And, I don't think the code and solver format allows us to have two connections between the same two cells?  I guess in this case, we'd have to add the flow areas so that A13 = A13 + A12.  I'm not sure we'd have to do anything with the CL12 array, except be sure to use the L13 and L31 lengths, and discard the L12 and L21 lengths.

[...]

PR #261, opened 2019-11-25
feat(disu idomain): add support for disu idomain

langevin-usgs:
* user can specify idomain for disu package
* vertical flow-through cells not supported (idomain must be 0 or 1)


langevin-usgs:
Addresses #33

jdhughes-usgs:
Do you want me to merge or do you want to do it?

PR #294, opened 2020-01-28
fix(csub): Add text to mfio document describing CSUB binary output

### LINKED ISSUE #290: CSUB flow data stored improperly
When the 'CSUB-ELASTIC' and 'CSUB-INELASTIC' data is stored in the cell-by-cell flow file in a model using DIS or DISV, NCOL, NROW, and NLAY are not stored properly. It looks like the data is always stored as if the model was a DISU model. This makes reading the data for these two data sets unnecessarily complex.
[CSubIssue.zip](https://github.com/MODFLOW-USGS/modflow6/files/4109422/CSubIssue.zip)

jdhughes-usgs:
closes #290


jdhughes-usgs:
@langevin-usgs Do you want to take a look at these changes? These fix specific issues identified by @rbwinst-usgs.

PR #297, opened 2020-01-30
feat(lnf): Add initial lnf geometry dfns and examples

jdhughes-usgs:
@spaulins-usgs after this is merged into the `lnf` branch could you look at the `test_lnf_disl01.py` autotest script to see how to resolve the issues with specifying `idomain` in the `disl` file and with the `cgeo`, `rgeo`, and `ngeo` packages (these issues are all commented out now). The  `cgeo`, `rgeo`, and `ngeo` packages probably also need to support multiple versions of the packages.


spaulins-usgs:
@jdhughes-usgs the main problem was with the grid class not supporting the disl discretization.  I have made disl a type of vertex grid and modified the vertex grid class to support cell1d.  I still need to write a few tests for this.

The ngeo package is another issue.  In the ngeo package dfn you define xzpairs as type "double precision" and length "npoint".  When defining it like this flopy does not understand that you expect pairs of numbers and therefore it only expects npoint double precision numbers, when there are really npoint*2 double precision numbers.  The fix for this depends on how you want flopy to store the npoints in a recarray.  For example, flopy could potentially store each double precision number in a separate column or it could store each pair together, as a tuple, in a single column (like what it is doing for cellid).  The first option only requires minimal code changes while the second option will require a little more code to implement.  Let me know what your preference is.

PR #370, opened 2020-03-30
Output in bmi

mjr-deltares:
- writing stdout to file or screen is now configurable through bmi

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3sto8.f90
+++ b/src/Model/GroundWaterFlow/gwf3sto8.f90
@@ -23,6 +23,7 @@ module GwfStoModule
     integer(I4B), dimension(:), pointer, contiguous  :: iconvert => null()       !confined (0) or convertible (1)
     real(DP),dimension(:), pointer, contiguous       :: sc1 => null()            !primary storage capacity (when cell is fully saturated)
     real(DP),dimension(:), pointer, contiguous       :: sc2 => null()            !secondary storage capacity (when cell is partially saturated)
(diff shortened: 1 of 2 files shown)


jdhughes-usgs:
How do you set `istdout_to_file`? Could you add it to the mf6lib autotest?

mjr-deltares:
In python you would do:

```
# write output to screen:
mf6.set_int("ISTDOUTTOFILE", 0)
```
But the default is 'to file' and that should be ok for the tests, right?

jdhughes-usgs:
Is there anyway this could be done in the `initialize()`

mjr-deltares:
In the bmi initialize() in Fortran, it now reads:

```
function bmi_initialize() result(bmi_status) bind(C, name="initialize")
  !DEC$ ATTRIBUTES DLLEXPORT :: bmi_initialize
    integer(kind=c_int) :: bmi_status
    !
    if (istdout_to_file > 0) then
      ! -- open stdout file mfsim.stdout
      istdout = getunit() 
      !
      ! -- set STDOUT to a physical file unit
      open(unit=istdout, file=simstdout)
    end if        
    !
    ! -- initialize MODFLOW 6
    call Mf6Initialize()

etc..
```

Is that what you mean? Or do you want to pass it as an argument to the initialize call? The latter would not have my preference, because there will be more of these features. You could probably argue that the recently added option `iforcestop` should also be exposed this way...

mjr-deltares:
Question: how do I change the version of gfortran in the build process on my local machine? On travis, gfortran 4.8 and 5 of my interfacemodel stuff are failing.

PR #377, opened 2020-04-03
refactor(lak-period): combine period block recarrays

spaulins-usgs:
the two recarrays in lak package period block have been combined into a single recarray


langevin-usgs:
@spaulins-usgs, can you fix this and resubmit?  It's a simple fix.

spaulins-usgs:
@langevin-usgs It looks like the tests are now failing because it is using flopy to run the nightly build tests and I haven't checked in the flopy code yet.  I will do the flopy pull request first and then re-submit this pull request.

PR #408, opened 2020-04-17
docs(gwf-mvr.dfn): clarify mvrtype definition

### LINKED ISSUE #393: mf6io.pdf - MVR: clarification of mvrtype parameter
The last sentence in the blurb defining `mvrtype` seems a little ambiguous.  Specifically, the sentence that starts, "These options are based [on] the options available in the SFR2..."

factor_001

Does "SFR2" imply Niswonger & Prudic (2005) Documentation of the Streamflow Routing (SFR2)...?  In which case, maybe it is appropriate to add the reference here.  However, I think what is intended here is to refer readers to the SFR section within mf6io.pdf (i.e., a few sections up)?  If that is the case, it may be helpful to point readers to the specific variable being referred to, which I think is `cprior`?  However, please note the difference between the use of "FACTOR" (in `mvrtype`) and "FRACTION" (in `cprior`), so maybe this needs further clarification?  Perhaps the easiest fix is something like, "These option are based on the options defined for the cprior variable in the SFR package, with the term "FRACTION" being functionally equivalent to "FACTOR." Or something similar?

Fraction

emorway-usgs:
attempts to resolve #393

### DIFF
--- a/doc/mf6io/mf6ivar/dfn/gwf-mvr.dfn
+++ b/doc/mf6io/mf6ivar/dfn/gwf-mvr.dfn
@@ -216,7 +216,7 @@ tagged false
 in_record true
 reader urword
 longname mover type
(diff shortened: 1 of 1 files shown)


langevin-usgs:
Way to go @emorway-usgs.  First mf6 commit.

PR #412, opened 2020-04-22
Develop bmifixes

mjr-deltares:
Minor changes to the bmi framework:

- output to file or screen can be controlled trough bmi variable: ISTDOUTTOFILE
- storage coefficients can be altered through bmi

### DIFF
--- a/srcbmi/mf6bmi.f90
+++ b/srcbmi/mf6bmi.f90
@@ -15,8 +15,12 @@ module mf6bmi
   implicit none
   
   ! Define global constants  
-  integer(c_int), BIND(C, name="MAXSTRLEN") :: MAXSTRLEN = MAXCHARLEN
+  integer(c_int), bind(C, name="MAXSTRLEN") :: MAXSTRLEN = MAXCHARLEN
   !DEC$ ATTRIBUTES DLLEXPORT :: MAXSTRLEN
+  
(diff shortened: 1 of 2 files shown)


srcbmi/mf6bmi.f90
```diff
@@ -833,7 +816,7 @@ function get_model_name(grid_id) result(model_name)
       end if
     end do
     
-    ! TODO_MJR: error message, should never get here...
+    write(*,*) 'BMI error: no model for grid id \"', grid_id, '\"'
```
jdhughes-usgs, line 819: You should probably use the standard `sim_message` subroutine in the `GenericUtilitiesModule` module to write this message.

srcbmi/mf6bmi.f90
```diff
@@ -390,7 +400,8 @@ function get_grid_type_model(model_name, grid_type_f) result(bmi_status)
       grid_type_f = "rectilinear"
     else if ((grid_type_f == "DISV") .or. (grid_type_f == "DISU")) then
       grid_type_f = "unstructured"
-    else
+    else
+      write(*,*) 'BMI error: cannot find grind type for model \"', model_name, '\"'
```
jdhughes-usgs, line 404: Same here use the `sim_message` subroutine for controlled output.

PR #418, opened 2020-04-29
docs(gwf-sfr.dfn): clarify 2nd idv definition

### LINKED ISSUE #417: mf6io.pdf - SFR: idv defined twice
The first occurrence of `idv` appears in the `DIVERSIONS` block of SFR, as shown next:

first_instance
and is defined in relation to `ndv` appearing in the `PACKAGEDATA` block:

first_instance_definition
So far, so good.  The second occurrence of idv appears in the `PERIOD` block of SFR:
second_instance

and is defined:

second_instance_definition

However, I don't think the diversion number is the same thing as what is defined in the `DIVERSIONS` block?  For example, a simulation may have 4 different stream reaches that host 1 diversion each, in which case ndv would be set equal to 1 for these 4 reaches and the corresponding idv value would be equal to 1 for all 4 diversions specified in the `DIVERSIONS` block.  However, when one wants to specify the diversion rate in the `PERIOD` block, and needs to reference which diversion they are intending to specify, I think the second instance of `idv` (in the `PERIOD` block) in this example needs to be numbered 1 through 4, where 1 would be the first diversion listed in the `DIVERSIONS` block, 2 would be the second listed diversion, and so on.  Thus, I'll add a clarifying comment to the second definition of `idv`, unless coming up with a distinct variable name for the second instance would be better (and would still bolster the variable's definition)?

emorway-usgs:
attempts to resolve #417

### DIFF
--- a/doc/mf6io/mf6ivar/dfn/gwf-sfr.dfn
+++ b/doc/mf6io/mf6ivar/dfn/gwf-sfr.dfn
@@ -669,7 +669,7 @@ tagged false
 in_record true
 reader urword
 longname diversion number
-description diversion number.
+description an integer value specifying which diversion of reach RNO that DIVFLOW is being specified for.  Must be less or equal to ndv for the current reach (RNO).
 numeric_index true
 
 block period


langevin-usgs:
Got it.  Thanks.

PR #429, opened 2020-05-08
refactor(uzf): use read_value_or_time_series_adv() for timeseries

jdhughes-usgs:


jdhughes-usgs:
Why do you have reaches with connections that do not receive flow? Are these errors in diversion reaches?

emorway-usgs:
Resolved with #435

PR #443, opened 2020-05-15
docs(gwf-drn.dfn): offer clarification to AUXDEPTHNAME definition

emorway-usgs:


emorway-usgs:
Offering what I hope is a clarification for the definition of `AUXDEPTHNAME`.  The current definition uses two variable names, `HDRN` and `DDRN`, that are not provided anywhere in the document.  In place of `HDRN`, I think what was intended was `ELEV` which is defined in the DRN documentation (in the PERIOD block section).  Also, `DDRN` is now referenced as an example string given for AUXDEPTHNAME in hopes that it clarifies for the reader what DDRN is.

jdhughes-usgs:
approved

PR #460, opened 2020-05-29
refactor(bmi): refactor grid routines

Hofer-Julian:
Previously we used the 'save' attribute to allocate the memory and
persist over the call to the routine.
This is not fool proof (multiple calls before using the results will mess things up).
After this commit, the caller has to allocate the appropriate memory,
similar to get_var_shape.

Note: This commit also includes whitespace & line ending fixes.
Unfortunately I found no good way to entangle this afterwards...

### DIFF
--- a/srcbmi/mf6bmi.f90
+++ b/srcbmi/mf6bmi.f90
@@ -809,7 +809,7 @@ module mf6bmi
       end if
     end do
     
-    write(error_msg,'(a,i6)') 'BMI error: no model for grid id ', grid_id
+    write(error_msg,'(a,i0)') 'BMI error: no model for grid id ', grid_id
     call sim_message(error_msg, iunit=istdout, skipbefore=1, skipafter=1)
   end function get_model_name
 end module mf6bmi
--- a/srcbmi/mf6bmi.f90
+++ b/srcbmi/mf6bmi.f90
@@ -1,8 +1,8 @@
 ! Module description:
-! 
+!
 ! This module holds the MODFLOW 6 BMI interface. The interface matches the CSDMS standard, with
 ! a few exceptions:
-  
+
 ! - This interface will build into a shared library that can be called from other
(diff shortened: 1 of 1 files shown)


jdhughes-usgs:
from a terminal in your fork repo could you run the following:

   `git add --renormalize .`
   `git commit -m "Introduce end-of-line normalization"`

This should clear up the line ending changes but updating your fork of the repo to a standard line ending.

srcbmi/mf6bmi.f90
```diff
@@ -830,13 +809,7 @@ function get_model_name(grid_id) result(model_name)
       end if
     end do
     
-    write(error_msg,'(a,i6)') 'BMI error: no model for grid id ', grid_id
-    call sim_message(error_msg, iunit=istdout, skipbefore=1, skipafter=1)
-    
-    
+    write(error_msg,'(a,i6)') 'BMI error: no model for grid id ', grid_id
```
jdhughes-usgs, line 812: Unless these is a reason to do otherwise we are using `i0` and `g0` for integers and floats.

Hofer-Julian:
> git commit -m "Introduce end-of-line normalization"

Thanks for the tip, what I tried was [this](https://stackoverflow.com/a/21014534), but without success.
Your command chain gives me this:
```nothing to commit, working tree clean```

Btw, I did end up with these whitespace&line ending changes, because visual studio told me there where inconsistencies and asked me if it should fix them.
So I assume these changes are valid, they just shouldn't be in this commit...

jdhughes-usgs:
do you have the `.gitattributes` file in the root directory of your repo? I presume that you do but just wanted to check.

I visual studio on windows I have been changing the line endings to LF before I commit.

image

srcbmi/mf6bmi.f90
Hofer-Julian: Done.

jdhughes-usgs:
ok. sounds like you have everything setup correctly.

PR #463, opened 2020-06-01
feat(maw): add defined flow correction option

jdhughes-usgs:


jdhughes-usgs:
@langevin-usgs take a look at `maw_calculate_conn_term()` method and use of method to see if it seems it will work for density terms.

src/Model/GroundWaterFlow/gwf3maw8.f90
```diff
@@ -2506,36 +2521,29 @@ subroutine maw_fn(this, rhs, ia, idxglo, amatsln)
...
-            rterm = -rate
-            !
-            ! -- calculate storage derivative
-            drterm = -this%area(n) / delt
-            !
-            ! -- fill amat and rhs with storage components
-            rhs(iloc) = rhs(iloc) - rterm + drterm * hmaw
-          end if
-        end if
+        !!
```
langevin-usgs, line 2524: Should this code be commented out?

jdhughes-usgs: Yes. Removed code since it was adding zero to the `RHS`.

PR #475, opened 2020-06-07
fix(drn): fix issue with drainage depth option when mover is on

jdhughes-usgs:


jdhughes-usgs:
test051_uzfp2 test the drainage depth option against uzf. Both use the mover package.

Issue #494, opened 2020-07-02
MODFLOW 6 version 1.1.1 crashes with the attached example


rbwinst-usgs:
MODFLOW 6 version 1.1.1 crashes with the attached model. The following error messages are written to the screen.
[Colarullo2.zip](https://github.com/MODFLOW-USGS/modflow6/files/4866688/Colarullo2.zip)

forrtl: severe (66): output statement overflows record, unit -5, file Internal Formatted Write
Image              PC                Routine            Line        Source             
mf6.exe            00007FF6718E8E74  Unknown               Unknown  Unknown
mf6.exe            00007FF6718E6E8D  Unknown               Unknown  Unknown
mf6.exe            00007FF6716D0E98  BLOCKPARSERMODULE         264  BlockParser.f90
mf6.exe            00007FF6716D0428  BLOCKPARSERMODULE         247  BlockParser.f90
mf6.exe            00007FF6718B9DA4  TIMESERIESMODULE_        1424  TimeSeries.f90
mf6.exe            00007FF6718BB7DF  TIMESERIESMODULE_        1379  TimeSeries.f90
mf6.exe            00007FF6718BBEBE  TIMESERIESFILELIS          48  TimeSeriesFileList.f90
mf6.exe            00007FF6718BCC24  TIMESERIESMANAGER         136  TimeSeriesManager.f90
mf6.exe            00007FF6716D6CA8  BNDMODULE_mp_BND_        1260  BoundaryPackage.f90
mf6.exe            00007FF6716D1495  BNDMODULE_mp_BND_         172  BoundaryPackage.f90
mf6.exe            00007FF671704719  GWFMODULE_mp_GWF_         343  gwf3.f90
mf6.exe            00007FF671864C32  MF6COREMODULE_mp_         210  mf6core.f90
mf6.exe            00007FF6718612EA  MF6COREMODULE_mp_          30  mf6core.f90
mf6.exe            00007FF671861102  MAIN__                      5  mf6.f90
mf6.exe            00007FF6719AC87E  Unknown               Unknown  Unknown
mf6.exe            00007FF6719ACC64  Unknown               Unknown  Unknown
KERNEL32.DLL       00007FFC89757974  Unknown               Unknown  Unknown
ntdll.dll          00007FFC89B5A271  Unknown               Unknown  Unknown

langevin-usgs:
[...]

Issue #488, opened 2020-06-25
MODFLOW 6 terminates prematurely without any error message.


rbwinst-usgs:
The attached model from Susan Colarullo terminates prematurely without any error message.
[Colarullo2.zip](https://github.com/MODFLOW-USGS/modflow6/files/4832870/Colarullo2.zip)
[Colarullo1.zip](https://github.com/MODFLOW-USGS/modflow6/files/4832872/Colarullo1.zip)

PR #499, opened 2020-07-07
refactor(blockparser): modified to read unlimited length input lines

langevin-usgs:


langevin-usgs:
Attempt to partially address #488 and #494 by allowing very long lines to be read from input files

PR #533, opened 2020-07-31
refactor(OBS): refactor observations to use AddObsIndex & ResetObsIndex

jdhughes-usgs:


src/Model/GroundWaterFlow/gwf3lak8.f90
```diff
@@ -4884,139 +4902,149 @@ subroutine lak_rp_obs(this)
...
-      end if
+      end do
       !
-      ! -- check that index values are valid
-      if (obsrv%ObsTypeId=='TO-MVR' .or.                                         &
-          obsrv%ObsTypeId=='EXT-OUTFLOW' .or.                                    &
-          obsrv%ObsTypeId=='OUTLET') then
-        do j = 1, size(obsrv%indxbnds)
-          nn1 =  obsrv%indxbnds(j)
-          if (nn1 < 1 .or. nn1 > this%noutlets) then
```
langevin-usgs, line 4984: We had a bug with this before, and I think you may have introduced it again here.  For some obs, we have to check against NOUTLETS, not MAXBOUND, or NLAKES.  I can't tell if that's been done here.

src/Exchange/GwfGwfExchange.f90
```diff
@@ -2,7 +2,7 @@ module GwfGwfExchangeModule
 
   use KindModule, only: DP, I4B
   use SimVariablesModule, only: errmsg
-  use ArrayHandlersModule,     only: ExpandArray
+  !use ArrayHandlersModule,     only: ExpandArray
```
langevin-usgs, line 5: Obviously, would be nice to remove all this old commented code.

src/Model/GroundWaterFlow/gwf3sfr8.f90
```diff
@@ -2499,23 +2503,27 @@ subroutine sfr_bd_obs(this)
           call this%obs%SaveOneSimval(obsrv, v)
         end do
       end do
-    end if
-    !
-    ! -- write summary of package block error messages
-    if (count_errors() > 0) then
-      call this%parser%StoreErrorUnit()
-      call ustop()
+      !
```
langevin-usgs, line 2506: Are there line ending issues here?  Why is this code showing up as changed?

src/Model/ModelUtilities/BoundaryPackage.f90
```diff
@@ -1669,7 +1659,8 @@ subroutine bnd_bd_obs(this)
     do i=1,this%obs%npakobs
       obsrv => this%obs%pakobs(i)%obsrv
       if (obsrv%BndFound) then
-        do n=1,size(obsrv%indxbnds)
+        !do n = 1, size(obsrv%indxbnds)
+        do n = 1, obsrv%indxbnds_count
```
langevin-usgs, line 1663: Looks good how you introduced a member variable instead of using Count.  Makes sense to me.

src/Model/GroundWaterFlow/gwf3lak8.f90
jdhughes-usgs: The code is the same except for `size(obsrv%indxbnds)` is replaced by `obsrv%indxbnds_count`

src/Exchange/GwfGwfExchange.f90
jdhughes-usgs: commented code removed in latest pull request commit

src/Model/GroundWaterFlow/gwf3sfr8.f90
jdhughes-usgs: There was only a single comment line before. Added

```
!
! -- write summary of package error messages
```

src/Model/ModelUtilities/BoundaryPackage.f90
jdhughes-usgs: ok

PR #549, opened 2020-08-14
fix(maw): fix bug in maw newton correction terms

### LINKED ISSUE #546: MAW head not correctly modeled when its head is below screen bottom
When the head of GWF nodes connected to the well falls below the well bottom, the model wont update the MAW head correctly.

If the head of the connected cell is below a MAW well bottom, the MAW head wont be properly updated to represent **dry** condition. 

An single cell and single well example showing this problem can be downloaded at: https://github.com/MODFLOW-USGS/modflow6/files/5068947/MAW_Unit_Test.zip

jdhughes-usgs:
Add additional error checking of maw input variables. Added
clarification of when and how user-specified values are reset for
the various conductance equations. Add additional test to evaluate
newton-raphson solution for a steady-state dry multi-aquifer well
problem (test_gwf_maw08.py).

Closes #546


langevin-usgs:
Just an FYI that we should probably update the release notes since this has been around for a while.

ougx:
Still failed for the second example in #546

PR #550, opened 2020-08-15
feat(BndStrtOpt) add STRT option to MAW and LAK to read starting head from a binary file

ougx:
It will save and read the head verbatim. Setting the head to `DHNOFLO` or `DHDRY` is removed to make sure the original array is saved.


langevin-usgs:
@ougx, thank you for this PR.  We appreciate your contributions and are glad to see that you're developing new features.  In working on this MAW issue, we were also thinking of adding a similar option.  Rather than implement it like you did for MAW and LAK, however, we are going to implement something more general that will work for all the different PERIOD blocks and input data for the advanced packages.  This will allow us to more generally read any of those input data from files (ascii or binary) or just as vectors of numbers.  Stay tuned.

ougx:
No problem @langevin-usgs. I look forward to it. For now I will use this in my project.

PR #564, opened 2020-09-15
Exchangemodel

mjr-deltares:
note that this is a draft PR, please do not merge


src/Utilities/List.f90
```diff
@@ -20,6 +20,7 @@ module ListModule
     procedure, public :: Add
     procedure, public :: Clear
     procedure, public :: Count
+    procedure, public :: Contains
```
jdhughes-usgs, line 23: contains is a fortran keyword. This should probably be something else to avoid confusion.

src/Model/Connection/GridSorting.f90
```diff
@@ -0,0 +1,84 @@
...
+  use GenericUtilitiesModule, only: is_same
+  implicit none
+  private
+
+  public :: quickSortGrid
+
+contains
+  ! Sort an array of integers
+  subroutine quickSortGrid(array, arraySize, idxToGlobal)
+      integer, intent(inout), dimension(:) :: array
```
jdhughes-usgs, line 14: should standard MODFLOW 6 datatypes (I4B, etc) be used here and in `qsort_inline.inc`?

msvs/testmf6.sln
```diff
@@ -0,0 +1,31 @@
+﻿
```
jdhughes-usgs, line 1: should this be moved to a different directory or subdirectory for better organization?

jdhughes-usgs (changes requested):
A few other things:

1. does `msvs\upgrade.log` need to be added to the repo.
2. Could the `swap` and `rshift` functions in the Grid Sorting be added as overloaded methods in `src\Utilities\sort.f90`? Maybe we will never use them again but just a thought.

I don't think we have to address all of these things right now in order to merge this PR. These could be topics to discuss on Wednesday's call.

src/Utilities/VectorInt.f90
```diff
@@ -0,0 +1,143 @@
...
+  implicit none
+  private
+  public :: VectorInt  
+  
+  integer(I4B), parameter :: defaultInitialCapacity = 4
+  
+  ! This is a dynamic vector type for integers
+  type :: VectorInt
+    integer(I4B), private, allocatable :: values(:) ! the internal array for storage
+    integer(I4B) :: size                            ! the number of elements (technically this stuff should be unsigned)
```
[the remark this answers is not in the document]
mjr-deltares: The intention for this type is to use it to introduce dynamic memory when building up the interface model. Once built, it can be deleted again and it need not be in the memory manager. I figured that 'allocatable' would be a bit more efficient in such a case.

msvs/testmf6.sln
mjr-deltares: Ah, well spotted, I believe that's a remnant from when I thought that unit testing through such a VS project could be a good long term strategy for MODFLOW. Now, I believe our way of working with the autotests, and the API for more invasive tests, is to be preferred. I will delete this solution file.

PR #572, opened 2020-09-29
docs(uzt.tex): fix typo in \doc\mf6io\gwt\uzt.tex

emorway-usgs:
partial redo of #571.  Will close that one and resubmit in parts


emorway-usgs:
@langevin-usgs This just doesn't make any sense.  I only changed 'wells' to 'cells' in one of the .tex files.  This was passing with the first commit associated with #571 before I tacked on additional typo fixes to that PR and it started crashing.  It seems like this one simple typo fix offering shouldn't be causing wide-spread failure.

PR #586, opened 2020-11-04
fix(mf6bmi): function counting total exposed variables was skipping MEMREADONLY

emorway-usgs:


emorway-usgs:
While attempting to expose QSTO within the LAK package, the omission of MEMREADONLY in the now amended if statement was keeping my C# example from picking up on the fact that I had opened up QSTO in the BMI DLL

mjr-deltares:
Hi @emorway-usgs, if the variable is MEMREADONLY then it would technically not be an input variable. I think that using get_**output**_item_count() instead, will give you what you were after.

emorway-usgs:
Thanks @mjr-deltares, that is indeed what I was trying to do

PR #590, opened 2020-11-10
feat(src-constraints): concentration constraints for gwt src package

langevin-usgs:
Implemented a Newton approach for reducing mass injection/withdrawal in order to meet concentration constraints


langevin-usgs:
Turns out this Newton approach for constraining mass source and sink rates falls apart under some common conditions, so it is being abandoned for now.

PR #606, opened 2020-12-05
fix(uzf): clean-up of gwet calculations in the uzf package

### LINKED ISSUE #605: UZF-calculated actual ET exceeds potential ET
In the case of a shallow groundwater system with extinction depth spanning multiple layers, gwet is extracted from multiple layers and results in actual ET exceeding pET

emorway-usgs:
Includes a new autotest to ensure that actual ET does not exceed potential ET for each row-column index in the test model domain

fixes #605

### DIFF
--- a/autotest/test_gwf_uzf_gwet.py
+++ b/autotest/test_gwf_uzf_gwet.py
@@ -1,0 +1,405 @@
+import os
+import sys
+import numpy as np
+
+try:
+    import pymake
+except:
+    msg = 'Error. Pymake package is not available.\n'
+    msg += 'Try installing using the following command:\n'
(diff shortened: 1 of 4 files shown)


emorway-usgs:
For the 3 UZF tests that failed, the .hds.cmp check was failing np.allclose() with rtol=1e-5.  Manual checks showed that they passed when rtol=1e-4.  It is expected that this bug fix would alter the simulated .hds solution slightly because less water is removed by the UZF ET process.  To address the failure associated with this PR, the established .hds.cmp files were updated with PR [#4](https://github.com/MODFLOW-USGS/modflow6-testmodels/pull/4) in the testmodels repo.  I think if that PR is accepted, the CI associated with this PR could be restarted.

PR #619, opened 2021-01-07
removed the memory access mechanism altogether

mjr-deltares:


langevin-usgs:
What do you think @mjr-deltares?  Are we good to on this?

mjr-deltares:
Hi @langevin-usgs, squash 'm all, I'd say,

PR #627, opened 2021-02-04
refactor(setup): minor change to compile on windows

langevin-usgs:
Requires that pymake is up to date


langevin-usgs:
Accepting this minor PR even though one test failed.  It was the intermittent mf5to6 convert failure on an lgr problem on macos

PR #632, opened 2021-02-08
refactor(vscode): setup vscode support for mac/linux and windows

langevin-usgs:


mjr-deltares (approved):
(tested this on Windows too)

PR #636, opened 2021-02-11
fix(sfr): enforce divflow<=1.0 for divs with cprior type 'FRACTION'

### LINKED ISSUE #635: Enforce DIVFLOW bounds when cprior is type FRACTION
Need error check in SFR that enforces the bounds documented in mf6io.pdf (see screen shot below) when `cprior = FRACTION` for a diversion.  For example, if a user inadvertently sets `divflow` in the PERIOD block to something greater than 1.0 when `cprior` type is `FRACTION`, MF6 will create water and keep computing, leading to misleading results.  Further, no issue shows up in the budget table because everything will balance by the created water. 

up_bound

emorway-usgs:
Fixes #635


src/Model/GroundWaterFlow/gwf3sfr8.f90
```diff
@@ -2804,6 +2806,16 @@ subroutine sfr_set_stressperiod(this, n, ichkustrm)
         call read_value_or_time_series_adv(text, ii, jj, bndElem, this%packName,     &
                                            'BND', this%tsManager, this%iprpak,   &
                                            'DIVFLOW')
+        !
+        ! -- if diversion cprior is 'fraction', ensure that 0.0 <= fraction <= 1.0
+        cp = this%divcprior(ii)
+        divq = this%divflow(ii) 
+        if (cp == 'FRACTION' .and. (divq < 0.0 .or. divq > 1.0)) then
```
langevin-usgs, line 2813: Should probably use the constants DZERO and DONE instead of 0.0 and 1.0.

langevin-usgs (commented):
Looks good.  One small comment below, which is trivial, and is intended to get you using the constants in the ConstantsModule

src/Model/GroundWaterFlow/gwf3sfr8.f90
emorway-usgs: Thanks for the pointer @langevin-usgs, have patched and PR is now rerunning

emorway-usgs: @langevin-usgs I looked at the failure and it occurred with one of the models in modflow6-testmodels.git, specifically one that resides in mf5to6:
mf5to6
Not being sure if the error was anomalous, I tried running the model locally after running the conversion program:
mf5to6_error
It doesn't seem like this error is related to the change made in this PR? Since the first commit on this PR passed, then changing 0.0 -> DZERO in the second commit caused it to fail, I'm wondering if we should try and get it to run again?

PR #637, opened 2021-02-12
feat(ats): exploration of adaptive time stepping

langevin-usgs:
* Overall wiring is set up for models, exchanges, and solutions to set a preferred time step size.  
* There is work to be done to redo a time step with a shorter length if the time step fails.  Probably need to skip the finalizeSolve() if redoing a time step.
* Need to implement ATS options in models, packages, exchanges, solutions, and so forth so that these individual components can determine a preferred step size
* Should add and test an optional courant stability constraint (and possibly others) to the GWT model
* Need to handle case where a time sliver might occur to meet the end of a period or a print time
* Need to consider support of print/save times in Output Control, however, the time step constraint that results from ending on a print/save time can bubble up to tdis through the model_tu() method.
* Time series will require some careful consideration


langevin-usgs:
Will break this up into separate PRs.  Closing.

PR #654, opened 2021-03-06
fix(uzf): set proper iuzno index for water-content observation type

### LINKED ISSUE #653: UZF 'water-content' observation type inappropriately erroring out
The small example model attached will error out with the following message:
```
ERROR REPORT:

  1. WATER-CONTENT for observation BNDNM_UZF09 specified depth (
     7.500000000000000 ) must be between 0.0 and  4.500000000000000.
```
However, the uzf object in question is in a cell with top elevation -5.0 and bottom elevation -20.0 (see attached).  According to the documentation, observations of this type are "... at a user specified depth (ID2) relative to the top of GWF cellid for a UZF cell," which should mean that a depth of 7.5 is halfway down the uzf object.  Looking in the source code, it seems line 2822 is the culprit: 
```
dmax = this%uzfobj%celtop(n) - this%uzfobj%celbot(n)
```
`n` is not the correct index to be using here.  

[uzf_obs_err.zip](https://github.com/MODFLOW-USGS/modflow6/files/6095386/uzf_obs_err.zip)

emorway-usgs:
fixes #653

### DIFF
--- a/doc/ReleaseNotes/ReleaseNotes.tex
+++ b/doc/ReleaseNotes/ReleaseNotes.tex
@@ -187,7 +187,7 @@ This section describes changes introduced into MODFLOW~6 for the current release
 
 	\underline{ADVANCED STRESS PACKAGES}
 	\begin{itemize}
-	        \item --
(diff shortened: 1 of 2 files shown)


emorway-usgs:
Superseded by #660.

PR #660, opened 2021-03-13
fix(uzf): set proper iuzno index for water-content observation type

emorway-usgs:
Redo of #654 owing to potential conflicts after changes made in #652


emorway-usgs:
Superseded by #664

PR #681, opened 2021-03-27
fix(SmoothingFunctions): process optional ta in sSlope and sSlopeDerivative

mwtoews:
The optional `ta` argument was not properly processed in functions `sSlope` and `sSlopeDerivative`.

These functions are not used anywhere else in the codebase.


langevin-usgs:
nice find

PR #682, opened 2021-03-27
fix(get_filename): initialize ipos, simplify backslash check logic

mwtoews:
The local variable `ipos` was potentially uninitialized. The logic to find the backslash can be simplified as shown.

### DIFF
--- a/src/Utilities/Sim.f90
+++ b/src/Utilities/Sim.f90
@@ -141,6 +141,9 @@ subroutine get_filename(iunit, fname)
   integer(I4B) :: ilen
 ! ------------------------------------------------------------------------------
   !
+  ! -- initialize variable returned by index; zero means substring not found
+  ipos = 0
+  !
   ! -- get file name from unit number
   inquire(unit=iunit, name=fname)
   !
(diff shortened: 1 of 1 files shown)


mwtoews:
I've been thinking about this one a bit more, and it might make sense to check for both `\` and `/` without consideration of the operating system that is running. This would enable the same behaviour on different operating systems with a simulation that originated from a different operating systems.

On Windows `/` is an illegal filename character, so this should checked in this function too.

On a related note, using forward slashes like `path/to/file.txt` works on all operating systems, including all current supported versions of Windows. The versions of Windows and DOS that didn't support these styles of file paths are long gone.

langevin-usgs:
Thanks for looking into this @mwtoews.  I'm not sure how much effort we want to put into this particular subroutine.  It has a very limited purpose at the moment to strip off the path prefix from a filename.  It is only used in a few cases for a deprecation warning as far as I can tell.  @jdhughes-usgs, any thoughts?

mwtoews:
Yet another library is the USGS ["Batteries Included Fortran Library"](https://doi.org/10.5066/P9K2IQ6Y), in particular [src/system/path_interface.f90](https://code.usgs.gov/fortran/bif/-/blob/main/src/system/path_interface.f90).

Adding a C dependency just for path support seems a bit excessive, unless there is any other need, like there was with the [GMG solver](https://pubs.usgs.gov/of/2004/1261/).

langevin-usgs:

PR #685, opened 2021-03-30
refactor(imslinear): standardize comments

jdhughes-usgs:


jdhughes-usgs:
@verkaik added you a reviewer but no need to start now since there is still alot to do. Feel free to add comments as I go or wait until it is ready.

PR #688, opened 2021-04-01
Rfct npf setdata

mjr-deltares:
These are the discussed changes to npf for our interface model, should have no impact on the current use cases.


langevin-usgs (approved):
I think this is a great way to get something implemented quickly.  I pulled it down and ran tests on my end and everything checks out.  We can continue to improve on this concept, like perhaps storing npfdata and npfoptions with the npf package itself, and perhaps passing them into subroutine readers, but we can hold off on that for now.

mjr-deltares:
that's great, let me just add a bit more documentation and then I will pull it in

PR #691, opened 2021-04-06
Extended time series functionality

damianmerrick:
Add internal option to extend time series to the end of the simulation.
Add internal option to remove completed time series links from a time series manager on call to ad procedure.
Both new options are inactive unless specified explicitly by the calling package (used in upcoming tvk and tvs code so far).


jdhughes-usgs:
@damianmerrick We don't want to ignore warnings for unused variables ("Omit #7712 unused variable remarks on Windows Intel Fortran compilations."). While we know we still get faulty warnings about unused variables with `this` related to overriden procedures seeing the warnings makes it easier to address unused variables that should be removed from the code.

damianmerrick:
> @damianmerrick We don't want to ignore warnings for unused variables ("Omit #7712 unused variable remarks on Windows Intel Fortran compilations."). While we know we still get faulty warnings about unused variables with `this` related to overriden procedures seeing the warnings makes it easier to address unused variables that should be removed from the code.

I just checked now, and there are currently 500 #7712warnings on a normal Release x64 build in VS. About 150 of those are "this" variables as you suggest, but the rest are other variables. To me, this makes it very difficult for a contributor to spot other warnings which may be more critical when modifying or adding new code.

It's certainly a minor detail and not the main point of this PR, so I am fine with changing this back if you insist. But I would encourage some consideration to either fixing the unused variable issues, or disabling the warnings until such time as you are ready to do this, to avoid hiding potential future warnings in the build log.

damianmerrick:
I have now removed the #7712 unused warning disable option from the Visual Studio project files as requested.

PR #693, opened 2021-04-06
TVS package feature addition (time-varying specific storage and specific yield)

damianmerrick:
Add time-varying SS and SY support to STO, as TVS sub-package available through additional file include in STO OPTIONS block.
Generalize STO formulate and budget routines to correctly integrate changes in stored water volume due to changes in SS and/or SY, when new STO type variable integratechanges is non-zero. This is applied by default when TVS is active in a model, and not applied otherwise unless the integratechanges variable is manually set non-zero (e.g. via BMI).
Time series extensions from previous PR are included in this PR as they are used by TVS.


damianmerrick:
Closing this PR in favour of the combined PR on develop

PR #694, opened 2021-04-08
refactor(numEx)

mjr-deltares:
Refactoring preparing the interface model approach, next to the existing exchanges. This PR should not affect any test outcome.


mjr-deltares:
Hey Chris, can you have a look at this and pull it in if you agree (assuming all tests pass)? It's the first step, now NumericalExchange is really only the interface to NumericalSolution. All exchange data, input file reading, etc. is inside GwfGwfExchange. There we should check what parts are common to both GWF-GWF and GWT-GWT, then we can probably isolate it for reuse.

PR #687, opened 2021-03-31
feat(uzf): new uzf output option - mean water content for all objects

emorway-usgs:
- Adds 'wc' keyword to UZF `OPTIONS` block
- New binary output file records mean water contents for every UZF object
- Includes new autotest that compares MODFLOW6/UZF6 output to MFNWT/UZF1 output recorded in the MT3D linker file (written by LMT package)
- Updates UZF definition file (.dfn) for supporting new binary output file with flopy
- Records these updates in the releaseNotes.tex file

PR #695, opened 2021-04-08
feat(uzf): new uzf output file, mean water content for each uzf object

emorway-usgs:
Redo of #687


emorway-usgs:
sticking with #687

Issue #534, opened 2020-08-01
LAK obs output file not reporting gw exchange


emorway-usgs:
I recently noticed that the reported inflows and outflows for the LAK package's observation output were not balancing.  Fortunately everything balances according to the list file.  It looks like lak.obs output is missing the calculated gwexchg terms.  For example, the output in the listing file suggests that for my particular simulation, the gwexchng term is quite large:

![Lak_bal_1](https://user-images.githubusercontent.com/3236576/89104732-591db980-d3d0-11ea-9ca5-a1e8ebca8dac.png)
![Lak_bal_2](https://user-images.githubusercontent.com/3236576/89104733-5cb14080-d3d0-11ea-9de5-e7e50f316add.png)

But as you can see here the modelfile.lakobs is writing 0's instead:
![lakobs](https://user-images.githubusercontent.com/3236576/89104753-7b173c00-d3d0-11ea-894b-089c9a409b3c.png)

Coincidentally, I'm also noticing that the `TO_MVR` in the listing file doesn't match what's written under the `LAKMVR_OUT` column heading in the LAK observation output file.

I should note that I just pulled all the latest on the develop branch, so this reflects the most recent code changes on that branch

emorway-usgs:
Here is the model I used that should recreate the issue.  One additional note is that I'm on the transport branch, so perhaps there are some updates in `develop` that need to be ported into `transport`?
[lakBudget_issue.zip](https://github.com/MODFLOW-USGS/modflow6/files/5017090/lakBudget_issue.zip)

langevin-usgs:
The transport branch is no longer under development.  Transport was merged into develop.  We are going to delete the transport branch.

emorway-usgs:
I can confirm that the issue is still present with all the latest developments on the `develop` branch

langevin-usgs:
So it turns out that this is actually an input error.  If you want a sum of all groundwater leakage to a single lake, then you need to specify the boundname for the lake, not the lake number, in the obs input file.  In your case, the lake is called "reservoir1", so if we use that in the lake obs input file as

![image](https://user-images.githubusercontent.com/3372403/89425875-1bd26800-d6ff-11ea-8c90-58bc879518ce.png)

Then the observation value matches the value shown in the list file.

![image](https://user-images.githubusercontent.com/3372403/89425753-ffcec680-d6fe-11ea-91b0-e81d90e9d690.png)

[...]

PR #710, opened 2021-04-15
fix(lakobs): throw error if ID2 missing when ID is integer

emorway-usgs:
- Addresses #534
- Includes autotest that generates error message, asserts that error occurred, then fixes error in lak.obs input file and reruns model successfully
- Release notes updated

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3lak8.f90
+++ b/src/Model/GroundWaterFlow/gwf3lak8.f90
@@ -5098,6 +5098,14 @@ contains
       if (obsrv%ObsTypeId=='LAK' .or. obsrv%ObsTypeId=='CONDUCTANCE' .or. &
           obsrv%ObsTypeId=='WETTED-AREA') then
         call extract_idnum_or_bndname(strng, icol, istart, istop, nn2, bndname)
+        if (len_trim(bndName) < 1 .and. nn2 < 0) then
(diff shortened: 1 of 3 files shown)


src/Model/GroundWaterFlow/gwf3lak8.f90
```diff
@@ -5098,6 +5098,14 @@ subroutine lak_process_obsID(obsrv, dis, inunitobs, iout)
       if (obsrv%ObsTypeId=='LAK' .or. obsrv%ObsTypeId=='CONDUCTANCE' .or. &
           obsrv%ObsTypeId=='WETTED-AREA') then
         call extract_idnum_or_bndname(strng, icol, istart, istop, nn2, bndname)
+        if (len_trim(bndName) < 1 .and. nn2 < 0) then
```
langevin-usgs, line 5101: @emorway-usgs, thanks for this!  By any chance, can you quickly check MAW and see if the same situation exists there?

emorway-usgs: @langevin-usgs Yes, I'll take a look at MAW and the other advanced packages.

emorway-usgs: @langevin-usgs Attempting to bring closure to this review comment with #715

PR #718, opened 2021-04-26
Messaging facility has a state that requires initialization.

mjr-deltares:
Running two simulations, one after the other, with the same dll was causing exceptions. This should fix it. (NB: we didn't have that at hand before the API)

(cherry picked from commit 0fec1a772c8dd82d5c45a9d387bdff2c6c6a6e92)

### DIFF
--- a/src/Utilities/Message.f90
+++ b/src/Utilities/Message.f90
@@ -25,6 +25,7 @@ module MessageModule
     
     contains
   
+    procedure :: init_message
     procedure :: count_message
     procedure :: set_max_message
     procedure :: store_message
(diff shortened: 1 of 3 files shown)


mjr-deltares:
Hi @langevin-usgs, could you review this and pull it in when you agree with the changes?

langevin-usgs (commented):
I can understand why this is necessary when a dll is called more than once.  My only question is what happens to the message array?  Should the init_message() method check to see if it's allocated and deallocate if so?  Otherwise, won't the message array continue to grow with multiple dll calls?  Presumably this is not an issue and you've thought about that.  If so, then I say it's good to go.

mjr-deltares:
Hi @langevin-usgs, the message array is currently only allocated when used, and then expanded when required. This happens in store_message() using ExpandArray. The deallocate function is always called on the message type: in Mf6Finalize we call final_message() which cleans up. Of course, the API user should always call the bmi finalize() before initializing a new simulation. I will pull this in.

Issue #699, opened 2021-04-09
Precisely define UZF water content for consistency


langevin-usgs:
UZF reports water content with depth as an observation, and with #687 water content can be reported for each uzf cell.  The definition of water content should be revisited for the case where the uzf cell is beneath the water table.  In this case, there is no unsaturated zone, so we should determine whether the water content should be zero or whether it should be the pore volume.  The water content with depth observation reports pore volume  in this case, but #687 will be report zero in its current form.  Whatever is done, the definition should probably be consistent with what UZF will report for its storage term.

emorway-usgs:
> In this case, there is no unsaturated zone, so we should determine whether the water content should be zero or whether it should be the pore volume

@langevin-usgs For fully saturated cells (water table > cell top elev), I tend to think it makes more sense to have the reported mean water content (activated with `WATER_CONTENT  FILEOUT  <filename.bin>` in the UZF input file's OPTIONS block) for these cells equal to the saturated water content, rather than a value of 0.0.  Currently, MF 6 is writing a value of 0.0 to the binary water content output file. Setting mean water contents to saturated water content would be (1) more intuitive (in my humble opinion) and (2) would match what MF 6 reports when using the specified depth option in the OBS6 approach (i.e., `OBS6 FILEIN <obs6_filename>`).

> Whatever is done, the definition should probably be consistent with what UZF will report for its storage term.

I don't think UZF currently reports total unsaturated storage.  For example, consider this table generated using the PRINT_FLOWS option in the UZF input file for one of the autotests:

```
UZF_0 PACKAGE - SUMMARY OF FLOWS FOR EACH CONTROL VOLUME   PERIOD      6   STEP        1
------------------------------------------------------------------------------------------
  NUMBER   CELLID      INFLOW      OUTFLOW        GWF      INFILTRATION          STORAGE 
------------------------------------------------------------------------------------------
    1      (1,1,2)   0.0000      -7.37329E-03  0.0000       1.00000E-03   ...  1.67670E-02 
    2      (1,1,3)   0.0000      -7.37329E-03  0.0000       1.00000E-03   ...  1.67670E-02 
    3      (1,1,4)   0.0000      -7.37329E-03  0.0000       1.00000E-03   ...  1.67670E-02 
    4      (1,1,5)   0.0000      -7.37329E-03  0.0000       1.00000E-03   ...  1.67670E-02 
[...]

PR #730, opened 2021-05-20
fix(uzf): reported water content for cells below water table set to thts

emorway-usgs:
Addresses #699


langevin-usgs:
If we calculate the change in storage based on these water content definitions, would we calculate the same change in storage that is reported in the UZF budget?

emorway-usgs:
@langevin-usgs I checked it two ways, filling a UZF object and draining a UZF object.  When filling, the change in the reported mean water content from some value less than saturation up to saturation matches the reported change in storage (this value was gleaned from the `PRINT_FLOWS` table is reported as a rate and therefore needed to be multiplied by the time step length).  For a UZF object that initially starts a time step at saturation and ends at some value less than saturation, the requested calculation works when using the saturated water content.  

For example, the included test has a saturated water content of 0.35, and this value is what is recorded in the binary `water-content` output file.  In stress period 2, the bottom layer (row 1, column 2) went from fully submerged to a water table that was 1.9991 m below the cell top.  The reported mean water content for this same cell dropped from 0.35 (again, this is the saturated water content) to 0.2497, which, for a 30 day stress period, equates to a volume of (**0.35** - 0.2497) * 1.9991 = 0.20050973

The reported change in storage (recorded as a rate) from the `PRINT_FLOWS` table generated by the UZF package for the same stress period (SP 2) was 6.64774E-03.  Multiplying by the length of the stress period yields: 6.64774E-03 * 30 = 0.1994322

This leaves a small discrepancy of ~0.00107753 m^3 (note that the `PRINT_FLOWS` table does record a small percent discrepancy of 1.91e-6 for the investigated cell).  If we were to instead use a water content of 0.0 (current output prior to this PR) instead of 0.35 in the first calculation above (see the value in bold type), we'd get a much larger error.

emorway-usgs:
Superseded by #739

PR #739, opened 2021-06-10
fix(uzf): reported water content for cells below water table set to thts

emorway-usgs:
Redo of #730.  #730 was polluted with merge commits


langevin-usgs:

PR #759, opened 2021-07-17
refactor(SimModule,BlockParser): encapsulate ustop in StoreErrorUnit

jdhughes-usgs:
Also add optional call to ustop in store_error, store_error_unit, and
store_error_filename.


langevin-usgs (approved):
Looks good.

PR #768, opened 2021-08-18
Combined TVK and TVS feature additions (time-varying hydraulic conductivities, specific storage and specific yield)

damianmerrick:
@langevin-usgs As requested, this PR on develop replaces the earlier separate TVK and TVS PRs on feat-tvk and feat-tvs.

TVK
Add time-varying K support to NPF, as TVK sub-package available through additional file include in NPF OPTIONS block.
Add nodekchange, kchangeper and kchangestp variables to NPF type to allow other affected packages to query for K changes in ad routines. No changes to affected packages have been included in this PR (affected packages: GWF-GWF exchange, HFB, LAK, MAW).

TVS
Add time-varying SS and SY support to STO, as TVS sub-package available through additional file include in STO OPTIONS block.
Generalize STO formulate and budget routines to correctly integrate changes in stored water volume due to changes in SS and/or SY, when new STO type variable integratechanges is non-zero. This is applied by default when TVS is active in a model, and not applied otherwise unless the integratechanges variable is manually set non-zero (e.g. via BMI).

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3sto8.f90
+++ b/src/Model/GroundWaterFlow/gwf3sto8.f90
@@ -19,6 +19,8 @@ module GwfStoModule
   use NumericalPackageModule, only: NumericalPackageType
   use BlockParserModule, only: BlockParserType
   use GwfStorageUtilsModule, only: SsCapacity, SyCapacity
+  use InputOutputModule, only: GetUnit, openfile
+  use TvsModule, only: TvsType, tvs_cr
 
   implicit none
   public :: GwfStoType, sto_cr
@@ -39,6 +41,11 @@ module GwfStoModule
     real(DP), dimension(:), pointer, contiguous      :: strgsy => null()         !< vector of specific yield rates
     integer(I4B), dimension(:), pointer, contiguous  :: ibound => null()         !< pointer to model ibound
     real(DP), pointer                                :: satomega => null()       !< newton-raphson saturation omega
+    integer(I4B), pointer                                :: integratechanges => null() !< indicates if mid-simulation ss and sy changes should be integrated via an additional matrix formulation term
+    integer(I4B), pointer                                :: intvs => null()            !< TVS (time-varying storage) unit number (0 if unused)
+    type(TvsType), pointer                               :: tvs => null()              !< TVS object
+    real(DP), dimension(:), pointer, contiguous, private :: oldss => null()            !< previous time step specific storage
+    real(DP), dimension(:), pointer, contiguous, private :: oldsy => null()            !< previous time step specific yield
   contains
     procedure :: sto_ar
     procedure :: sto_rp
@@ -54,6 +61,7 @@ module GwfStoModule
     !procedure, private :: register_handlers
     procedure, private :: read_options
     procedure, private :: read_data
+    procedure, private :: save_old_ss_sy
   end type
 
 contains
@@ -131,6 +139,11 @@ contains
     ! -- read the data block
     call this%read_data()
     !
+    ! -- TVS
+    if (this%intvs /= 0) then
+      call this%tvs%ar(this%dis)
+    end if
+    !
     ! -- return
     return
   end subroutine sto_ar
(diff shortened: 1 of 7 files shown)


src/Model/GroundWaterFlow/gwf3sto8.f90
```diff
@@ -312,22 +350,35 @@ subroutine sto_fc(this, kiter, hold, hnew, njasln, amat, idxglo, rhs)
       !
       ! -- set saturation used for ss
       ss_sat1 = snnew
-      ssh0 = hold(n)
-      ssh1 = DZERO
```
jdhughes-usgs, line 316: The volume of water released or added to specific storage needs to be based on the top of the cells. I am not sure this is correct.

src/Model/GroundWaterFlow/gwf3sto8.f90
```diff
@@ -339,15 +390,15 @@ subroutine sto_fc(this, kiter, hold, hnew, njasln, amat, idxglo, rhs)
           if (this%iconf_ss == 0) then
             zold = bt + DHALF * tthk * snold
             znew = bt + DHALF * tthk * snnew
-            rhsterm = -rho1 * (snold * (hold(n) - zold) + snnew * znew)
+            rhsterm = -rho1old * snold * (hold(n) - zold) - rho1 * snnew * znew
           else
-            rhsterm = -rho1 * ssh0 + rho1 * ssh1
+            rhsterm = -rho1old * ssh0 - rho1 * ssh1
```
jdhughes-usgs, line 395: If the above is not correct the sign of the second term need to be the same as before.

jdhughes-usgs, line 395: This is ok. 

Taking another look at the way I implemented this option I think it is confusing. 

I think the variable names should be modified. I think it would be easier to follow if the variable `ssh0` was changed to `ssb0` and set equal to `hold - top` when `hold > tp` and `0` otherwise. Similarly, `ssh1` should probably just be `tp` and `rho1` should be multiplied by `ss_sat1` like `aterm` is. So it would look like

`rhsterm = -rho1old * ssb0 - rho1 * ss_sat1 * tp`

@damianmerrick I can take care of this refactoring after the PR gets merged. Or feel free to make the change.

src/Model/GroundWaterFlow/gwf3sto8.f90
```diff
@@ -312,22 +350,35 @@ subroutine sto_fc(this, kiter, hold, hnew, njasln, amat, idxglo, rhs)
...
       ! -- set saturation used for ss
       ss_sat1 = snnew
-      ssh0 = hold(n)
-      ssh1 = DZERO
+      ssh0 = hold(n) - tp
+      ssh1 = tp
       if (this%iconf_ss /= 0) then
         if (snold < DONE) then
-          ssh0 = tp
+          ssh0 = DZERO
```
jdhughes-usgs, line 357: I would change the variable to `ssb0` (see detailed explanation add at line 395)

@damianmerrick I can take care of this refactoring after the PR is merged.

src/Model/GroundWaterFlow/gwf3sto8.f90
```diff
@@ -312,22 +350,35 @@ subroutine sto_fc(this, kiter, hold, hnew, njasln, amat, idxglo, rhs)
       !
       ! -- set saturation used for ss
       ss_sat1 = snnew
-      ssh0 = hold(n)
-      ssh1 = DZERO
+      ssh0 = hold(n) - tp
```
jdhughes-usgs, line 353: Refactor to

`ssb0 = hold(n) - tp`

jdhughes-usgs (approved):

src/Model/GroundWaterFlow/gwf3sto8.f90
damianmerrick: Yes, I agree. It will make more sense following your suggested changes.

@jdhughes-usgs I will leave the refactoring to you post-PR if you are happy to do that. Thanks!

PR #772, opened 2021-08-25
Fix and extension of gwt-fmi advancement of bfr and hfr

aprovost-usgs:


aprovost-usgs:
Withdrawing and will resubmit

PR #775, opened 2021-08-25
refactor(sto): refactor confined ss option formulation to clarify eq.

jdhughes-dev:
Also refactor allocation of oldss and oldsy variables to use the
memorymanager to allocate and deallocate storage.


damianmerrick:
@jdhughes-usgs I'm a bit late to the party, but this looks good. Nice to see the fc and cq logic pulled out into a single place. Thanks for doing the refactor!

PR #779, opened 2021-08-30
TVK and TVS tests

damianmerrick:
Added three new tests: gwf_npf_tvk02, gwf_npf_tvk03 and gwf_sto_tvs01.

@langevin-usgs Does something further need to be done to add these to the set of CI build tests, or will they be picked up automatically?

@jdhughes-usgs While writing the TVS tests, I noticed that part of the recent refactor was causing TVS to crash, as the memory allocation for OLDSS and OLDSY was moved to occur prior to when the integratechanges flag is set by TVS. I have kept the modification to use mem_allocate and mem_deallocate, but changed where and when it occurs back to using a "lazy" allocation approach, which I believe is a better option here. In addition to supporting the existing order of operations on TVS initialisation, it allows the integratechanges flag potentially to be changed at any point in the simulation and result in the expected behaviour (e.g. via Python script), while ensuring that additional memory for the OLDSS and OLDSY variables is allocated only if needed. My changes also guarantee deallocation when the arrays have been used, which relying solely on the integratechanges flag would miss if the flag is changed to zero mid-simulation.

### DIFF
--- a/autotest/test_gwf_npf_tvk02.py
+++ b/autotest/test_gwf_npf_tvk02.py
@@ -1,0 +1,272 @@
+import os
+import sys
+import numpy as np
+
+try:
+    import pymake
+except:
+    msg = "Error. Pymake package is not available.\n"
+    msg += "Try installing using the following command:\n"
+    msg += " pip install https://github.com/modflowpy/pymake/zipball/master"
+    raise Exception(msg)
+
+try:
+    import flopy
+except:
+    msg = "Error. FloPy package is not available.\n"
+    msg += "Try installing using the following command:\n"
+    msg += " pip install flopy"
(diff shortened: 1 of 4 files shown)


langevin-usgs (approved):
Thanks for this @damianmerrick.  I'll go ahead and merge this in.  Tests look good.  Regarding the memory allocate issue, I think we will want to follow our general approach, which for your case would be to mem_allocate() with a size of 1 and if those arrays are needed, then we mem_reallocate().  We typically do it this way for all of our class member variables so that none of them are ever null (which can cause problems), and then we can just do a regular mem_deallocate() without having to check association.  Clearly not a big deal other than it's our convention.  So whoever is in here next can go ahead and make that change.

PR #781, opened 2021-09-01
Use meson as build system

Hofer-Julian:
Motivation:
- Easy to add dependencies
- Faster (incremental) compilation
- Easy to add tests and doxygen generation
- Readable build files

Things which should be handled in follow-up PRs:
- Add existing tests as meson tests
- If possible, remove nosetests and pymake dependency of tests
- Adapt workflow files to only use meson wherever possible
- Simplify CI helper scripts
- Get rid of number of compiler warnings either by silencing some of them or fixing them


Hofer-Julian:
Note: CI fails because I've renamed compilerversion.f90 to compilerversion.F90 to signal to the compiler that it should be pre-processed 

https://riptutorial.com/fortran/example/31322/extensions-and-meanings

langevin-usgs:
Hey @Hofer-Julian, is the uppercase .F90 extension to indicate preprocessing used broadly?  I've never heard of that convention before, and don't see any description of it anywhere other than the riptutorial in the link you provided.  Just curious if you've found this to be widely used.  It makes sense to me, and I like the idea of adopting a preprocess file naming convention.

Hofer-Julian:
> 
> 
> Hey @Hofer-Julian, is the uppercase .F90 extension to indicate preprocessing used broadly? I've never heard of that convention before, and don't see any description of it anywhere other than the riptutorial in the link you provided. Just curious if you've found this to be widely used. It makes sense to me, and I like the idea of adopting a preprocess file naming convention.

It is only used by gfortran (as can be seen [here](http://fortranwiki.org/fortran/show/File+extensions) and [here](https://gcc.gnu.org/onlinedocs/gfortran/Preprocessing-Options.html)).
I am actually surprised that it worked for you without using uppercase F.

With intel fortran I have to set an additional compiler flag.
Considering that gfortran is one of the two compilers we support, I feel it might still be good to follow this convention.

Hofer-Julian:
@langevin-usgs and @jdhughes-usgs this PR is now ready to be reviewed.
This will allow @mjr-deltares and me to start to experiment with MPI + Modflow 6 on multiple platforms and with multiple MPI implementations.
I've also reduced the number of `meson.build` files as requested by you.
A few tasks are still left to do before I'd consider the move to meson to be done. I've listed them above.

Hofer-Julian:
I have also added two workflow files, which only check if building Modflow 6 with all Linux/macOS/Windows and gfortran/ifort permutations work fine, but not if tests are green.

jdhughes-usgs:
@Hofer-Julian defmacro.f90 also has preprocessor directives. Does this need the .F90 extension also?

Hofer-Julian:
> 
> 
> @Hofer-Julian defmacro.f90 also has preprocessor directives. Does this need the .F90 extension also?

That is a good question, and I honestly do not know why it compiles fine.
It seems as if preprocessor directives like `#ifdef` are processed fine without .F90 extension.
What does make trouble is something like `__DATE__` which is only present in `compilerversion.F90`.

To be sure we don't depend on buggy behavior of gfortran, I have renamed `defmacro.f90` as well

Hofer-Julian:
Thinking more about it, if the the extension is only taken into account for certain preprocessor directives, maybe we shouldn't rely on it at all.

I'd suggest to continue to using .f90 for everything and add the `-cpp` flag for gfortran similar to how we handle ifort.
What do you say @jdhughes-usgs and @langevin-usgs?

jdhughes-usgs (approved):

jdhughes-usgs:
@Hofer-Julian I am fine with using *.F90 for files that need to be preprocessed. Might be better to use this standard instead of preprocessing every file. Compile times could become an issue in the future.

Hofer-Julian:
Sure, if that's the case, then this PR is ready to be merged from my side :)

PR #793, opened 2021-09-23
Move to pytest

Hofer-Julian:


Hofer-Julian:
This pull request is now ready to be reviewed.

jdhughes-usgs:
@Hofer-Julian could you add a few switches to pytest so we get similar output to before. Here is what we use for pymake:

`Run pytest -v -m regression --durations=0 --cov=pymake --cov-report=xml autotest/`

Hofer-Julian:
@jdhughes-usgs I've added `-v --durations=0` to the pytest calls.

PR #798, opened 2021-09-29
Extend meson with installation and vscode tasks

Hofer-Julian:
This is a draft PR and not ready to be merged yet.
It:
- allows to install binaries via meson. That is nice to have, but also needed to run tests with binaries compiled with meson.
- add vscode tasks based on meson.

Things left to do:

- [ ] Extend meson github workflow to run tests
- [ ] Test vscode tasks on Linux
- [ ] Attempt to avoid asking user to hardcode conda path for vscode tasks
- [ ] Modify README to explain new workflow with vscode tasks

PR #801, opened 2021-10-05
Debug large tests

Hofer-Julian:


Hofer-Julian:
The PR proved that only a single line needs to be changed.
Closing it in favor of https://github.com/MODFLOW-USGS/modflow6/pull/798 where I've fixed this

PR #812, opened 2021-10-16
fix(urword,obs): correctly parse boundnames with double quotes

jdhughes-usgs:


jdhughes-usgs:
@langevin-usgs There are some discrepancies in the way the water-content in UZF is implemented that we should discuss. I also don't think the testing is sufficient for observations that use boundnames.

Also not sure the parsing of observation ID2 is done how we want it in observations.

PR #815, opened 2021-10-19
Remove multiple model build-up with pytest

Hofer-Julian:
This pull request removes multiple model build-up with pytest.

There are still some instances of data race.
Few enough to run tests in parallel locally, but too frequent to enable it in the CI just yet.


autotest/framework.py
```diff
@@ -46,6 +46,27 @@ def build_mf6_models(self, build_function, idx, exdir):
         if regression is not None:
             regression.write_simulation()
 
+    def build_mf6_models_legacy(self, build_function, idx, exdir):
```
Hofer-Julian, line 49: @jdhughes-usgs you mentioned that `write_input` is relevant for older versions of MODFLOW.
This is why I added the `legacy` suffix.
I am open to better naming suggestions

PR #817, opened 2021-10-20
feat(sfr): add option to vary cross-section roughness

jdhughes-usgs:


jdhughes-usgs:
Closing. Will resubmit a clean PR

PR #830, opened 2021-11-03
Fix failure if bin dir does not exist

Hofer-Julian:
While working on related tasks I've noticed that the vscode task fails if there is no bin folder


langevin-usgs (approved):

langevin-usgs:
Accepting PR.  The failure is due to the intermittent known failure of test001h_rch_array4.

PR #832, opened 2021-11-03
feat(sfr): n-point cross-section manning's variations

jdhughes-usgs:


autotest/test_gwf_disu01.py
```diff
@@ -131,7 +134,7 @@ def test_disu_idomain_simple():
     budobj = flopy.utils.CellBudgetFile(fname, precision="double")
     flowja = budobj.get_data(text="FLOW-JA-FACE")[0].flatten()
     assert flowja.shape[0] == 126
-
+    shutil.rmtree(ws, ignore_errors=True)
```
langevin-usgs, line 137: Should we think about checking a --keep argument before removing?

langevin-usgs (commented):
I think this all looks really good.  I think the text descriptions in the release notes and IO guide could use more information.  I started to add that here, but I think it would be better to do it on a subsequent PR, which I am happy to do.  I had a real mental block with the word "depth" for the cross sections, but now understand the logic.  I think it would be worth adding a figure, like the one in the SFR1 manual to the IO guide to clearly show how the cross section data is translated into the model.  I can do that on the subsequent PR.

autotest/test_gwf_disu01.py
jdhughes-usgs: Added a general function to look for the `--keep` argument in `framework.py` and added check in `autotest/test_gwf_disu01.py`

doc/mf6io/mf6ivar/examples/utl-sfrtab-example.dat
```diff
@@ -1,19 +1,19 @@
...
+        0.1     1.0         10.0
+        0.2     1.0          1.0
+        0.3     0.0          1.0
+        0.4     0.0          1.0
+        0.5     0.0          1.0
+        0.6     0.0          1.0
+        0.7     0.0          1.0
+        0.8     1.0         10.0
+        0.9     1.0         10.0
+        1.0     1.0         10.0
```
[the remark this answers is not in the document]
jdhughes-usgs: added a comment

PR #839, opened 2021-11-16
refactor(makefile): add missing files and restore gfortran compat

rodrperezi:
- compilerversion.F90 and defmacro.F90 file extensions to lowercase
 - add src/Model/Connection missing files

### DIFF
--- a/make/makedefaults
+++ b/make/makedefaults
@@ -1,4 +1,4 @@
-# makedefaults created on 2021-07-30 14:34:49.615882
+# makedefaults created on 2021-11-16 12:14:02.967101
 # by pymake (version 1.2.1) for the 'mf6' executable 
 # using the 'gfortran' fortran compiler(s).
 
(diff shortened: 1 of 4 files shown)


langevin-usgs:
Hi @rodrperezi, thanks for your PR.  The capitalized .F90 file suffix is an indicator that the file must be preprocessed, so I don't think we want to change that.  And we are slowly working on a way to ensure the makefile on the develop branch is consistent with the present state of the source code.  We are planning to do that by adding a makefile compile test to a github workflow, which will force contributors to update the makefile is source files as added or removed.  If you undo the .F90 changes, we will look at the PR.  Hopefully the makefile that you are committing was generated with pymake.

rodrperezi:
Thanks Chris, the truth is that I made the changes by hand directly on make/makefile until compilation. The issue was that files at make/ were not consistent with those at pymake. Now I see that with pymake updating the makefiles is pretty straightforward.
I've restored the file extensions and updated makefiles with the following command from pymake/ folder
`
python -m pymake ../src ../bin/mf6 -mc --subdirs -fc gfortran -mf -ff='-O2' && cp makedefaults ../make/ && cp makefile ../make/
`

jdhughes-usgs:
@rodrperezi I have added a test to our automated CI that builds mf6, zbud6, and mf5to6 using make with gfortran and ifort. That way we can make sure the makefiles are current. 

The script `distribution/build_makefiles.py` can be used to rebuild the makefiles if new source files are added or renamed.

I am closing this PR since it is out of date. Thanks for the help.

PR #846, opened 2021-11-24
refactor(kind): use ISO_FORTRAN_ENV for integer, logical, and real types

jdhughes-usgs:


langevin-usgs (approved):
Looks good to me.  I like it.  Much clearer.

PR #848, opened 2021-11-30
refactor(CompilerVersion): use iso_fortran_env for compiler information

jdhughes-usgs:
Refactored write_message() in GenericUtilitiesModule to handle messages
with and without counters in a more robust and clear manner. Refactored
write_centered() to handle strings that exceed passed linelen and
automatically convert to multi-lined centered strings. Added compiler
options (--COMPILER-OPT) command line option to output compiler options
to the screen. Compiler options are also written to the listing file.


langevin-usgs (approved):
Code looks good.  The failure looks to be minor because it is just checking an expected error message, which will be slightly different based on your changes.  All these makefile changes reiterate the need to be clean up the dag so that it gives the same file order each time.  And we might think about removing the date and time stamps from the makefiles, as that information would already be a part of the file properties and we don't want to version those changes.

PR #862, opened 2022-01-20
fix(run_command): use process.returncode for ierr

mwtoews:
The Python function `run_command` is used in a few places, and should use the process [returncode](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode) for ierr.

This fixes instances like [this one](https://github.com/modflowpy/flopy/runs/4877208422?check_suite_focus=true#step:10:215) which shows:
```
running...createpackages.py
successfully ran...createpackages.py
```
when really the command was not successful due to an unrelated error in the script.

### DIFF
--- a/autotest/build_mfio_tex.py
+++ b/autotest/build_mfio_tex.py
@@ -171,14 +171,13 @@ def delete_files(files, pth, allow_failure=False):
 
 
 def run_command(argv, pth, timeout=10):
-    buff = ""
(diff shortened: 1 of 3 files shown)


langevin-usgs:
This looks like a nice addition, @mwtoews.  Thanks!  The failure here has been fixed separately, so I went ahead and pulled it in.

PR #877, opened 2022-02-02
Feat/var meta

jmccreight:
This is really a shoe horn to get a human-readable format (e.g. a flattened csv in the source) into a new memory manager field for attributes (using keys/hashes to ensure some sanity, at least for me).

Currently the the attribute keys are a constant and are not stored in the memory manager (only the values are). I've only started with 3 attributes: "variable name", "long name", and "units". This can easily be changed but gets back to the idea that this is currently the same for all Attrs instances.

In theory the Attribute object is usable for all variables and subclassable for special cases. Here I've only implemented for scalar variables in Numerical Solution. 

The bmi interface has been extended to allow access to the variable metadata: `get_attrs_item_count`, `get_attrs_keys` (not-variable specific, as mentioned above), and `get_var_attrs`. It can do something like this:

image

Forthcoming draft PR on modflow API for that.


jmccreight:
I did remove trailing whitespace, you may want to review with the equivalent of -w.

I did a small set of tests locally before pushing. So, I'll take a look at what I broke.

langevin-usgs:
Closing this PR as we are headed a different direction with the input data model.

PR #878, opened 2022-02-02
perf: flush files after writing data for a time step

jdhughes-usgs:
Remove unused travis subdirectory and sanitize autotest comments that
list specific CI infrastructure (Travis, GitHub Actions).


langevin-usgs (approved):
The granularity of the implementation is clearly efficient.  I think we discussed this -- how is the efficiency affected by a single flush statement, versus individual flush statements throughout?  

Also, I noticed on the gnu page for the flush statement that it says: "This does not guarantee that the data is committed to disk."  There is also mention that an additional fsync call is required on POSIX systems.

Maybe in the future, we should implement our own flush subroutine, which calls the flush statement, and call our flush subroutine from the different locations in the code base.  That way it's all containerized in one spot and we could check for flush errors, or deal with fsync (if ever needed).

PR #888, opened 2022-02-09
fix(gwfexg): implement, fix, and test observations for gwf-gwf exchanges

langevin-usgs:
* gwf-gwf supports boundnames, but they were not described in docs
* interface model did not fully implement gwf-gwf exchanges, but they do now
* test_gwf_ifmod_xt3d01.py was upgraded to fully test gwf-gwf observations
* mf6io.pdf was updated to better describe gwf-gwf observations, which can be a boundname or an exchange index
* gwfgwfexchange was updated to use simvals throughout, instead of reproducing the flow calculation


mjr-deltares:
@langevin-usgs : this looks really nice!

PR #921, opened 2022-03-07
Specify FC env var in CI

Hofer-Julian:
Not strictly needed, but why shouldn't we be explicit about it.

(seems like my yaml formatter disagrees with your formatting, sorry for the noise)

### DIFF
--- a/.github/workflows/ci-check-warnings-gfortran.yml
+++ b/.github/workflows/ci-check-warnings-gfortran.yml
@@ -8,6 +8,9 @@ on:
     branches:
       - develop
 
+env:
+  FC: gfortran
+
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


Hofer-Julian:
Is there something blocking this PR?

langevin-usgs:
Hey @Hofer-Julian, I was meaning to task you about this.  There are a bunch of white space changes in your PR, as you indicated.  Is there a standard indent for yml files or should we adopt a standard?  If we all handle this differently, then we lose the ability to track version changes.

Hofer-Julian:
@langevin-usgs  yes agreed, if we format files we should also agree on a standard and enforce it in the CI.

I removed the whitespace changes.

PR #925, opened 2022-03-16
feat(maw-mfrcsv): add option to write MAW flow reduction to csv

kzeiler:
add option to write MAW extraction and injection flow reductions to csv.


langevin-usgs (approved):
Thanks @kzeiler.  This looks good.

PR #934, opened 2022-04-27
fix(GwfGwtExchange): update indexing variable when MVR is active

emorway-usgs:
- exg_df() needs code similar to what is in initialize_gwfterms_from_gwfbndlist() in FMI.
- Two new autotests patterned after the test_gwt_ssm01FMI autotest drop RIV, add SFR, and add MVR and MVT to transfer water and solute from a WEL to SFR.  This condition resulted in an error message that _not_ should have been printed to the screen.  However, the error condition only happened when the flow and transport simulations were run simultaneously (the 2nd of the two new autotests).  The first autotest runs the models separately and has been added for completeness (tests the same configuration that runs GWF & GWT simultaneously).  Both autotests use auxiliary variables and mover in the WEL package, the condition condition that previously triggered an error msg that shouldn't have been triggered (see below).
- Ran `black` on both autotests.

ssm_error


langevin-usgs:
@emorway-usgs, thanks for doing this.  I was able to reproduce your problem.  Couple of things that would be worth considering.

1.  I think the calls to set_auxname (including your new ones) in GwfGwtExchange.f90 can all be moved into the subroutine initialize_gwfterms_from_gwfbndlist() in gwt1fmi.f90 instead.  I implemented this and worked with you new test and all the other tests passed as well.  I think this is a better place for this assignment.  The new part of that routine should look like the following:

```
    !
    ! -- Assign values in the fmi package
    iterm = 1
    do ip = 1, ngwfpack
      !
      ! -- set and store names
      packobj => GetBndFromList(this%gwfbndlist, ip)
      budtxt = adjustl(packobj%text)
      call this%gwfpackages(iterm)%set_name(packobj%packName, budtxt)
      this%flowpacknamearray(iterm) = packobj%packName
      call this%gwfpackages(iterm)%set_auxname(packobj%naux,         &
                                               packobj%auxname)
      iterm = iterm + 1
      !
      ! -- if this package has a mover associated with it, then add another
      !    term that corresponds to the mover flows
      imover = packobj%imover
      if (packobj%isadvpak /= 0) imover = 0
      if (imover /= 0) then
        budtxt = trim(adjustl(packobj%text)) // '-TO-MVR'
        call this%gwfpackages(iterm)%set_name(packobj%packName, budtxt)
        this%flowpacknamearray(iterm) = packobj%packName
        call this%gwfpackages(iterm)%set_auxname(packobj%naux,         &
                                                 packobj%auxname)
        this%igwfmvrterm(iterm) = 1
        iterm = iterm + 1
      end if
    end do
    return
  end subroutine initialize_gwfterms_from_gwfbndlist
```
 
2.  Make sure you remove the set_auxname calls in GwfGwtExchange.f90 and any other variables that are no longer needed.

3.  The names for your tests should be changed.  The first one should be called ssm06fmi and the second one should just be called ssm06.  These are new tests, and we only want to put fmi in the source file name if flow and transport are run separately as two separate models.  


Does this seem reasonable?  I think it keeps the transfer of information from gwf to gwt more compartmentalized.  I suspect there were reasons for having those calls in exe_df() but they do not seem to be valid any longer.

emorway-usgs:
@langevin-usgs Thank you for the additional guidance on where to move the `set_auxname` calls to.  Tests renamed.  The two new autotests worked locally with the updated code, have pushed the requested updates in hopes all autotests will pass muster.

autotest/test_gwt_ssm06fmi.py
```diff
@@ -0,0 +1,429 @@
...
+
+    # Load the csv representation of the budget for confirming that the model ran
+    fname = f"{gwtname}.cbc.csv"
+    fname = os.path.join(wst, fname)
+    d0 = np.genfromtxt(fname, names=True, delimiter=",", deletechars="")
+    print(d0.dtype.names)
+    return
+
+
+def test_ssm01fmi():
```
langevin-usgs, line 415: this should be renamed to test_ssm06fmi, which is what pytest will echo to the screen

autotest/test_gwt_ssm06fmi.py
```diff
@@ -0,0 +1,429 @@
...
+        shutil.rmtree(d)
+    return
+
+
+if __name__ == "__main__":
+    # print message
+    print("standalone run of {}".format(os.path.basename(__file__)))
+
+    # run tests
+    test_ssm01fmi()
```
langevin-usgs, line 429: same here

autotest/test_gwt_ssm06.py
```diff
@@ -0,0 +1,392 @@
...
+        shutil.rmtree(d)
+    return
+
+
+if __name__ == "__main__":
+    # print message
+    print("standalone run of {}".format(os.path.basename(__file__)))
+
+    # run tests
+    test_ssm01fmi()
```
langevin-usgs, line 392: rename to test_ssm06 (without the fmi)

autotest/test_gwt_ssm06.py
```diff
@@ -0,0 +1,392 @@
+# Derived from test_gwt_ssm01fmi.py, but drops RIV and adds SFR.
+# See test_gwt_ssm01fmi_b.py for additional detail on what this test is about.
```
langevin-usgs, line 2: comment needs to be updated

langevin-usgs (changes requested):
Few minor last tweaks, otherwise looks good.  Thanks.

autotest/test_gwt_ssm06fmi.py
emorway-usgs: resolved with c4a2a03

autotest/test_gwt_ssm06fmi.py
emorway-usgs: resolved with c4a2a03

PR #936, opened 2022-05-04
add get_version routine to modflow api (xmi)

mjr-deltares:


jdhughes-usgs:
@langevin-usgs any objections to this PR?

langevin-usgs (commented):
This seems fine to me, but I don't think we have a good system in place for assigning a reasonable string to VERSION.  Right now, the develop VERSION is set to '6.4.0 release candidate 03/04/2022'.  That date isn't updated unless we do it manually, and that is normally done right after we make a release.  So the info provided back through bmi may be misleading.

jmccreight:
THat appears hard coded here?
https://github.com/MODFLOW-USGS/modflow6/blob/ec46d6d9e2889da49564f1fa92e9e69cfb683e7d/src/Utilities/version.f90#L18

We did something a bit more dynamic with WRF-Hydro. I'm not sure it is what you are looking for but I can drop a comment here after I dig it up

jmccreight:
In the Makefile we set the contents of a file into an env var and pass it in as flags to the compile. 

https://github.com/NCAR/wrf_hydro_nwm_public/blob/7cf193b9ea3be508895e129d28cda87baa17dd37/trunk/NDHMS/utils/Makefile

In a more extreme case you could imagine passing `hash_var=$(git rev-parse HEAD )` in... and probably also another varible indicating if the repo was clean at the time of the compile.

mjr-deltares:
Good point @langevin-usgs, and what if we strip the date? So the API will return '6.4.0 release candidate'

mjr-deltares:
> In the Makefile we set the contents of a file into an env var and pass it in as flags to the compile.
> 
> https://github.com/NCAR/wrf_hydro_nwm_public/blob/7cf193b9ea3be508895e129d28cda87baa17dd37/trunk/NDHMS/utils/Makefile
> 
> In a more extreme case you could imagine passing `hash_var=$(git rev-parse HEAD )` in... and probably also another varible indicating if the repo was clean at the time of the compile.

Thanks for sharing this James, I like that too (better), maybe discuss this offline?

Hofer-Julian:
For what it's worth, we already maintain the version number within them [`meson.build`](https://github.com/MODFLOW-USGS/modflow6/blob/706739c723dab4532eb1aa82908cf21c1a6c10f7/meson.build#L4) file.
We could just use that, assuming that way of passing versions through command line arguments works for all permutations of compilers and operating systems.

langevin-usgs:
An alternative might be to use or combine with the information in compiler version.f90.  This information about compiler version and build date might be more beneficial for the purpose at hand.  Agreed that it would also be nice to get git commit hash and actual version number (6.3.0) and whether the develop flag is on.

langevin-usgs:
@mjr-deltares, where are we at with this?  Should we merge or discuss one of the alternatives?

mjr-deltares:
So the current setup is that when released, the xmi get_version routine returns "6.x.x", but when in develop mode it will be "6.x.x-dev". @langevin-usgs is that an acceptable first implementation?

langevin-usgs:
That sounds good to me.  We have that nagging windows failure, but I'll go ahead and pull this in anyways.

langevin-usgs (approved):

PR #947, opened 2022-05-27
test(gwf_disv_uzf): New autotest checks uzet and gwet results

emorway-usgs:
- New autotest confirms relative uzet and gwet behavior is as expected
- Autotest was developed on report of actual ET exceeding specified potential ET by @MostafaGomaa93.  However, on that occasion, an input issue was discovered.  An autotest would be helpful, particularly one that runs a model using DISV and UZF.
- Ran 'black'


langevin-usgs (changes requested):
use .output to get obs instead of pandas.  https://flopy.readthedocs.io/en/latest/_notebooks/tutorial01_mf6_output.html

emorway-usgs:
Hopefully the second commit on 5/27 addressed the requested change by removing the pandas dependency and using the more standard output retrieval methods in its place, though it's entirely possible I mis-understood the requested change.  Please don't hesitate to let me if its the latter.

langevin-usgs (changes requested):
An autotest is not an example that we distribute with the distribution, so you should not updated ReleaseNotes.tex.

langevin-usgs (approved):

langevin-usgs (changes requested):
@emorway-usgs, in looking at this again, I see that you have 50 stress periods with 50 time steps each; which means this little test needs 2500 time steps to finish.  Unless all these steps are required to serve the purpose of this test, then it would be good if you could revise this test to run much faster with only 5 to 10 time steps, if possible.

emorway-usgs:
I might be mistaken, but I think the lines in question: 
```
perlen = [10] * 5
nstp = [10] * 5
``` 
results in 5 stress periods which are each 10 days long with 1 time step per day, yielding a model that is 50 time steps total?  In other words, `nstp` looks like `[10, 10, 10, 10, 10]`.

However, the simulation was taking about 4 seconds to run, so I knocked it down to about 2 seconds by using two-day long time steps instead.

runtime

langevin-usgs (approved):

langevin-usgs:
Thanks, @emorway-usgs.  Sorry I mixed up the calculation.  But I guess that broader issue is that we need to make sure that tests are highly focused, especially as we continue to increase the number of them.  So it would be good practice to consider if having 50 time steps is testing more than the same test with 5 or 10 time steps.  I think we are all guilty of this, which is why our tests take so long to run.  We will have to deal with this soon.

PR #963, opened 2022-06-27
style(gwf): apply fprettify formatting

mjreno:
* working toward consistent code formatting

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3.f90
+++ b/src/Model/GroundWaterFlow/gwf3.f90
@@ -1,26 +1,26 @@
 module GwfModule
 
-  use KindModule,                  only: DP, I4B
-  use InputOutputModule,           only: ParseLine, upcase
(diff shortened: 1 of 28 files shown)


jdhughes-usgs:
@mjreno The windows failure is a result of GitHub Actions using a different version of mingw-64. They are using a MYSYS2 version now. We could fix by switching the workflow from using `windows-latest` to `windows-2019`. The better fix would be to install MYSYS2 on windows and see what the actual issue is and resolve it (if possible). Is this something you are interested in tackling??

mjreno:
@jdhughes-usgs Thanks for the explanation, I was wondering as I hadn't seen this failure before last week.  Yes I will look into it and try to identify the issue.  Do you want to switch the workflow temporarily to avoid the failure or hold off?

langevin-usgs (approved):

PR #967, opened 2022-06-29
ci: update scripts to use meson to build instead of pymake

jdhughes-usgs:


autotest/build_exes.py
```diff
@@ -60,131 +80,57 @@ def create_dir(pth):
...
+            + f"--bindir={os.path.abspath(libdir)} "
+            + f"--libdir={os.path.abspath(libdir)} "
+            + "--prefix="
+        )
+        if is_windows:
+            cmd += "%CD%"
+        else:
+            cmd += "$(pwd)"
+        if pl.Path("builddir").exists():
+            cmd += " --wipe"
```
Hofer-Julian, line 124: I found `--wipe` a bit annoying to use in practice.
See [here](https://github.com/MODFLOW-USGS/modflow6/blob/e02997ba89607741d4040f419c4fb1d70427ed1e/.vscode/build_vscode.py#L17) for how I handled that with the vscode task

autotest/build_exes.py
```diff
@@ -60,131 +80,57 @@ def create_dir(pth):
...
+        print(f"running...\n{cmd}")
+        if os.system(cmd) != 0:
```
Hofer-Julian, line 126: `subprocess.run` is preferable over `os.system`.
If you set `check=True` you don't even have to check the returncode.

autotest/build_exes.py
```diff
@@ -60,131 +80,57 @@ def create_dir(pth):
...
-    msg = f"{pm.target} does not exist."
-    assert pm.returncode == 0, msg
+        elif arg.lower().startswith("-fc="):
+            fc = arg.split("=")[1]
+    # determine the FC environmental variable needs to be set
+    set_env_var = False
+    env_var = os.environ.get("FC")
+    if env_var is None:
+        set_env_var = True
+        if fc is None:
```
Hofer-Julian, line 96: Maybe you want to raise an Exception if `fc` is neither gfortran nor ifort?

autotest/build_exes.py
jdhughes-usgs: modified as suggested

autotest/build_exes.py
jdhughes-usgs: Have implemented a check. This will require future maintenance to support additional compilers (for example, `mpifort`, `mpif90`)

autotest/build_exes.py
Hofer-Julian: Okay!

PR #968, opened 2022-06-30
fix(ci): workaround for windows-2022 gfortran executable run failures

mjreno:
Creating a PR to test the workaround.


.github/common/link-gfortranlib5.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+
+FCDIR=/c/ProgramData/Chocolatey/bin
+LNDIR=/c/ProgramData/Chocolatey/lib/mingw/tools/install/mingw64/bin
+if [ -d "$FCDIR" ] && [ -d "$LNDIR" ] && [ ! -f "$FCDIR/libgfortran-5.dll" ]; then
+    ln -s "$LNDIR/libgfortran-5.dll" "$FCDIR/libgfortran-5.dll"
```
langevin-usgs, line 6: Hey @mjreno, would it be possible to try adding ..../mingw64/bin to the end of the system PATH variable, so that the dll can be found?  If you put it at the end of the path, then hopefully that would prevent a wrong version of python or other tools from being used.  This would have the advantage that if libgfortran changed from 5 to 6, for example, this would still work.  Just an idea.  If this isn't worth pursuing for some reason, then I will merge it in as is.

mjreno: Hey @langevin-usgs updating PATH is definitely another way to solve the issue, the reason I went with this one is because it doesn't rely on PATH as we really don't control it- in my limited testing it appeared that it was reset at each action step.  If this is true PATH would need to checked, and/or updated, at each new step where a gfortran executable was generated/run.  Updates (new steps added) would also need to be mindful of this behavior.

The other reason I went with this implementation is I think the dll should be in that directory (and shouldn't change without another windows-2022 MinGW version update).  I think sooner than seeing the library version change we will see the dll show up in that directory- that's why the script is checking if the file is there and if this happens the code should probably be removed.  I could be wrong about this and could create an issue in the actions/virtual-environments repo to clarify.

Having said this I'm ok with either solution- if you prefer the PATH solution I'll do a bit more testing to verify if/when PATH is getting reset as action steps are being run.

langevin-usgs: Sounds good to me. Will merge now.

mjreno:
@langevin-usgs I was thinking of this PR as a shorter term fix for a problem related to the container.  Do you want me to create an issue so that we can track it until I can verify that either way?  I can reach out to github actions support and if it turns out to be a generic problem that we need to handle then I'll come back with a more general solution.

.github/common/link-gfortranlib5.sh
Hofer-Julian: For future reference, env variables can be changed between steps by modifying GITHUB_ENV https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables

PR #977, opened 2022-07-06
Add conda `environment.yml`

Hofer-Julian:
This PR adds an `environment.yml` which should make it easier to keep docs and CI in sync.
It also changes the behavior of the vscode tasks.
It expects an environment called `modflow6` created with the following command.
```
conda env create --force -f environment.yml
```

You might also want to change your `"fortran.fortls.path"` setting to the `fortls` in your `modflow6` environment.

Ideally, you would use this environment for your daily development work as well and adapt it if our dependencies change.

@jmccreight, @mjr-deltares, @langevin-usgs, @jdhughes-usgs, @mjreno

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,22 @@
+name: modflow6
+
+channels:
+  - conda-forge
+  - defaults
+
+dependencies:
+  - appdirs 
+  - fprettify
+  - numpy
+  - matplotlib
+  - meson!=0.63.0
+  - ninja
+  - pip
+  - pip:
+      - git+https://github.com/modflowpy/flopy.git
+      - git+https://github.com/modflowpy/pymake.git
+      - git+https://github.com/Deltares/xmipy.git
+      - git+https://github.com/MODFLOW-USGS/modflowapi.git
+  - pytest
+  - pytest-xdist
+  - requests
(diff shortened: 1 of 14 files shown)


environment.yml
```diff
@@ -0,0 +1,21 @@
...
+  - ninja
+  - pip
+  - pip:
+      - git+https://github.com/modflowpy/flopy.git
+      - git+https://github.com/modflowpy/pymake.git
+      - git+https://github.com/Deltares/xmipy.git
+      - git+https://github.com/MODFLOW-USGS/modflowapi.git
+  - pytest
+  - pytest-xdist
+  - requests
```
Hofer-Julian, line 21: I am not 100% sure that we still need all of these dependencies

Hofer-Julian:
@mjreno This PR also modifies the vscode docs, so I changed a couple things that you touched.
Main things are:
- No installation instructions of `fprettify` as that is handled by `environment.yml`
- I overlooked at the review that you set formatting for all files. I've limited it now to fortran files

@jdhughes-usgs @langevin-usgs would be nice if one of you could merge this PR.
A few of the past merged PRs conflicted with this one, so I had to solve a couple of merge conflicts recently.

PR #978, opened 2022-07-06
ci: fix Intel OneAPI build on Windows

jdhughes-usgs:
The issue is caused by the latest meson release. Exclude meson version 0.63.0.

### DIFF
--- a/.github/common/install-python-std.sh
+++ b/.github/common/install-python-std.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 pip install wheel
-pip install requests appdirs numpy matplotlib pytest pytest-xdist meson ninja
+pip install requests appdirs numpy matplotlib pytest pytest-xdist meson!=0.63.0 ninja
 pip install https://github.com/modflowpy/flopy/zipball/develop
 pip install https://github.com/modflowpy/pymake/zipball/master
 pip install https://github.com/Deltares/xmipy/zipball/develop
--- a/meson.build
+++ b/meson.build
@@ -63,6 +63,8 @@ endif
 if fc_id == 'intel-cl'
   # windows
   compile_args += ['/fpe:0',              # Activate all floating point exceptions
+                   '/heap-arrays:0',
+                   '/traceback',
                    '/fpp',                # Activate preprocessing
                    '/Qdiag-disable:7416', # f2008 warning
                    '/Qdiag-disable:7025', # f2008 warning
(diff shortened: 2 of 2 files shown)


Hofer-Julian:
Interesting, I see nothing relevant in the release notes: https://mesonbuild.com/Release-notes-for-0-63-0.html

We probably should test newer meson versions in the CI to catch those in the pre-releases.

eli-schwartz:
What's the exact issue? I looked at recent CI logs and saw https://github.com/MODFLOW-USGS/modflow6/runs/7183579347?check_suite_focus=true but the failure state here seems to be that the compiler doesn't like some fortran source code?

I don't notice any changes to the Meson compiler support that might affect this.

> We probably should test newer meson versions in the CI to catch those in the pre-releases.

For a while now, Meson has been doing release candidate builds before any new version in order to give people a chance to do this, so if you want to have the CI install meson with `pip install --pre` that is certainly possible and we would love to see bug reports if it is breaking. :)

Hofer-Julian:
@eli-schwartz after investigating a bit, we've noticed that `fc.get_supported_arguments` (used [here](https://github.com/MODFLOW-USGS/modflow6/blob/c3f415362edd6a8e033204967281fd10061dacf8/meson.build#L88)) gives back a "not supported" for every compiler flag you throw at it starting with meson 0.63 and using the intel fortran compiler on Windows.

Since we need some of these compiler flags for a successful build, the CI failed with meson 0.63.

Additionally meson now seems to add a flag `/utf-8` which is not recognized by intel fortran. That one doesn't break the build but only spams the output.

Do you agree that we should open two issues for these behavior changes?

eli-schwartz:
That sounds pretty bad, indeed.

Can you upload the file `meson-logs/meson-log.txt` from a build directory that has this problem (i.e. was configured with 0.63.0)? It should show the debug details including the exact code snippet, compiler command line, and stdout/stderr of the `get_supported_arguments()` checks.

Hofer-Julian:
Sure, here it is: [meson-log.txt](https://github.com/MODFLOW-USGS/modflow6/files/9063445/meson-log.txt)

Hofer-Julian:
Thanks for the detective work @eli-schwartz.
Do you need us to do anything to report this bug or are you already on it?

> Unfortunately, Meson has no CI for ifort. It's proprietary freeware, so I guess there's no legal reason we couldn't :)

Indeed, if you need inspiration you can see here how we test `ifort` in our CI: https://github.com/MODFLOW-USGS/modflow6/blob/develop/.github/workflows/ci-tests-ifort.yml

> On second thought, we currently skip fortran tests entirely on Windows, because a fortran compiler is not part of the virtual-environments hosted tools cache

Are you sure about that?
I don't see that we have to install gfortran on Windows on our CI.
However, we had to add this workaround for problems with the latest Windows image: https://github.com/MODFLOW-USGS/modflow6/pull/968

eli-schwartz:
Perhaps I should clarify, that we have msys2 tests, cygwin tests, and MSVC tests, and the first two use GCC with gfortran out of the box. Those aren't "native" windows, though, and the native MSVC compiler (vcxproj as well as ninja backends) tests don't have fortran available.

> Do you need us to do anything to report this bug or are you already on it?

I'm on it.

PR #988, opened 2022-07-19
ci(actions):  cache mamba conda downloads and environment

mjreno:


Hofer-Julian (commented):
I like the idea but keep in mind that a lot of our Python packages are directly installed from git.
Caching would mean that they would become outdated after some time potentially breaking the CI.

This change also doesn't really seem to reduce CI times.
1min 50s vs 1min 30s

mjreno:
It's not a huge amount of time saved but where I looked it was closer to 2 minutes.  According to the documents the cache is invalidated once daily.

PR #1012, opened 2022-08-18
fix(gwf-npf.f90): leave k22 and k33 allocated if tvk active

### LINKED ISSUE #1000: NPF and TVK -- allocation of K22 and K33 arrays
**Describe the bug**
If K22 or K33 (or both) is not specified explicitly in the NPF input, it is not allocated its own array but instead points to K11.  This is done to save on storage and is fine unless TVK changes the value of that conductivity (K22 and/or K33).  In that case, the value TVK specifies for that conductivity overwrites the value of K11 because that conductivity points to K11.  In general terms, the fix would be to take the needs of TVK into account when (re)allocating storage for K22 and K33.

**To Reproduce**
1. Start with the tvk01 test problem, in which K11 and K33 are explicitly specified in the NPF input.  K22 is not and so points to K11.
2. Modify the TVK input so that K11 (i.e., K) is set to a value other than 1.0000000 at one of the cells in period 1.  For example, set K to 5.0000000 instead of 1.00000000 at cell (3, 3, 3).  If rerun at this point, the heads and flows will obviously be somewhat different than in the original problem.
3. Now add a line to the bottom of the period 1 input that sets K22 to 1.00000000.  This will cause the value of K to be set back to the original value of 1.00000000.  Rerunning the problem now gives the original heads and flows.
4. Starting fresh with the tvk01 test problem, modify the NPF input to specify K22 explicitly to some value.  This causes K22 to be allocated its own array, i.e., it does not point to K11.  Also, add a line to the bottom of the period 1 input that sets K22 at cell (3, 3, 3) to a value other than 1.0000000, say, 5.0000000.  Rerunning the problem still gives the original heads and flows because K22 does not overwrite K11 (i.e., K), and the value of K22 itself has no effect on the solution in this problem.

**Expected behavior**
The expected behavior when TVK changes K22 or K33 is that only that conductivity changes, without overwriting K11.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Windows 10 Enterprise
 - Version 10.0.19044 Build 19044

emorway-usgs:
Includes new test that does not specify K33 in NPF; however, TVK then modifies K33 in the second stress period which would previously stomp on K11.

Fixes #1000

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3npf8.f90
+++ b/src/Model/GroundWaterFlow/gwf3npf8.f90
@@ -146,7 +146,7 @@ contains
 ! ------------------------------------------------------------------------------
     ! -- modules
     ! -- dummy
-    type(GwfNpftype), pointer :: npfobj
+    type(GwfNpfType), pointer :: npfobj
     character(len=*), intent(in) :: name_model
     integer(I4B), intent(in) :: inunit
     integer(I4B), intent(in) :: iout
@@ -187,7 +187,7 @@ contains
     use SimModule, only: store_error
     use Xt3dModule, only: xt3d_cr
     ! -- dummy
-    class(GwfNpftype) :: this !< instance of the NPF package
+    class(GwfNpfType) :: this !< instance of the NPF package
     class(DisBaseType), pointer, intent(inout) :: dis !< the pointer to the discretization
     type(Xt3dType), pointer :: xt3d !< the pointer to the XT3D 'package'
     integer(I4B), intent(in) :: ingnc !< ghostnodes enabled? (>0 means yes)
@@ -243,7 +243,7 @@ contains
     use SparseModule, only: sparsematrix
     use MemoryManagerModule, only: mem_allocate
     ! -- dummy
-    class(GwfNpftype) :: this
+    class(GwfNpfType) :: this
     integer(I4B), intent(in) :: moffset
     type(sparsematrix), intent(inout) :: sparse
     ! -- local
@@ -266,7 +266,7 @@ contains
     ! -- modules
     use MemoryManagerModule, only: mem_allocate
     ! -- dummy
-    class(GwfNpftype) :: this
+    class(GwfNpfType) :: this
     integer(I4B), intent(in) :: moffset
     integer(I4B), dimension(:), intent(in) :: iasln
     integer(I4B), dimension(:), intent(in) :: jasln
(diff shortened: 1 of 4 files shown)


emorway-usgs:
2nd test of this bug fix specifies alternate values of k33 in TVK, but they should have no effect on flows between stress periods 1 and 2.  The test asserts that this condition holds true

src/Model/GroundWaterFlow/gwf3npf8.f90
```diff
@@ -1754,6 +1754,15 @@ subroutine read_grid_data(this)
     ! -- set ik33 flag
     if (lname(3)) then
       this%ik33 = 1
+    else if (.not. lname(3) .and. this%intvk > 0) then
+      this%ik33 = 1
+      if (this%ik33overk /= 0) then
+        write (errmsg, '(a)') 'K33OVERK option specified but K33 not specified.'
+        call store_error(errmsg)
+      end if
```
langevin-usgs, line 1762: @emorway-usgs, I see that this error message has been duplicated.  Is there any way to rework this logic so that the error message doesn't need to be duplicated?

src/Model/GroundWaterFlow/gwf3npf8.f90
```diff
@@ -1767,6 +1776,15 @@ subroutine read_grid_data(this)
     ! -- set ik22 flag
     if (lname(4)) then
       this%ik22 = 1
+    else if (.not. lname(4) .and. this%intvk > 0) then
+      this%ik22 = 1
+      if (this%ik22overk /= 0) then
+        write (errmsg, '(a)') 'K22OVERK option specified but K22 not specified.'
+        call store_error(errmsg)
+      end if
```
langevin-usgs, line 1784: Same comment here.

langevin-usgs (commented):
Looks good overall.  Thanks for tackling this.

src/Model/GroundWaterFlow/gwf3npf8.f90
emorway-usgs: Sorry about that, attempted to condense logic with 8301c7c

src/Model/GroundWaterFlow/gwf3npf8.f90
emorway-usgs: Resolved with 8301c7c (hopefully)

langevin-usgs:
Replaced by #1031

PR #1016, opened 2022-08-22
ci(gfortran): bump ubuntu 18.04 to 20.04

langevin-usgs:
* 18.04 no longer supported


langevin-usgs:
Looks like gcc-5 and gcc-6 are not possible now that Ubuntu 18.04 is deprecated.  Will need to disable the gfortran-endoflife workflows.

PR #1023, opened 2022-08-29
API string support

mjr-deltares:
This will support the first version of string access through the MODFLOW API. NB it has

* no support for pointers to strings (i.e. get_value_ptr_string is not implemented)
* no support for settings string values (yet)

We will update xmipy accordingly to use this with Python


mjr-deltares:
also see https://github.com/Deltares/xmipy/pull/81 with the supporting changes in xmipy

langevin-usgs (approved):
Looks great.  Glad we could get this worked out!

PR #1030, opened 2022-09-01
fix(stringlist): allocate string before assignment

mjreno:


src/Utilities/CharString.f90
```diff
@@ -100,4 +101,11 @@ subroutine write_unformatted(this, unit, iostat, iomsg)
     end if
   end subroutine write_unformatted
 
+  function strlen(this) result(length)
+    class(CharacterStringType), intent(in) :: this
+    integer :: length
+
```
langevin-usgs, line 107: should we say:

```
if (allocated(this%charstring)) then
  length = len(this%charstring))
else
  length = 0
end if
```

src/Utilities/StringList.f90
```diff
@@ -63,11 +63,13 @@ function GetStringFromList(list, indx) result(string)
     class(*), pointer :: obj
     type(CharacterStringType), pointer :: charcont
     !
-    string = ''
     obj => list%GetItem(indx)
     charcont => CastAsCharacterStringType(obj)
     if (associated(charcont)) then
+      allocate (character(len=charcont%strlen()) :: string)
       string = charcont
```
langevin-usgs, line 70: I think this should be:

```
string(:) = charcont
```

but please verify

langevin-usgs (changes requested):
I was hoping we could overload the intrinsic len function for this, but I couldn't get it to work.  

I made a couple minor suggestions.

src/Utilities/CharString.f90
mjreno: I was thinking len should report 0 if not allocated but should have verified.  I'll make the change.

src/Utilities/StringList.f90
mjreno: Will do.

mjreno: The suggestion is to replace line 70 right?

langevin-usgs (approved):

PR #1031, opened 2022-09-05
refactor(npf): always allocate and store k22 and k33

### LINKED ISSUE #1000: NPF and TVK -- allocation of K22 and K33 arrays

christianlangevin:
* Prior to this PR k22 and k33 would point to k11 if they weren't specified by the user in order to save memory
* With this PR, k22 and k33 are always allocated and stored, even if they aren't specified by the user
* This simplifies the code and makes API access to k22 and k33 more straightforward, at the memory cost of two grid-sized arrays in some cases
* Closes #1000
* Add two new tests for TVK

### DIFF
--- a/autotest/test_gwf_npf_tvk04.py
+++ b/autotest/test_gwf_npf_tvk04.py
@@ -1,0 +1,247 @@
+import os
+import sys
+
+import numpy as np
+import pytest
+
+try:
+    import pymake
+except:
+    msg = "Error. Pymake package is not available.\n"
+    msg += "Try installing using the following command:\n"
+    msg += " pip install https://github.com/modflowpy/pymake/zipball/master"
+    raise Exception(msg)
+
+try:
+    import flopy
+except:
+    msg = "Error. FloPy package is not available.\n"
+    msg += "Try installing using the following command:\n"
+    msg += " pip install flopy"
+    raise Exception(msg)
+
+from framework import testing_framework
+from simulation import Simulation
+
+ex = [
+    "tvk05",
+]
+exdirs = []
+for s in ex:
+    exdirs.append(os.path.join("temp", s))
+ddir = "data"
+
+time_varying_k = [1.0, 10.0]
+
+
+def build_model(idx, dir):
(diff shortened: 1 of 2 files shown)


mjr-deltares (approved):
looks good!

PR #1041, opened 2022-09-14
docs(mf6io): fix discrepancy in EVT example

emorway-usgs:
data in the period block contains pxpd1 and pxpd2, implying nseg=3, but nseg not specified which suggests a default value of 1


emorway-usgs:
A little nervous I've misunderstood the example problem offered for EVT, but I do think there is a minor discrepancy between what's shown in the `PERIOD` data block (i.e., `PXPD1` & `PXPD2`) and the `DIMENSIONS` block.  When not explicitly set, `NSEG` appears to take on a value of 1, which I think would mean `PXPD1` and `PXPD2` would be omitted.  Thinking that `NSEG` needs to be set to 3 to accommodate the two PXPD values.

langevin-usgs (changes requested):
Hey @emorway-usgs, thanks for this.  It looks like your change is correct.  Can you also change PXPD1 and PXPD2 in the comment in this file to read PXDP1 and PXDP2?

emorway-usgs:
I should've caught that the first time around.  Variables in comment renamed

langevin-usgs (approved):

PR #1058, opened 2022-09-30
fix(autotest): resolve inconsistent length units in Prudic model

### LINKED ISSUE #1053: Incorrect input file for ex-gwt-prudic2004t2
In the Lake package, the following line should be added to the Options section.
  LENGTH_CONVERSION   3.28081
because an outlet is specified and the model length unit is feet.

emorway-usgs:
As @rbwinst-usgs point out in #1053, the unit conversion in SFR implies length units of ft - see the SFR instantiation script below (i.e., 86400 * 3.28081 = 128390, which is the value entered for the argument `unit_conversion` below).  Length units not explicitly set in the DIS instantiation.  For consistency with the length units, can confirm that the LAK instantiation should have the argument `length_conversion=3.28081`.  

Have also updated the baked-in answers provided in the autotests.  Even so, the autotest titled "test_gwt_prudic2004t2fmiats.py" will fail because the following line of text is no longer written into the mfsim.lst file in the `transport` folder:

_Failed solution for step 19 and period 2 will be retried using time step of    80.00000_

The post-processing in the autotest looks for this line of text.

Still want to run the PR, though, because I would expect the other two variants of the Prudic model to pass muster.  Will need to talk to @langevin-usgs about the adaptive time stepping for this particular problem...apparently getting the length units spruced up in the LAK instantiation eliminated the need for ATS to kick in.

```
sfr = flopy.mf6.ModflowGwfsfr(             
    gwf,                                   
    print_stage=True,                      
    print_flows=True,                      
    stage_filerecord=gwfname + ".sfr.bin", 
    budget_filerecord=gwfname + ".sfr.bud",
    mover=True,                            
    pname="SFR-1",                         
    unit_conversion=128390.00,             
    boundnames=True,                       
    nreaches=len(rivlist),                 
    packagedata=sfrpackagedata,            
    connectiondata=sfrconnectiondata,      
    perioddata=sfrperioddata,              
)
```

Intended to fix #1053 (eventually)


langevin-usgs:
Does making this length unit conversion help with the concentration comparison shown [here](https://modflow6-examples.readthedocs.io/en/master/_examples/ex-gwt-prudic2004t2.html)?

emorway-usgs:
Hmm, it does not appear that updating to consistent units improved fits over on the [MODFLOW6-examples repo](https://modflow6-examples.readthedocs.io/en/master/_examples/ex-gwt-prudic2004t2.html).  In the plot below, the thinner of the solid lines correspond to the results when running with the `length_conversion=3.28081` parameter included in the `OPTIONS` block of the `LAK` input file.  The thick lines come from a plot I grabbed off of readthedocs.io (I used powerpoint to over-plot the two graphs).  

Prudic_results_with_length_conversion_in_LAK

langevin-usgs (commented):
Thanks for looking into this.  It does look like that length conversion should be added here, though I'm not sure the outlets approach is identical to what was done in the original Prudic t2 problem, though I haven't looked at that in detail.

What about this test?  `test_gwt_prudic2004t2gwtgwt.py`.  I think it also needs the length conversion.  And then there is also the prudic example in MODFLOW-USGS/modflow6-examples.  I guess if we are going to make this change, we should probably do it everywhere.  Before that, though, I wonder if we should look at the lake outflows in the original prudic problem and make sure there isn't a better way to handle this issue.

emorway-usgs:
@langevin-usgs @w-bonelli This may need a `rebase` with the `squash` option?  Basically, the baked-in answers were restored but I think the commit history is showing them changed and then changed back to their original values.  I'm wondering if a `squash` would effectively combine the offsetting commits such that the corresponding changes would be nullified in the history, if that makes sense?

langevin-usgs (approved):

langevin-usgs (approved):

PR #1088, opened 2022-11-28
feat(input-data-model): update buoyancy and fmi packages for idm

mjreno:
* update generic parser to read a block and load variable data if shape not known
* add generic handling of filein/fileout tag and removal of subpackages from ModflowInputType
* mem_set_value extended to support string1d types
* buy and fmi package idm updates


mjreno:
Will update and reopen after stress period data progress.

PR #1118, opened 2022-12-16
ci: add nightly job to cache intel compilers

w-bonelli:
According to the [cache docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) a cache created on the default branch (`develop` in this repo) can be restored by feature branches:

> Workflow runs can restore caches created in either the current branch or the default branch

Testing [on my fork](https://github.com/w-bonelli/modflow6/pull/53) before marking ready. If it works, we can cache oneAPI compilers overnight and restore from `develop` and PR branches all day. Since the key is now invalidated daily, the cache currently starts each day cold until a PR is merged to `develop`.

### DIFF
--- a/.github/workflows/large.yml
+++ b/.github/workflows/large.yml
@@ -3,6 +3,16 @@ on:
   schedule:
     - cron: '0 6 * * *' # run at 6 AM UTC every day
 jobs:
+  cache_ifort:
+    name: Cache Intel OneAPI compilers
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


w-bonelli:
This [seems to work](https://github.com/w-bonelli/modflow6/actions/runs/3731673533/jobs/6330198779) on my fork &mdash; in the linked example, a feature branch successfully restores the cache created by the large model job which ran earlier this morning. Scheduled runs are associated with the repo's default branch, so the cache was created for `develop`. As the docs state, feature/PR branches can restore from default branch caches, but the other way around does not work.

langevin-usgs (approved):

PR #1121, opened 2022-12-21
refactor(gwfgwf)

mjr-deltares:
- split model budget routines in Exchange for parallel dev
- allocate NPF angle arrays at 0 size when not present


mjr-deltares:
@langevin-usgs can you have a look at this? It's ready to go in, and I can do the same for GWT next.

langevin-usgs (approved):

PR #1127, opened 2022-12-24
refactor(idm):  consolidate package input context load in model create

mjreno:
This PR moves idm loads to model create.  This is the first of 2 steps in migrating idm out of model code completely.  More work will need to be done to accomplish that, including removing or reworking how the namefile_obj works in model create routines.  This object reads model namfiles and opens modlow6 files- operations that will need to move out of the model along with idm.  

This merge is relevant now because it sets the stage for idm stress period data related work- with input_load calls consolidated the code is better structured to support creating idm period data loader objects and adding those objects to simulation scope lists.

### DIFF
--- a/src/Model/NumericalModel.f90
+++ b/src/Model/NumericalModel.f90
@@ -468,10 +468,6 @@ contains
       call input_load('DISU6', 'GWF', 'DISU', modelname, pkgname, inunit, iout)
     case ('DISV6')
       call input_load('DISV6', 'GWF', 'DISV', modelname, pkgname, inunit, iout)
-    case ('DSP6')
-      call input_load('DSP6', 'GWT', 'DSP', modelname, pkgname, inunit, iout)
-    case ('NPF6')
-      call input_load('NPF6', 'GWF', 'NPF', modelname, pkgname, inunit, iout)
     case default
     end select
     !
--- a/src/Model/NumericalModel.f90
+++ b/src/Model/NumericalModel.f90
@@ -441,14 +441,10 @@ contains
   end function GetNumericalModelFromList
 
   !> @brief Load input context for supported package
-  !!
-  !!
   !<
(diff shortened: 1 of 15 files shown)


src/Model/NumericalModel.f90
```diff
@@ -439,4 +440,43 @@ function GetNumericalModelFromList(list, idx) result(res)
...
+    case ('DIS6')
+      call input_load('DIS6', 'GWF', 'DIS', modelname, pkgname, inunit, iout)
+    case ('DISU6')
+      call input_load('DISU6', 'GWF', 'DISU', modelname, pkgname, inunit, iout)
+    case ('DISV6')
+      call input_load('DISV6', 'GWF', 'DISV', modelname, pkgname, inunit, iout)
+    case ('DSP6')
+      call input_load('DSP6', 'GWT', 'DSP', modelname, pkgname, inunit, iout)
+    case ('NPF6')
+      call input_load('NPF6', 'GWF', 'NPF', modelname, pkgname, inunit, iout)
```
langevin-usgs, line 474: I'm not sure DSP and NPF should be here in NumericalModel as these packages are specific to GWT and GWF, respectively.  I suspect this is a temporary location.

langevin-usgs (commented):
I like where this is headed.  Should we keep this as draft or should we try to get it in?  Also, I think your pymake may need to be updated.

src/Model/NumericalModel.f90
mjreno: My focus was on minimizing code impact as I do see it as a temporary move but you make a good point and the changes will still be minimal- commit coming.

mjreno: I was thinking that with consensus it could be immediately merged but I'm open either way.  Do you have a preference?

langevin-usgs (approved):

PR #1130, opened 2023-01-04
refactor(tests): use devtools fixtures

w-bonelli:
Update tests that load models from external repos to use devtools fixtures, with a bit of other refactoring. This was carved out of #1066 for a smaller changeset and more incremental refactor.

- Use fixtures from `modflow-devtools` to:
  - map target executable names to their paths (`targets` fixture)
  - parametrize tests with models loaded from external repos
    - the set of models tested is not changed &mdash; coupled GWT models are skipped, as before
    - support example selection by model name/package via `--model`/`--package` pytest CLI options
      - ~~there is a package filtering bug in devtools, soon to be fixed in https://github.com/MODFLOW-USGS/modflow-devtools/pull/44~~
    - support `--original-regression` flag as a pytest CLI option/fixture (configured in new `conftest.py` file)
    - update `DEVELOPER.md` to reflect devtools convention that external model repos live side by side with `modflow6`, or optional `REPOS_PATH` environment variable to specify a different location
- Use devtools and flopy instead of pymake in `build_exes.py` and `get_exes.py`. Give both scripts an `argparse` CLI with optional `--path` argument to specify a different bin path (defaults to the existing location, `bin` in the project root).
- Remove pymake usage from framework/simulation and external model tests. Move some utility functions previously in pymake into `common_regression.py`. A followup PR could replace pymake usages in the rest of the tests.
- Use Github API token in CI step downloading exes (bypass rate limit)

### DIFF
--- a/autotest/test_z01_testmodels_mf6.py
+++ b/autotest/test_z01_testmodels_mf6.py
@@ -1,265 +1,55 @@
-import os
-import subprocess
-import sys
-
 import pytest
 
-try:
-    import pymake
-except:
-    msg = "Error. Pymake package is not available.\n"
-    msg += "Try installing using the following command:\n"
-    msg += " pip install https://github.com/modflowpy/pymake/zipball/master"
-    raise Exception(msg)
 
-try:
-    import flopy
-except:
-    msg = "Error. FloPy package is not available.\n"
-    msg += "Try installing using the following command:\n"
-    msg += " pip install flopy"
-    raise Exception(msg)
-
-from common_regression import (
-    get_example_basedir,
-    get_example_dirs,
-    get_home_dir,
-    get_select_dirs,
-    get_select_packages,
-    is_directory_available,
-    set_mf6_regression,
-)
+from conftest import should_compare
 from simulation import Simulation
-from targets import get_mf6_version
-
-# find path to examples directory
-home = get_home_dir()
-
-find_dir = "modflow6-testmodels"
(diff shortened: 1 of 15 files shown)


autotest/test_z01_testmodels_mf6.py
```diff
@@ -1,264 +1,55 @@
...
+    "test034_nwtp2_1d": ("6.2.1",),
+    "test045_lake1tr_nr": ("6.2.1",),
+    "test045_lake2tr": ("6.2.1",),
+    "test045_lake2tr_nr": ("6.2.1",),
+    "test051_uzfp2": ("6.2.1",),
+    "test051_uzfp3_lakmvr_v2": ("6.2.1",),
+    "test051_uzfp3_wellakmvr_v2": ("6.2.1",),
+    "test045_lake4ss": ("6.2.2",),
+    "test056_mt3dms_usgs_gwtex_dev": ("6.2.2",),
+    "test056_mt3dms_usgs_gwtex_IR_dev": ("6.2.2",),
```
langevin-usgs, line 28: FYI, #1119 changes these two to 6.4.1.

langevin-usgs (commented):
Looks great, @w-bonelli.  Nice cleanup and nice documentation.

autotest/test_z01_testmodels_mf6.py
w-bonelli: Thanks, updated this PR to 6.4.1.

PR #1134, opened 2023-01-14
fix(SfrCrossSectionUtils): fix for n-point x-section area calculation

emorway-usgs:
Manually verified the calculation using a trapezoid x-section configuration.  That autotest is currently part of the SFE work


emorway-usgs:
Smoke tests finally passed, but the other builds were failing.  Owing to all the merge commits along the way, this PR has been superseded by #1172 for a clean look at what's going on.  Closing.

PR #1137, opened 2023-01-18
feat(multi-domain sorption): generalize formulation

### LINKED ISSUE #851: Allow user to specify immobile domain fraction within a cell for GWT IST Package
The volumetric fraction of an immobile domain within a model cell is approximated with a rather limiting assumption that is also present in MT3D.  A better approach is to allow the user to optionally specify the fraction of a cell that is immobile (f_im).

langevin-usgs:
* close #851
* add new chapter to Supplemental Technical Information
* update release notes
* revise and simplify code
* change definitions for MST and IST bulk density


aprovost-usgs (approved):
I have reviewed the changes and the PR looks ready to be merged. One thing that is probably correct/inconsequential but might warrant looking into prior to release is the reversal of dimensions from (nlay, ncpl) to (ncpl, nlay) in gwf-disv-griddata.dat.

PR #1147, opened 2023-02-14
fix(NPF): negative icelltype without THICKSTRT produced unexpected results

### LINKED ISSUE #1146: HFB Package Inactivated with ICELLTYPE=-1
**BLUF:**
When I build a model using the HFB package, the HFB has no impact on the solution if ICELLTYPE is -1 unless the Newton-Raphson formulation is activated (and linear acceleration is BICGSTAB).

**Detailed description of the issue:**
I have attached a zip file with three jupyter notebooks with a simple model built using flopy.  Some images are pasted below.  More are created in the jupyter notebooks.

The model is a simple box with three layers.  The heads are specified at the left and right boundaries to force a gradient with flow from left to right.  A well is located in layer 2 in the upper left quadrant.  A hydraulic barrier blocks half of the upstream boundary condition from the well and forces flow to go around the wall to reach the well or to reach the downstream boundary.  All four models are identical except for the ICELLTYPE, the NEWTON UNDER_RELAXATION option in the GWF and the LINEAR_ACCELERATION option in the IMS package.  The four files are:

*WallTest_ICELL-1_NoNewton.ipynb* - Here the ICELLTYPE is -1, and the Newton option has NOT been activated.  The Layer 2 solution is:
image
Note that while the boundary conditions and well are acting as expected, there is no indication of an HFB in the solution.  The following image shows a cross-section which passes right through where the HFB should be (black boxes are CHD locations):
image
The gradient is almost linear because this cross-section is not close to the well.  The HFB package has been inactivated somehow and does not impact the solution.  

*WallTest_ICELL1_NoNewton.ipynb* - Here the ICELLTYPE is 1 and the Newton option has NOT been activated.  The Layer 2 solution is:
image
Note that there is a clear indication of an HFB wall with heads building up upstream of the wall.  The following image shows a cross-section which passes right through where the HFB is:
image
With this setup, the HFB acts exactly as one would expect.

*WallTest_ICELL-1_Newton.ipynb* - During testing, I discovered that ICELLTYPE can be set to -1 as long as the Newton Raphson option is included.  Here NEWTON UNDER_RELAXATION has been added to the name file and LINEAR_ACCELERATION BICGSTAB is added to the IMS package.  The Layer 2 solution is: 
image
This looks identical to the previous file except a few cells in Layer 1 were dry until the Newton formulation was added (which was expected).  The results clearly show the impact of the wall.  The cross-section through the HFB is:
image

I also note that if ICELLTYPE is -1 and the Newton-Raphson option is not active, but THICKSTRT is on in the NPF package, the HFB package works as expected.

**Expected behavior:** I would expect the HFB package to impact the solution no matter what options are selected.  Based on my reading of the MODFLOW6 documentation, the best ICELLTYPE option for this model would be 1, but without THICKSTRT turned on, it seems like ICELLTYPE =1 and ICELLTYPE = -1 would be the same.

**Version Information:** These runs were made using MODFLOW 6 version 6.3.0 compiled Mar 02 2022.  Flopy version 3.3.6 was used to create the screenshots and to run the attached Jupyter notebooks.  The same results were obtained using GMS 10.7.1, which also uses version 6.3.0 of MODFLOW 6.  All were run on windows.

[HFB_issue_Jupyter.zip](https://github.com/MODFLOW-USGS/modflow6/files/10709289/HFB_issue_Jupyter.zip)

langevin-usgs:
* The ICELLTYPE input variable in the Node Property Flow (NPF) Package behaves differently depending on whether or not the THICKSTRT option is specified by the user.  In some cases, the program would give unexpected results if a negative value was specified for ICELLTYPE and the THICKSTRT option was not active.  For example, the Horizontal Flow Barrier (HFB) Package did not work property when negative values for ICELLTYPE were specified by the user, but the THICKSTRT option was not activated.  The program was modified so that negative ICELLTYPE values provided by the user are automatically reassigned a value of one when the user does not activate the THICKSTRT option.  This is the intended behavior and is consistent with the input and output guide.
* close #1146


doc/ReleaseNotes/v6.5.0.tex
```diff
@@ -25,12 +25,12 @@
 	%	\item xxx
 	\end{itemize}
 
-	%\underline{INTERNAL FLOW PACKAGES}
-	%\begin{itemize}
-	%	\item xxx
+	\underline{INTERNAL FLOW PACKAGES}
+	\begin{itemize}
+		\item The ICELLTYPE input variable in the Node Property Flow (NPF) Package behaves differently depending on whether or not the THICKSTRT option is specified by the user.  In some cases, the program would give unexpected results if a negative value was specified for ICELLTYPE and the THICKSTRT option was not active.  For example, the Horizontal Flow Barrier (HFB) Package did not work property when negative values for ICELLTYPE were specified by the user, but the THICKSTRT option was not activated.  The program was modified so that negative ICELLTYPE values provided by the user are automatically reassigned a value of one when the user does not activate the THICKSTRT option.  This is the intended behavior and is consistent with the input and output guide.
```
jdhughes-usgs, line 30: ...properly when negative values for ICELLTYPE were specified by the user and the THICKSTRT option...

jdhughes-usgs (changes requested):

doc/ReleaseNotes/v6.5.0.tex
langevin-usgs: Good catch.  Fixed on last commit.

PR #1150, opened 2023-02-17
refactor(ci): convert fortran lint bash script to python

mjreno:


langevin-usgs (approved):
Nice addition.  Thanks @mjreno

PR #1151, opened 2023-02-21
feat(parallel): add parallel capabilities (with mpi and petsc) without enabling

mjr-deltares:


langevin-usgs:
@mjr-deltares, looks like this is coming along nicely!

mjr-deltares:
@langevin-usgs I think this is in a decent state now and we should probably move forward and pull it in. Let's discuss details offline.

langevin-usgs (approved):
Lots of great stuff in here.  I went through it all, and we discussed the details.  I think it's ready.  Time to push the merge button.

PR #1152, opened 2023-02-28
doc: update LaTeX documents to use latest bst file in usgslatex install

jdhughes-usgs:


langevin-usgs (commented):
So then what happens with the `modflow6/doc/usgs.bst` file?  Is that file dead?  My only concern is that we have to rely on a usgslatex install of the bst file.  Maybe that's fine.

jdhughes-usgs:
We already rely on the usgslatex install for building the documentation in CI/CD. We don't need the usgs.bst in the repo anymore. Will remove it.

PR #1154, opened 2023-03-03
feat(sfr): add alternative length_conversion and time_conversion varibles

jdhughes-usgs:
Replacement for unit_conversion variable and is consistent with approach used in lake (LAK) package.


doc/mf6io/mf6ivar/dfn/gwf-sfr.dfn
```diff
@@ -306,9 +306,26 @@ name unit_conversion
 type double precision
 reader urword
 optional true
+deprecated 6.5.0
 longname conversion factor
 description value (or conversion factor) that is used in calculating stream depth for stream reach. A constant of 1.486 is used for flow units of cubic feet per second, and a constant of 1.0 is used for units of cubic meters per second. The constant must be multiplied by 86,400 when using time units of days in the simulation.
```
langevin-usgs, line 311:
```suggestion
description value (or conversion factor) that is used in calculating stream depth for a stream reach. A constant of 1.486 should be specified by the user if flow units are cubic feet per second; a constant of 1.0 should be specified by the user for flow units of cubic meters per second. The constant specified by the user must be multiplied by 86,400 if time units of days are used for the simulation.
```

doc/mf6io/mf6ivar/dfn/gwf-sfr.dfn
```diff
@@ -306,9 +306,26 @@ name unit_conversion
...
 
+block options
+name length_conversion
+type double precision
+reader urword
+optional true
+longname length conversion factor
+description real value that is used in converting Manning's equation to consistent length units. LENGTH\_CONVERSION should be set to 3.28081, 1.0, and 100.0 when using length units (LENGTH\_UNITS) of feet, meters, or centimeters in the simulation, respectively. LENGTH\_CONVERSION does not need to be specified if LENGTH\_UNITS are meters.
```
langevin-usgs, line 319:
```suggestion
description real value that is used in converting Manning's equation to consistent length units. LENGTH\_CONVERSION should be set to 3.28081, 1.0, and 100.0 when the model has length units (LENGTH\_UNITS) of feet, meters, or centimeters in the simulation, respectively. LENGTH\_CONVERSION does not need to be specified if LENGTH\_UNITS are meters.
```

langevin-usgs (commented):
The unit conversion for SFR seems a little confusing.  I wonder if we can find a better way to describe how it works.  The way it is explained here is that these are the conversion factors for converting Manning's equation from SI units to consistent model units.  Is there some additional description that we could provide here?

doc/ReleaseNotes/v6.5.0.tex
```diff
@@ -6,6 +6,7 @@
...
+		\item Add LENGTH\_CONVERSION and TIME\_CONVERSION variables to replace the UNIT\_CONVERSION variable in the SFR Package input file. The unit conversion factor used to convert Manning's equation from SI units to consistent model units. LENGTH\_CONVERSION does not need to be specified if LENGTH\_UNITS are meters. TIME\_CONVERSION does not need to be specified if TIME\_UNITS are seconds. Warning messages will be issued if UNIT\_CONVERSION variable is specified. The model will terminate with an error if UNIT\_CONVERSION and LENGTH\_CONVERSION and TIME\_CONVERSION variables are specified. The UNIT\_CONVERSION variable will eventually be deprecated.
```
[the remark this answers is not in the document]
jdhughes-usgs: Modified to address concerns.

doc/mf6io/mf6ivar/dfn/gwf-sfr.dfn
jdhughes-usgs: Modified to address concerns.

doc/mf6io/mf6ivar/dfn/gwf-sfr.dfn
jdhughes-usgs: Modified to address concerns.

langevin-usgs (approved):

PR #1155, opened 2023-03-07
feat(idm): update simulation create to source from simnam input context

mjreno:


mjr-deltares:
Hi @mjreno, great stuff! Is this now ready to go in? If so, you can assign me as reviewer if you want.

PR #1156, opened 2023-03-08
doc(parallel): first version of developer docs for parallel MODFLOW

mjr-deltares:


mjr-deltares:
Hey @w-bonelli, I thought it would be a good idea if you can have a look at this. It should be a start for our parallel developer docs. I'd be interested to hear your opinion on this.

w-bonelli (approved):

PR #1161, opened 2023-03-12
feat(ems): add hooks for explicit model solution

langevin-usgs:
This PR adds the hooks for an explicit model solution.  An explicit model solution does not do any solving, but rather, it calls a solve method on each model that is assigned to it.  It is being added now to support models that are currently under development, including the particle tracking (PRT) model, the surface water flow (SFW) model, and the watershed flow model.

One thing to note here is that the base model class was modified to include four new methods.  We may need to consider if there is an alternative for this that is a bit cleaner.


src/Solution/ExplicitSolution.f90
```diff
@@ -0,0 +1,394 @@
...
+    class(ExplicitSolutionType) :: this !< ExplicitSolutionType instance
+    ! -- local variables
+    class(BaseModelType), pointer :: mp => null()
+    integer(I4B) :: im
+    real(DP) :: ttsoln
+    !
+    ! -- particle solve
+    call code_timer(0, ttsoln, this%ttsoln)
+    do im = 1, this%modellist%Count()
+      mp => GetBaseModelFromList(this%modellist, im)
```
mjr-deltares, line 289: I think that we should consider if ExplicitSolution has a list of ExplicitModel types, just NumericalSolution holds a list of NumericalModel types. However, no need to stall this PR for that.

mjr-deltares (approved):

PR #1166, opened 2023-03-14
refactor(mode): correct file modes

mwtoews:
This PR only changes file modes, which are generally only visible to POSIX operating systems. Many files were mistakenly set with an "executable" mode, when they were not. These have been fixed with:

    $ git ls-files | xargs -I {} chmod -x "{}"

and of course there are a few files that actually should be set with "executable", found with:

    $ git grep "^#!"

then manually inspected and set back as "executable" with `chmod +x filename` (marked with "100644 → 100755")


langevin-usgs (approved):
Thanks, @mwtoews.

PR #1167, opened 2023-03-14
fix size of mshape

mjr-deltares:
This would help parallel a lot, creating this PR to investigate impact

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3disu8.f90
+++ b/src/Model/GroundWaterFlow/gwf3disu8.f90
@@ -1408,6 +1408,8 @@ contains
     !
     ! -- Initialize
     this%mshape(1) = this%nodesuser
+    this%mshape(2) = -1
+    this%mshape(3) = -1
     !
     ! -- Return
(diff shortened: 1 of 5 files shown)


langevin-usgs:
Would be good to get feedback from @mjreno as he is heavily invested in using mshape for the input data model.

mjreno:
Seems reasonable to me.  The input data model is setting MODEL_SHAPE under the input model mempath and NCELLDIM will be needed there as well to support reading cell ids for period and connection data.  This could readily follow the same pattern you are using here in the List Reader.  

Consistency would be nice, maybe it's worth a thought though- is it possible that the input model could structure the data in a way that is meaningful for parallel now, i.e. suffice for the needs of parallel development at least in the short term?

mjr-deltares:
@mjreno @langevin-usgs I was investigating this route, but it leads to other, undesirable complications. The fix in PR #1168 was my alternative, and it is more straightforward I think. I will close this one for now, if that's ok with you.

PR #1169, opened 2023-03-14
fix(listreader): add error checking for binary list input

langevin-usgs:


jdhughes-usgs (commented):
Might be nice to include the full user provided cellid [(l, r, c), (l, node), (node)] inn the error message. The list entry (1 -> nbound) would be another good piece of information. Otherwise it looks good.

langevin-usgs:
Implemented the following error format:

```
ERROR REPORT:

  1. List entry 2 contains cellid (1,-1,100) but this cell is invalid for
     model with shape (1,1,100)

UNIT ERROR REPORT:

  1. ERROR OCCURRED WHILE READING FILE 'gwf_chd01.chd'
```

PR #1172, opened 2023-03-17
fix(SfrCrossSectionUtils): fix for n-point x-section area calculation

emorway-usgs:
- Supersedes #1134 which was contaminated by merge commits.
- Ran black, fprettify
- The bug at the crux of this PR is shown in the screen grab below.  The area of the triangle should be calculated as 1/2 * base * width.  
- mf6io.pdf was provided with a new paragraph (and figure) that highlights what happens if the simulated stream stage exceeds the defined points.
- A complex n-point cross-section was manually checked (see later post in this thread).

crux


emorway-usgs:
Smoke tests are passing, but a number of [modflow6-testmodels](https://github.com/MODFLOW-USGS/modflow6-testmodels) are failing.  This result is not unexpected since many of the test models use SFR and have baked-in answers that were derived prior to this PR.  That is, the answers being compared to were calculated with an implementation of Mannings equation that uses the hydraulic radius; however, the hydraulic radius wasn't taking the vertically wetted sides of the rectangular channel into account.

Here is the equation being solved by `gwf3sfr8.f90` (Equation 7-7 in main MF6 documentation): 
qmann

And here is the associated code:
```
...
aw = this%calc_area_wet(n, depth)
wp = this%calc_perimeter_wet(n, depth)
if (wp > DZERO) then
  rh = aw / wp
else
  rh = DZERO
end if
qman = this%unitconv * aw * (rh**DTWOTHIRDS) * sqrt(s) / r
...
```
Note that the function `calc_perimeter_wet()` was not including the vertically wetted sides of a wide rectangular geometry prior to this PR, which might be OK if the equation being solved by SFR was the more simple Equation 7-8 in the main documentation:

qmann_simp

As best I can tell, 7-8 isn't used anywhere, and so this PR includes a fix for the hydraulic radius calculation as well.  

The tests in the modflow6-testmodels repository that will need updating include:

- test028_sfr
- test028_sfr_dev
- test028_sfr_mvr_dev
- test051_uzfp3_lakmvr_2uzfmodels_dev
- test051_uzfp3_lakmvr_dev
- test051_uzfp3_lakmvr_nouzf_dev
- test051_uzfp3_lakmvr_v2
- test051_uzfp3_lakmvr_v2_dev
- test051_uzfp3_wellakmvr_v2
- test051_uzfp2
- test051_uzfp2TS_dev
- test051_uzfp2_dev
- test051_uzfp2_mvr_dev
- test051_uzfp2_nouzf_dev
- test051_uzfp2_openclose_dev
- test051_uzfp3_aeET_lakmvr_dev
- mf5to6/test1tr
- mf5to6/test1tr_chd
- mf5to6/test1tr_icalc012
- mf5to6/testUzfLakSfr
- mf5to6/testUzfLakSfr_laketable
- mf5to6/testlgrsfr

emorway-usgs:
@jdhughes-usgs, thanks for chatting with me about this PR.  After re-evaluating, there's a decision to be made with regard to whether or not the n-point cross-section approach assumes a wide rectangular channel geometry when calculating the wetted perimeter.  The situation is depicted below.  
geometry_types
Currently, SFR without n-point cross-sections uses the specified width of the channel to calculate the wetted perimeter used in Manning's equation (the red line shown in the left image above).  However, if a n-point geometry such as the one shown on the right were configured the same way, I'm thinking that it should take into account the vertically wetted sides since in other arbitrarily configurated x-section geometries it likely makes sense to include the vertically wetted sides when calculating the hydraulic radius for determining the flow depth.  For example, the figure below depicts one of the autotests and it strikes me as counterintuitive to restrict the wetted perimeter to the length of wetted channel between [0.0, 0.0] and [1.0, z] and omit the vertically wetted sides.  
npoint
It's easy enough to modify the PR to make this change, though I wonder if it would be helpful to draw this distinction in the mf6io.pdf documentation, for example?

emorway-usgs:
After some further follow-up conversation with @jdhughes-usgs, only the changes related to the channel x-sectional area should be included in this PR.  When using SFR without the n-point cross-section option, i.e., wide rectangular channel, the vertically wetted sides should not be considered to maintain fidelity with the MF2k5 based codes.  Noting that the code uses eqn 7-7 in the main MF6 documentation (provided in an earlier post of this PR thread), which includes hydraulic radius, it seems a little non-intuitive that the hydraulic radius doesn't include the entire wetted perimeter, but again, the code will remain consistent with the MF2K5 code base.

src/Model/GroundWaterFlow/gwf3sfr8.f90
```diff
@@ -4271,6 +4271,7 @@ subroutine sfr_calc_xs_depth(this, n, qrch, d)
       else
         derv = DZERO
       end if
+      if (derv > 10) derv = 10
```
jdhughes-usgs, line 4274: I don't think we want these kind of conditionals. We should look for a better approach.

emorway-usgs: @jdhughes-usgs, Agreed.  Mainly wanted to get something to pass.  Is there a common (or more sophisticated) approach for dealing with really steep slopes (`derv`) in the search algorithm?  In some of the autotests, `derv` was spiking as high as 800K+ leading to unstable results (in some cases, negative flows).

emorway-usgs:
Here is the new paragraph in mf6io.pdf:

new_para

doc/mf6io/gwf/sfr.tex
```diff
@@ -94,6 +94,16 @@ \subsection{Streamflow Routing Package Cross-Section Table Input File} \label{se
...
 \end{figure}
 
+Where irregular cross-sections are used to define variable stream geometries, the wetted perimeter used in Manning’s equation [(Equation 7-7, \cite{modflow6gwf}] depends on the number of points defining the cross-section and the simulated stage.  Using only the minimum number of points (i.e., 2-point cross section), \mf will not account for the vertically wetted streambed area when the stage rises above the elevation of the defined points. For example, the 2-point cross-sections shown in fig.~\ref{fig:sfr-n-point-wp}A-C depict the cross-sectional areas (light blue) and wetted perimeters (orange) calculated by the SFR package and used in \cite{modflow6gwf} (Equation 7-7).  In applications where the wetted perimeter should account for vertically wetted sides, additional points above the maximum anticipated stage should be defined (fig.~\ref{fig:sfr-n-point-wp}D).  Note that when the simulated stream stage rises above the points representing the top of the channel, the additional cross-sectional flow area above the defined points will be accounted for but the corresponding wetted perimeter will not extend above the defined points (fig.~\ref{fig:sfr-n-point-wp}E,F).
```
langevin-usgs, line 97:
```suggestion
Where irregular cross sections are used to define cross-sectional stream geometries, the wetted perimeter used in Manning’s equation [(Equation 7-7, \cite{modflow6gwf}] depends on the number of points defining the cross section and the simulated stage.  Using only the minimum number of points (i.e., 2-point cross section), \mf does not include perimeter lengths above the uppermost defined points in the wetted perimeter calculations. For example, the 2-point cross sections shown in fig.~\ref{fig:sfr-n-point-wp}A-C depict the cross-sectional areas (light blue) and wetted perimeters (orange) calculated by the SFR package and used in \cite{modflow6gwf} (Equation 7-7).  In applications where the intent is for the wetted perimeter to include the entire lengths of wetted sides, additional points above the maximum anticipated stage should be defined (fig.~\ref{fig:sfr-n-point-wp}D).  Note that when the simulated stream stage rises above the points representing the top of the channel, the additional cross-sectional flow area above the defined points will be accounted for but the corresponding wetted perimeter will not extend above the defined points (fig.~\ref{fig:sfr-n-point-wp}E,F).
```

emorway-usgs:
Addressed some further issues highlighted by a more complex channel geometry (shown next).  
x-sec

The complex channel geometry was manually checked using the attached excel file vs the attached python script that invokes the `cross_section_functions.py` script that was updated by this PR.  
[irreg_xsec_check.xlsx](https://github.com/MODFLOW-USGS/modflow6/files/11400852/irreg_xsec_check.xlsx)
[irreg_xsec_check.py.txt](https://github.com/MODFLOW-USGS/modflow6/files/11400860/irreg_xsec_check.py.txt)

jdhughes-usgs (approved):
Looks good. Thanks for finding the issue with the area calculations.

langevin-usgs (approved):

PR #1173, opened 2023-03-17
bug(lak): address available water issue for evaporation

jdhughes-usgs:


doc/ReleaseNotes/v6.5.0.tex
```diff
@@ -7,6 +7,7 @@
...
+		\item Add MAXIMUM\_ITERATIONS and MAXIMUM\_STAGE\_CHANGE variables in the LAK Package input file. The MAXIMUM\_ITERATIONS variable is used to change the maximum number of iteration and would only need to be increased from the default value if one or more lake in a simulation has a large water budget error.  The MAXIMUM\_STAGE\_CHANGE variable defines the stage closure tolerance for each lake. The MAXIMUM\_STAGE\_CHANGE would only need to be increased or decreased from the default value if the water budget error for one or more lake is too small or too large, respectively.
```
langevin-usgs, line 10:
```suggestion
		\item Add MAXIMUM\_ITERATIONS and MAXIMUM\_STAGE\_CHANGE variables in the LAK Package input file. The MAXIMUM\_ITERATIONS variable is used to change the maximum number of iterations and would only need to be increased from the default value if one or more lakes in a simulation has a large water budget error.  The MAXIMUM\_STAGE\_CHANGE variable defines the stage closure tolerance for each lake. The MAXIMUM\_STAGE\_CHANGE would only need to be increased or decreased from the default value if the water budget error for one or more lakes is too small or too large, respectively. 
```

doc/ReleaseNotes/v6.5.0.tex
```diff
@@ -43,7 +44,8 @@
...
-		
+		\item Added an additional convergence check to the Lake (LAK) Package to ensure that the lake residual meets solver tolerance. In previous versions, it was possible for a time step to terminate even if there was a relatively large lake residual. The lake residual is converted to a depth using the time step length and area, and the depth is compared to the Iterative Model Solution (IMS) DVCLOSE input parameter. If the residual depth is greater than DVCLOSE, then the iteration is marked as not converged. The maximum residual depth and the advanced package feature number is written for each outer iteration to a comma-separated value file, provided the PACKAGE\_CONVERGENCE option is specified in the options block. This fix may increase the number of iterations required before convergence is achieved in existing models that use the LAK Package.
```
langevin-usgs, line 47:
```suggestion
		\item Added an additional convergence check to the Lake (LAK) Package to ensure that the lake residual meets solver tolerance. In previous versions, it was possible for a time step to converge even if there was a relatively large lake residual. The lake residual is converted to a depth using the time step length and lake surface area.  This depth is compared to the Iterative Model Solution (IMS) DVCLOSE input parameter. If the residual depth is greater than DVCLOSE, then the iteration is marked as not converged. The maximum residual depth and the advanced package feature number (lake number) is written for each outer iteration to a comma-separated value file, provided the PACKAGE\_CONVERGENCE option is specified in the options block. This fix may increase the number of iterations required to reach convergence for existing models that use the LAK Package.
```

langevin-usgs (commented):
Looks pretty good to me.  Minor suggestions are included in the review.

src/Model/GroundWaterFlow/gwf3lak8.f90
```diff
@@ -5458,7 +5521,7 @@ subroutine lak_solve(this, update)
     end do
 
     iicnvg = 0
-    maxiter = 150
+    maxiter = this%maxlakit
```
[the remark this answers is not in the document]
jdhughes-usgs: Made consistent with comparable variable in the SFR Package

PR #1177, opened 2023-03-20
feat(idm):  generate component dfn list selectors from dfn2f90.py

mjreno:


langevin-usgs (approved):
This looks really clean.  The only thing that jumps out at me is whether "multi_package" would be better described as "supports_multi_instance", which is a bit long?  Maybe something to kick around as we move forward.

PR #1189, opened 2023-03-31
feat(api): add initialization with MPI to API

mjr-deltares:


mjr-deltares:
Hi @langevin-usgs I think this can go in without too many consequences, mostly relevant to be able to study/debug parallel development through the API. If you agree, please go ahead and merge

srcbmi/mf6xmi.F90
```diff
@@ -97,6 +97,29 @@ module mf6xmi
 
 contains
 
+#if defined(__WITH_MPI__)
+  function xmi_initialize_mpi(mpi_comm) result(bmi_status) &
```
langevin-usgs, line 101: Maybe in the future we think about putting this routine in a separate MPI labeled Fortran source file to keep MPI directives containerized?

langevin-usgs (approved):

PR #1193, opened 2023-04-03
feat(par): add load balancing for gwf and gwt

mjr-deltares:
where GWT in parallel is not functional yet

### DIFF
--- a/autotest/test_par_gwf03.py
+++ b/autotest/test_par_gwf03.py
@@ -13,17 +13,18 @@ from simulation import TestSimulation
 #
 # a: 1 cpus, 1 model
 # b: 1 cpus, 4 models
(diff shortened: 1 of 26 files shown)


mjr-deltares:
Hi @mjreno I've tried to synchronize the model load mask for idm with the models create routine. Basically through the subroutine `create_load_mask`, maybe you want to have a look at this PR. Not settled on where this common logic should go, but we can discuss offline.

mjr-deltares:
HI @langevin-usgs , I think this can go in (I've talked to @mjreno and he is on par as far as the integration with idm is concerned here).

langevin-usgs (approved):
Looks great.  It's all coming together nicely.

PR #1196, opened 2023-04-05
ci: don't trigger ci.yml on changes to LaTeX files

w-bonelli:
no need to run CI test workflow if only LaTeX files have changed


w-bonelli:
this is already covered by the `doc/` pattern, my mistake

PR #1197, opened 2023-04-06
fix(ptc): fix ptc residual calculation

jdhughes-usgs:
Also modify order of model_ptcchk so that it is not called if 
NO_PTC ALL or NO_PTC FIRST option are specified.

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3.f90
+++ b/src/Model/GroundWaterFlow/gwf3.f90
@@ -24,6 +24,7 @@ module GwfModule
   use SimModule, only: count_errors, store_error, store_error_filename
   use BaseModelModule, only: BaseModelType
   use MatrixBaseModule
+  use VectorBaseModule
 
   implicit none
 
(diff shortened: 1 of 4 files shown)


src/Model/GroundWaterFlow/gwf3.f90
```diff
@@ -683,13 +684,19 @@ subroutine gwf_ptc(this, kiter, neqsln, matrix, &
         ! get the maximum volume of the cell (head at top of cell)
         v = this%dis%get_cell_volume(n, this%dis%top(n))
         !
-        ! -- calculate the residual for the cell
-        resid = DZERO
+        ! calculate the residual for the cell
+        ! diagonal element
+        resid = matrix%get_diag_value(jrow) * x(jrow)
```
mjr-deltares, line 689: This should probably remain = DZERO, otherwise the diagonal value is counted twice. Let's discuss.

jdhughes-usgs:
Merging after talking to @mjr-deltares who will troubleshoot the rtd issue on his parallel feature branch.

PR #1198, opened 2023-04-07
docs(gwf-sfr.dfn): cross_section keyword is a part of sfrsetting

emorway-usgs:
The `false` setting was leading to an oddity in mf6io.pdf

x-sec


langevin-usgs:
Good catch, @emorway-usgs.

langevin-usgs (approved):

PR #1203, opened 2023-04-14
fix(idm): improved error reporting for unrecognized keywords

mjreno:


langevin-usgs (commented):
This looks great, @mjreno.  One minor request.  Because this will change behavior for the user (in a good way), we should add mention to the release notes in the file [here](https://github.com/MODFLOW-USGS/modflow6/blob/develop/doc/ReleaseNotes/v6.5.0.tex).  Under BASIC FUNCTIONALITY, we should say: In some cases, unrecognized keywords and invalid auxiliary input did not terminate with a useful error message. The program was corrected to provide error handling for these cases.

Also, we might think about adding a general message of some kind to the release notes about IDM in general, and how reporting of some file loading messages is not written to individual model list files but rather it is written to mfsim.lst.  We can think best how to phrase this.

PR #1206, opened 2023-04-17
rfct(par): add parallel ptc

mjr-deltares:
- add calculation of (global) residual to solution instead of model ptc
- extend matrix and vector functionalities to do so

### DIFF
--- a/autotest/test_gwf_ifmod_idomain.py
+++ b/autotest/test_gwf_ifmod_idomain.py
@@ -1,0 +1,390 @@
+import os
+
+import flopy
+import numpy as np
+import pytest
+from framework import TestFramework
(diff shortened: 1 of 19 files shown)


mjr-deltares:
@jdhughes-usgs: I think this is ready to go in (when green), backtracking for parallel will follow in a new PR otherwise this one grows too large.

also @langevin-usgs: I have added a couple of extra tests for idomain situations at the model interface, and for newton. Both for serial (interface model tests) and for parallel. The parallel newton test case will trigger the global PTC calculation that has been implemented.

langevin-usgs (approved):
Great tests.  Looks good to me.

jdhughes-usgs (approved):

PR #1209, opened 2023-04-24
Add `get_value_ptr`

Hofer-Julian:
## Summary

By exposing `get_value_ptr_double` and `get_value_ptr_int` we are strictly speaking not BMI conform.
This PR fixes this aspect by adding `get_value_ptr`.
The corresponding PR on the xmipy side can be found here: https://github.com/Deltares/xmipy/pull/98
If merged, we should coordinate the next xmipy and MODFLOW 6 release

## Impact

- New versions of xmipy will not work with older versions of MODFLOW.
- The xmipy API stays the same.

## Alternative Approach
- Remove  `get_value_ptr_double` and `get_value_ptr_int` completely

### DIFF
--- a/srcbmi/mf6bmi.f90
+++ b/srcbmi/mf6bmi.f90
@@ -630,7 +630,7 @@ contains
     else if (index(mem_type, "INTEGER") /= 0) then
       bmi_status = get_value_ptr_int(c_var_address, c_arr_ptr)
     else
-      write (bmi_last_error, fmt_unsupported_rank) trim(var_name)
+      write (bmi_last_error, fmt_unsupported_type) trim(var_name)
       call report_bmi_error(bmi_last_error)
       bmi_status = BMI_FAILURE
       return
--- a/srcbmi/mf6bmi.f90
+++ b/srcbmi/mf6bmi.f90
@@ -593,6 +593,54 @@ contains
 
   end function get_value_string
 
+
+  !> @brief Copy the value of a variable into the array
+  !!
+  !! The copied variable is located at @p c_var_address. The caller should
+  !! provide @p c_arr_ptr pointing to an array of the proper shape (the
+  !! BMI function get_var_shape() can be used to create it). Multi-dimensional
+  !! arrays are supported.
+  !<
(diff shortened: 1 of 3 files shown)


mjr-deltares:
I think we will have to take this into consideration also in light of related, non-Fortran kernels having their BMI set up currently where we might want to use xmipy for those too. And if xmipy is extended with a connection to a native get_value_ptr function, we would want to have MODFLOW adapted. Not sure if we are instantly going to drop backward compat by removing the Fortran flavored get_value_ptr_double etc. Maybe deprecation is better here.

jdhughes-usgs:
@mjr-deltares should we merge and deal with the deprecation warning after this is in?

srcbmi/mf6bmi.f90
```diff
@@ -593,6 +593,51 @@ function get_value_string(c_var_address, c_arr_ptr) result(bmi_status) &
...
+    end if
+
+    call get_mem_type(var_name, mem_path, mem_type)
+
+    if (index(mem_type, "DOUBLE") /= 0) then
+      bmi_status = get_value_ptr_double(c_var_address, c_arr_ptr)
+    else if (index(mem_type, "INTEGER") /= 0) then
+      bmi_status = get_value_ptr_int(c_var_address, c_arr_ptr)
+    else
+      write (bmi_last_error, fmt_unsupported_rank) trim(var_name)
```
mjr-deltares, line 633: This should be unsupported type, not rank

mjr-deltares (changes requested):

srcbmi/mf6bmi.f90
mjr-deltares: > @mjr-deltares should we merge and deal with the deprecation warning after this is in?

after fixing the error message above, we can do that. But I realized while looking at it again that we also have the get_value_* functions for each of the types, and then also the set_value_* counter parts. We should then probably treat those similarly.

jdhughes-usgs (approved):

mjr-deltares (approved):

PR #1213, opened 2023-04-28
fix(gwf3lak8): error msg should be tripped when input variable connlen equals zero

emorway-usgs:


emorway-usgs:
While working on #1212, discovered that in a LAK test problem that used `EMBEDDEDV`, there was no gw/sw interaction because `connlen` was accidently set to 0.  When looking into it further, there was a trap for this condition, but the line of code:

`if (rval < DZERO) then`

should instead be less than or equal to to match the input instructions:

`if (rval <= DZERO) then`

The error msg thrown now looks like:
example_error_msg

src/Model/GroundWaterFlow/gwf3lak8.f90
```diff
@@ -6322,6 +6326,13 @@ subroutine lak_fill_budobj(this)
       do j = this%idxlakeconn(n), this%idxlakeconn(n + 1) - 1
         n2 = this%cellid(j)
         q = this%qleak(j)
+        lkstg = this%xnewpak(n)
```
langevin-usgs, line 6329: Are these changes supposed to be part of this PR?

Issue #1074, opened 2022-11-04
Simulation file proposal for parallel simulations


jdhughes-usgs:
**Is your feature request related to a problem? Please describe.**
We need a way to define a parallel simulation in the simulation file.

**Describe the solution you'd like**
We have worked out a proposed solution which is shown below. Have attached the modified definition file for reference.

**Describe alternatives you've considered**
This a a modification of the approach used in the [current parallel prototype](https://github.com/verkaik/modflow6-parallel)

**Additional context**
<img width="1033" alt="image" src="https://user-images.githubusercontent.com/2073498/200027094-cf01adb2-023a-49b6-96f7-b332d05c06d6.png">

Descriptions:
<img width="1079" alt="image" src="https://user-images.githubusercontent.com/2073498/200027159-e72823bf-4990-457e-8a73-8cc2c944f4cd.png">

<img width="1066" alt="image" src="https://user-images.githubusercontent.com/2073498/200028060-880b994d-74b2-44a4-8986-73e4313f3b64.png">

<img width="1061" alt="image" src="https://user-images.githubusercontent.com/2073498/200028114-2c720f38-c31e-4113-adf0-6c2faee975e6.png">

<img width="1045" alt="image" src="https://user-images.githubusercontent.com/2073498/200028389-0874d0ec-c515-4d67-b1b3-ce761fa08e35.png">

File:
[sim-nam.dfn.txt](https://github.com/MODFLOW-USGS/modflow6/files/9940402/sim-nam.dfn.txt)

scharlton2:
@jdhughes-usgs -- The 'current parallel prototype' link is dead (404).

jdhughes-usgs:
@scharlton2 actually was the wrong repo. Updating now.

langevin-usgs:
This is part of the evolving parallel development effort, which may lead to a new input file for parallel options.  Closing.

PR #1216, opened 2023-05-10
feat(par): add partition_number parameter to mfsim.nam models block (#1074)

mjreno:


langevin-usgs:
It looks as though support for processor id in the models block of mfsim.nam will require changes to flopy, in addition to changes to the IDM file reader.  My feeling on this is that we should consider putting these parallel settings in another file, as we discussed earlier.    @mjr-deltares, any thoughts?

mjr-deltares:
I have been going back and forth on this a bit... In the same file with an optional column has the advantage that we do not introduce yet another config file, but the alternative with the separate file is probably more future proof. Also it gives a natural location to set parallel options. Maybe we should discuss one more time to converge on this issue.

PR #1222, opened 2023-05-23
Use dependabot to update github actions

Hofer-Julian:
We are quite happy with it at https://github.com/Deltares/Ribasim.
After this is merged you will get pull requests like this one: https://github.com/Deltares/Ribasim/pull/220

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"


w-bonelli (approved):
Thanks @Hofer-Julian. May occasionally need to [ignore](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) actions pinned to a specific version, but I don't think we pin anything now. Does it only bump on major updates? Since the actions convention is to update major tag on minor/patch updates.

Maybe helpful for EC Python repos too, noticed it also works for [reusable workflows](https://github.blog/changelog/2023-03-13-dependabot-updates-support-reusable-workflows-for-github-actions/) not just actions, which would be nice if a shared release workflow [is introduced](https://github.com/MODFLOW-USGS/modflow-devtools/issues/84)

Hofer-Julian:
>Does it only bump on major updates?

I think it updates for every release. You can set `update-types` if you want to change that behavior.

PR #1228, opened 2023-05-26
fix(SfrCrossSectionUtils): vertical channel walls not accounted for in wetted perimeter calcs

emorway-usgs:
- Related to #1172 
- Ran black
- Ran fprettify

### DIFF
--- a/autotest/cross_section_functions.py
+++ b/autotest/cross_section_functions.py
@@ -135,6 +135,31 @@ def wetted_area(
     return area
 
 
+def add_wetted_vert(x, h, depth, vert_neighbs, idx):
+    left_wet_len = 0
+    right_wet_len = 0
+
+    # left side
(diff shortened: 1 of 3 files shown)


emorway-usgs:
As @langevin-usgs pointed out, the wetted perimeter for this case still wasn’t being handled correctly:
vert_sides_case

This PR offers an approach for evaluating whether neighboring segments are vertically aligned and if so, provides a method to account for their contribution to wetted perimeter.  2 new autotests incorporated into `test_gwf_sfr_npoint01.py` for testing vertically-aligned neighbors (in addition to the case above):
new_case
(and its horizontally-flipped counterpart).

langevin-usgs:
Will merge this in for now, but we should probably consider using a composite manning approach for these simpler cross section geometries.

PR #1230, opened 2023-05-31
fix(gwf3lak8): fix shared wetted area in LAK package

emorway-usgs:
-Set shared wetted area between a lake and gw cell to zero when lake stage is below the bottom of a connected cell.  
-See release notes amendment for additional information.
-Ran fprettify
-New autotest to accompany GWE PR will test this issue, current autotests should not be affected.


emorway-usgs:
Related to #1212

langevin-usgs (approved):

PR #1233, opened 2023-06-05
doc(idm):  update mf6io and add release note

mjreno:


langevin-usgs (approved):
This looks great, @mjreno.  I will take another look as we get ready for release.

PR #1235, opened 2023-06-06
refactor(tdis): Calculate exact last delt to avoid accumulation errors with totim

mwtoews:
This PR ensures that `totim` should more accurately track along `perlen` for the end of each stress period, without introducing floating point errors during geometric progression.

For instance, take a simulation with `nper=4`, `nstp=3`, `perlen=7.0` and `tsmult=1.2`. The total duration is 28.0. Using mf6.4.1 (12/09/2022), here are the `totim` values from the end of each stress period:
```
7.000000000000002
14.000000000000004
21.000000000000004
28.0
```
Using the latest development snapshot:
```
6.999999999999998
13.999999999999996
20.999999999999993
28.0
```
awkward! Only the last `totim` is exact since it is substituted by `totalsimtime`. Older MODFLOW releases have similar accumulation errors. The PR changes the last `totim` values to the expected sequence:
```
7.0
14.0
21.0
28.0
```

This is draft for now for several reasons. I'm not sure if there already was an existing time variable that can be used.  And if not, the name `topertim` is perhaps a placeholder for a better name. Also, there is a similar algorithm in `check_tdis_timing` that may similarly need to be adjusted. Also, tests have not been added yet; suggestions welcome to where they should fit -- any simulation with `tsmul != 1.0`.

### DIFF
--- a/autotest/conftest.py
+++ b/autotest/conftest.py
@@ -1,3 +1,4 @@
+import platform
 from pathlib import Path
 
 import pytest
@@ -29,6 +30,17 @@ def should_compare(
 @pytest.fixture(scope="session")
 def bin_path() -> Path:
     return project_root_path / "bin"
+
+
+@pytest.fixture(scope="session")
+def libmf6_path(bin_path) -> Path:
+    ext = {
+        "Darwin": ".dylib",
+        "Linux": ".so",
+        "Windows": ".dll",
+    }[platform.system()]
+    lib_name = bin_path / f"libmf6{ext}"
+    return lib_name
 
 
 @pytest.fixture(scope="session")
--- a/autotest/test_gwf_tdis.py
+++ b/autotest/test_gwf_tdis.py
@@ -1,0 +1,86 @@
+"""Test TDIS package."""
+import flopy
+import numpy as np
+import pytest
+from xmipy import XmiWrapper
+
+
+@pytest.fixture
+def simple_sim(tmp_path):
+    """Create a simple and bogus GWF simulation without TDIS package."""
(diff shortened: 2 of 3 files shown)


langevin-usgs:
@mwtoews, glad to see a fix for an annoyance that has been around quite some time.  This seems like a reasonable way to do it.  Few thoughts:

1. `topertim` seems fine to me.  `tatperstart` was my only other thought.
2. I don't see a great place to add a test of this in one of our existing autotests, so maybe a very short new dedicated test like `test_gwf_tdis.py` based on your example above?  
3. I'm not sure  `check_tdis_timing` needs to change (because it is testing for an obscure case), but maybe you could update it with your revised approach so that it is consistent.
4. We are shooting for a release later this month, and it would be great to get this in.

mwtoews:
Thanks, I'll keep `topertim` for now unless there are clear objections.

I've added tests based on xmipy, which I see is a first in this project (finally?). I'm choosing xmipy over modflowapi only because I'd rather have a lower-level implementation (modflowapi is a higher level).

---

Also, it should be noted that this PR does not fix other unexpected time values, for instance `perlen = 1.1`, which will still have a `totim` sequence for the end of each stress period:

    [0.0, 1.1, 2.2, 3.3000000000000003, 4.4]

Most programming languages will consider these floats `1.1 + 1.1 + 1.1` not equal to `3.3` (only Excel has this correct). This simple issue unfortunately has no simple solution that I can see.

langevin-usgs (approved):

langevin-usgs (approved):
Thanks for this fix, @mwtoews.  Will merge PR now.  We do have some libmf6 tests in the autotest folder, so maybe we will change the name of your test to be consistent with those.

PR #1236, opened 2023-06-07
feat(BMI): Add get/set for logical (bool) scalar values

mwtoews:
This PR adds the following libmf6 symbols: `get_value_bool`, `get_value_ptr_bool`, and `set_value_bool`, supporting only scalar array types (`MemoryManagerModule`'s `mem_setptr` interface only has `setptr_logical`).

Tested locally with a modified XmiWrapper to be submitted later. E.g.
```python
from xmipy import XmiWrapper

mf6 = XmiWrapper("/path/mto/libmf6.so", working_directory="/path/to/sim")
mf6.initialize()
mf6.update()
mf6.get_value('TDIS/ENDOFSIMULATION')  # array([False])
```

Some notes:

- BMI does not seem to have any references to "bool"/"boolean"/"logical", so this is yet another extension
- The variable suffix `_bool` is picked as the middle ground of name choices for the data type, but alternative suggestions welcome. Docstrings keep "logical" Fortran lingo.

### DIFF
--- a/srcbmi/mf6bmi.f90
+++ b/srcbmi/mf6bmi.f90
@@ -515,6 +515,52 @@ contains
 
   end function get_value_int
 
+  !> @brief Copy the logical scalar value into the array
+  !!
+  !! The copied variable us located at @p c_var_address. The caller should
+  !! provide @p c_arr_ptr pointing to a scalar array with rank=0.
+  !<
+  function get_value_bool(c_var_address, c_arr_ptr) result(bmi_status) &
+    bind(C, name="get_value_bool")
+    !DIR$ ATTRIBUTES DLLEXPORT :: get_value_bool
+    ! -- modules
+    use MemorySetHandlerModule, only: on_memory_set
+    ! -- dummy variables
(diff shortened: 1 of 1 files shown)


srcbmi/mf6bmi.f90
```diff
@@ -515,6 +515,52 @@ function get_value_int(c_var_address, c_arr_ptr) result(bmi_status) &
...
+  !> @brief Copy the logical scalar value into the array
+  !!
+  !! The copied variable us located at @p c_var_address. The caller should
+  !! provide @p c_arr_ptr pointing to a scalar array with rank=0.
+  !<
+  function get_value_bool(c_var_address, c_arr_ptr) result(bmi_status) &
+    bind(C, name="get_value_bool")
+    !DIR$ ATTRIBUTES DLLEXPORT :: get_value_bool
+    ! -- modules
+    use MemorySetHandlerModule, only: on_memory_set
```
mjr-deltares, line 527: This is probably a remnant and can be left out

PR #1237, opened 2023-06-08
feat(GWE): Groundwater Energy Transport

emorway-usgs:
Two important areas that need review: 

1. Is the generalized transport (tsp) base class from which `gwt` and `gwe` inherit a good idea?
2. `GweGweConnection` and `GweGweExchange` essentially duplicate `GwtGwtConnection` and `GwtGwtExchange`, respectively.  Perhaps there is a more generalized approach available?

More autotests to follow


jdhughes-usgs:
@w-bonelli the makefile should not include any of the parallel files. There is an exclude files option (`--excludefiles EXCLUDEFILES`), like the extra files `--extrafiles EXTRAFILES`, for pymake.

w-bonelli:
@jdhughes-usgs it looks like some parallel files (e.g. src/Solution/PETSc) [are already in develop](https://github.com/MODFLOW-USGS/modflow6/blob/develop/make/makefile#L12). The `build_makefiles.py` pymake call [does use the exclude files option](https://github.com/MODFLOW-USGS/modflow6/blob/develop/distribution/build_makefiles.py#L54). Should more files be added to `pymake/excludefiles.txt`?

jdhughes-usgs:
> @jdhughes-usgs it looks like some parallel files (e.g. src/Solution/PETSc) [are already in develop](https://github.com/MODFLOW-USGS/modflow6/blob/develop/make/makefile#L12). The `build_makefiles.py` pymake call for mf6 makefile [does use the exclude files option](https://github.com/MODFLOW-USGS/modflow6/blob/develop/distribution/build_makefiles.py#L54). Should more files be added to `pymake/excludefiles.txt`?

I think all of the parallel files that Meson adds if `-Dparallel=true` should be excluded.

emorway-usgs:
This PR has effectively been replaced with #1306 and #1493.  Closing...

PR #1239, opened 2023-06-09
Fix backspace

rleander:


langevin-usgs:
Hi @rleander, thanks for submitting this PR.  We will all be quite happy to get compiling again with the latest OneAPI.  Can you take a look at the few minor checks that are failing.  You will need to run fprettify to format the source code (hopefully the error message will give you a hint on how to do that), and it looks like a few more changes will be required to the utility programs so they can also use the revised utility.  Thank you for your help with this.

mjr-deltares:
Hi @langevin-usgs , @rleander and myself were also discussing if we could somehow contain these changes to the file parsing part and not have them go up and affect RCH or EVT etc. So maybe that will be a minor tweak as well.

mjr-deltares:
Hi @rleander , we noticed you have another PR up on this. Does that replace the current? (and can it then be closed)

mjr-deltares:
discussed with @rleander to close this (for now)

PR #1242, opened 2023-06-09
misc(gwt1dsp.f90): Package naming consistency

emorway-usgs:


emorway-usgs:
Relatively minor, but will help maintain naming consistency within GWT (and eventually with GWE)

langevin-usgs (approved):

PR #1245, opened 2023-06-10
build: refactored meson files for parallel build

jdhughes-usgs:
Changed name of parallel MODFLOW 6 exe and lib to mf6pro and libmf6pro. Skip build of mf5to6 and zbud6 if doing a parallel build. Bumped minimum version of meson to 1.1.0. Updated PARALLEL.md with information on installing OpenMPI and PETSc on MacOS using Homebrew.


mjr-deltares (approved):

mjr-deltares:
Hi @jdhughes-usgs , review ok with one side remark: maybe we can think about extracting the particular dependency configuration into separate meson files for better organization? So we get something like extdeps.default, extdeps.denali, extdeps.win32, etc.

Just to make sure, a lot of the touched files in this PR appear to be from other changesets. Not sure if that will give issues when merging...

jdhughes-usgs:
Replaced by #1258

PR #1248, opened 2023-06-12
minimal change, only in InputOutput and ListReader, keeping changes o…

rleander:
minimal change replacement (low-level) of fortran backspace intrinsic (ifort 2021.9 problem)


mjr-deltares:
discussed with @rleander to close this (for now)

PR #1249, opened 2023-06-12
feat(par): activating mpich in meson script

mjr-deltares:


mjr-deltares:
this goes in with #1245

PR #1250, opened 2023-06-12
refactor(io): remove u8rdcom

langevin-usgs:
The u9rdcom line reader was introduced to replace the u8rdcom line reader, however, u8rdcom was not removed completely.  This PR removes u8rdcom entirely.  This simplification may help with ongoing efforts to remove calls to the Fortran backspace intrinsic so that mf6 will compile with latest versions of ifort.


mjr-deltares:
CC @rleander

PR #1282, opened 2023-06-29
fix(autotests): update use of unit_conversion to length_conversion and time_conversion

emorway-usgs:
Relates to #1154


langevin-usgs:
Could this PR be updated using a rebase instead of a merge.  It's hard to tell what changes are unique to this PR.

emorway-usgs:
@langevin-usgs, I'm surprised it is showing up as a merge, I tried rebasing using the following command from a bash console

rebase

I must not have pushed correctly.  Sorry about this, I'll try again.

emorway-usgs:
A note on replacing the baked-in answers for the 4 smoke tests that were updated with `length_conversion` and `time_conversion` to clean-up deprecation warnings: Previously, when using `unit_conversion` in SFR, it was implied that `unit_conversion` was the product of a length conversion factor and time conversion factor.  This was based on how the input instructions were written.  However, after #1154, the following was introduced to the SFR source code, inside the function `sfr_check_conversion`:

[this%unitconv = this%unitconv * this%lengthconv**DONETHIRD](https://github.com/MODFLOW-USGS/modflow6/pull/1154/files#diff-fc073a3b77a1825be81768e85f662d3f14cfd024c1222501cd968396a0768945R4314-R4319)

Owing to the fact that `lengthconv` is raised to the 1/3 power, the results forthcoming from the smoke tests will be different when using `time_conversion` and `length_conversion` as opposed to `unit_conversion`.

langevin-usgs (commented):
Something doesn't seem quite right here.  Shouldn't we be able to get the same results for the prudic problem?  The simulated concentrations are quite different and don't match any more like they do [here](https://modflow6-examples.readthedocs.io/en/master/_examples/ex-gwt-prudic2004t2.html).  I think we want to produce results that look more or less like this:  
image
Or maybe I'm missing something?

emorway-usgs:
@langevin-usgs, it was me that was missing something.  The factor 1.486 already reflected the fact that a cube root was applied to 3.28084, kind of embarrassing I didn't pick up on that before.  I should've cubed 1.486 when I switched from `unit_conversion` to `length_conversion` originally.  It after the autotests complete on f96818f, this should be good to go.

langevin-usgs (commented):
Hey Eric, this looks much better.  Only one final comment.  Should length_conversion be more precisely specified as 3.28084 instead of the value that you use (3.281379)?  It doesn't matter for the comparisons.

langevin-usgs (approved):

PR #1283, opened 2023-06-29
ci(release): reinit release notes, update dist scripts

w-bonelli:
- reinitialize release notes
- substitute version into folder tree shown in release notes (previously was 6.x.x)
- add OS tag to distribution name substituted into release notes and mf6io
- refactor `update_version.py` with [`packaging.version.Version`](https://packaging.pypa.io/en/stable/version.html#usage) and convert `version.txt` to plain text
  - consistent with flopy and what Python projects [will need to adopt](https://github.com/modflowpy/flopy/pull/1854)
  - replace "+" suffix indicating dev status with ".dev0" to comply with PEP 440 and allow using Python's built-in versioning utils
- use `.mf6minsim` for sample mf6 output in `build_docs.py` instead of ex-gwf-twri01 [from `ex-gwf-twri.py`](https://github.com/MODFLOW-USGS/modflow6-examples/blob/master/scripts/ex-gwf-twri.py)
- add `--repo-owner` option to `build_docs.py` to allow testing on a fork
- add `meson.options` to full distribution
- fix `check_dist.py` example script tests: previously expected run scripts to be in top-level example model dir, but some examples have nested subdirs


mwtoews:
To me, the development version should be `6.4.3dev0`. This would fix the "major" and "minor" positions, and conform to version comparisons ("greater than"). For instance, compare:
```python
from packaging.version import Version

# This is the latest release to compare to
v642 = Version("6.4.2")

v = Version("6.dev4")
print(f"{v.major=} {v.minor=} {v.dev=} {v > v642}")
# v.major=6 v.minor=0 v.dev=4 v > v642=False

v = Version("6.4.3dev0")
print(f"{v.major=} {v.minor=} {v.dev=} {v > v642}")
# v.major=6 v.minor=4 v.dev=0 True
```
and if you want to get fancy, things like setuptools-scm use [git describe](https://git-scm.com/docs/git-describe) for the dev local info based on the git hash
```python
from subprocess import check_output

check_output('git describe --dirty --tags --long --match "*[0-9]*"', text=True)
# 6.4.2-3-g1ce12a2de0

v = Version("6.4.3dev3+g1ce12a2de0")
print(f"{v.major=} {v.minor=} {v.dev=} {v.local=} {v > v642}")
# v.major=6 v.minor=4 v.dev=3 v.local='g1ce12a2de0' True
```

w-bonelli:
Thank you @mwtoews the hope was less frequent need to change develop branch version numbers but I see it was wrong / out of order. Updated to your suggestion

w-bonelli:
Would you suggest any way to achieve something like `v6.4.2+`? Motivation being not to specify the version under development, as it's not known what the next release number will be, but still indicate development has continued?

mwtoews:
From what I've seen, usually the "next release" is just to increment the minor number, so long as it has something like "dev" in it. The actual version number, when published, can jump up to a major version. Often these decisions are made close to the release, depending on things like big new features or breaking changes.

doc/ReleaseNotes/v6.dev4.tex
mwtoews, line 1: should this simply be `develop.tex`? And then renamed to (e.g.) `v6.4.3.tex` on release. Any change here obviously need  to xref to `doc/ReleaseNotes/ReleaseNotes.tex`

distribution/update_version.py
```diff
@@ -181,25 +130,17 @@ def log_update(path, version: Version):
...
-            "label = " + (("'" + version.label + "'") if version.label else "''") + "\n"
-        )
-        f.write("__version__ = '{:d}.{:d}.{:d}'.format(major, minor, micro)\n")
-        f.write("if label:\n")
-        f.write("\t__version__ += '{}{}'.format(__version__, label)")
-        f.close()
-    log_update(version_file_path, version)
-    py_path = project_root_path / "doc" / version_file_path.name.replace(".txt", ".py")
-    shutil.copyfile(version_file_path, py_path)
+        f.write(f"__version__ = '{version}'")
```
mwtoews, line 143:
```suggestion
        f.write(f'__version__ = "{version}"\n')
```

w-bonelli:
@mwtoews after discussion it was decided to bump minor instead of micro, otherwise I believe the plan is to adopt your proposed scheme here, flopy and other related repos. Thanks for the guidance here

PR #1301, opened 2023-07-18
feat(ArrayReaders): add better error trapping for binary input data

jdhughes-usgs:
* modify ArrayReaders to use standard errmsg variable
* standardize store_error calls in ArrayReaders
* modify list reader so cellid is only checked if no error condition
* add binary array input examples to mf6io
* add binary array input file specifications for each discretization type
* add binary list input file specifications
* add description of change to ReleaseNotes


doc/mf6io/framework/array_data.tex
```diff
@@ -121,41 +121,21 @@ \subsubsection{READARRAY Examples}
...
+  STRT
+     OPEN/CLOSE strt.bin (BINARY)  #layers1-4
 \end{lstlisting}
 
-If the (BINARY) keyword is found on the control line, then the file is opened as an unformatted file on unit 99, and the list is read.  There are a number of requirements for using the (BINARY) option for lists.  All stress package lists begin with integer values for the \texttt{cellid} (layer, row, and column, for example).  These values must be represented as integer numbers in the unformatted file.  Also, all auxiliary data must be included in the binary file; auxiliary data must be represented as double precision numbers.  Lastly, the (BINARY) option does not support entry of \texttt{boundname}, and so the BOUNDNAMES option should not be activated in the OPTIONS block for the package.  
\ No newline at end of file
+A consequence of the way binary input files have been implemented in \mf, simulated dependent variable binary output (for example, head and concentration) cannot be used as binary array input for a model. Instead, simulated dependent variable binary output must be processed and split into separate binary files for each layer or combined into a single three-dimensional array.
```
langevin-usgs, line 141:
```suggestion
A consequence of the way binary input files have been implemented in \mf, simulated dependent variable binary output (for example, head and concentration) cannot be used as binary array input for a model. Instead, simulated dependent variable binary output must be processed and split into separate binary files for each layer or combined into a single array equal to the size of the grid (for DIS grids this would be an array equal to NCOL * NROW * NLAY).
```

langevin-usgs (commented):
Nice.  Once small tweak for you to consider.

langevin-usgs (approved):

PR #1302, opened 2023-07-18
feat(bmi): support setting isim_mode prior to bmi initialize

mjreno:


srcbmi/mf6bmi.f90
```diff
@@ -23,19 +23,21 @@ module mf6bmi
...
   use MemoryTypeModule, only: MemoryType
   use MemoryHelperModule, only: create_mem_address
-  use SimVariablesModule, only: simstdout, istdout
+  use SimVariablesModule, only: simstdout, istdout, isim_mode
   use InputOutputModule, only: getunit
   implicit none
 
   integer(c_int), bind(C, name="ISTDOUTTOFILE") :: istdout_to_file = 1 !< output control: =0 to screen, >0 to file
   !DIR$ ATTRIBUTES DLLEXPORT :: istdout_to_file
+  integer(c_int), bind(C, name="ISIM_MODE") :: sim_mode = MNORMAL !< simulation mode MVALIDATE=0, MNORMAL=1, MRUN=2
```
langevin-usgs, line 39: We typically start all integer variables with characters between I and N, though we are not strict about this.  I wonder if we shouldn't call this isim_mode_c_int or something like that to designate that it is different from most variables.  Just an idea.  Otherwise this seems like it should work.  Would probably be good to add a test at some point, even if it's just looking at mfsim.lst for a message that indicates what mode the simulation is in.

mjreno: I verified mfsim.lst logging in my test environment before creating the PR, checking both normal and validate modes.  I agree that testing is a good idea- maybe also coverage for many of the bmi/xmi interfaces that could be added as we work through completeness for those that need updates.

mjreno: @langevin-usgs This might be a simple approach to also expose other command line flags through the shared library.  An alternative for isim_mode might be to remove the command line flag and add the variable to the mfsim.nam options block.  IDM could then set isim_mode in a similar way to how it is currently handling iparamlog (PRINT_INPUT).

mjreno: I rechecked- the "mode=" logging statements are in the model listing files, not mfsim.lst.

PR #1304, opened 2023-07-19
fix(par_nur_has_converged): set mpi_world using get_mpi_world()

jdhughes-usgs:
* add test of ex-gwf-bump with and without being split into two models with the flopy model splitter
* add parallel test of of ex-gwf-bump for one and two processors


autotest/test_gwf_newton_under_relaxation.py
```diff
@@ -0,0 +1,144 @@
...
+delc = ylen / float(nrow)
+extents = (0, xlen, 0, ylen)
+shape2d = (nrow, ncol)
+shape3d = (nlay, nrow, ncol)
+nouter = 50
+ninner = 100
+hclose = 1e-8
+hclose_outer = hclose * 10.
+rclose = 1e-3
+botm = np.loadtxt(pl.Path("data/ex-gwf-bump/bottom.txt")).reshape(shape3d)
```
langevin-usgs, line 31: This may be fine, but I think we have a `project_root_path` approach for accessing data paths.  This is from the prudic example:

```
import os
import sys

import flopy
import numpy as np
import pytest
from conftest import project_root_path
from framework import TestFramework
from simulation import TestSimulation

ex = ["prudic2004t2"]
data_path = project_root_path / "autotest" / "data"
model_path = data_path / "prudic2004test2"
fname = str(model_path / "lakibd.dat")
lakibd = np.loadtxt(fname, dtype=int)
```

autotest/test_gwf_newton_under_relaxation.py
```diff
@@ -0,0 +1,144 @@
...
+chd_spd = [[0, i, 0, H1] for i in range(nrow)]
+chd_spd += [[0, i, ncol - 1, H2] for i in range(nrow)]
+base_heads = flopy.utils.HeadFile(
+    pl.Path("data/ex-gwf-bump/results.hds.cmp")
```
langevin-usgs, line 35: same `project_root_path` issue here

langevin-usgs (commented):
nice

langevin-usgs (approved):

PR #1311, opened 2023-07-25
feat(sfr): add support for CELLIDs with zero values for unconnected reaches

jdhughes-usgs:
* add deprecation waring for use of NONE for unconnected reaches
* update mf6io with SFR changes
* update ReleaseNotes with SFR changes
* add test of new SFR functionality for unconnected reaches


langevin-usgs (commented):
I'm surprised that this worked, because in a preceding call a couple of lines up from the SFR modification there is this:

```
this%igwfnode(n) = this%dis%noder_from_cellid(cellid, this%inunit, &
                                                      this%iout, &
                                                      flag_string=.true.)
```

In looking at `noder_from_cellid` and the call to `nodeu_from_cellid` it would seem that we should have to specify `allowzero' to be .true.  Might be worth running that through the debugger to see why/how it works.

mwtoews:
There is a potential issue with https://github.com/MODFLOW-USGS/modflow6/blob/049db5342e0271bd5ae362caab8376331fc8792e/doc/mf6io/mf6ivar/examples/gwf-sfr-example.dat#L32

which should be changed to something like:
```diff
- 16 1  510 2500.  10    1.81E-03 1077.727    3.0 0.00003  0.03     1   1.0    0 
+ 16 1 5 10 2500.  10    1.81E-03 1077.727    3.0 0.00003  0.03     1   1.0    0 
```
are these example files actually tested in any way?

langevin-usgs:
> There is a potential issue with
> 
> https://github.com/MODFLOW-USGS/modflow6/blob/049db5342e0271bd5ae362caab8376331fc8792e/doc/mf6io/mf6ivar/examples/gwf-sfr-example.dat#L32
> 
> which should be changed to something like:
> 
> ```diff
> - 16 1  510 2500.  10    1.81E-03 1077.727    3.0 0.00003  0.03     1   1.0    0 
> + 16 1 5 10 2500.  10    1.81E-03 1077.727    3.0 0.00003  0.03     1   1.0    0 
> ```
> 
> are these example files actually tested in any way?

Good find, @mwtoews.  These files are not used in any way other than being written as an example in the mf6io guide.

PR #1312, opened 2023-07-26
fix(msvs): msvs project file needs to be updated for backspace change

langevin-usgs:


mjreno:
@langevin-usgs mf6lib.vfproj may also need to be updated

PR #1326, opened 2023-08-04
fix(imscsv): fix issue with csv solver output in parallel

jdhughes-usgs:
* add function to append processor id to file names (`append_processor_id()`)
* update mf6core to use `append_processor_id()` to append processor id to `mfsim.lst`
* add test for inner and outer solver csv output

### DIFF
--- a/autotest/test_par_gwf_ims_csv.py
+++ b/autotest/test_par_gwf_ims_csv.py
@@ -1,0 +1,263 @@
+import os
+
+import flopy
+import numpy as np
+import pytest
+from framework import TestFramework
+from simulation import TestSimulation
+
+# Test for parallel MODFLOW running on two cpus.
+# It contains two coupled models with
+#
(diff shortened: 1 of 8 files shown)


autotest/test_par_gwf_ims_csv.py
```diff
@@ -0,0 +1,263 @@
...
+# global convenience...
+name_left = "leftmodel"
+name_right = "rightmodel"
+
+# solver data
+nouter, ninner = 100, 300
+hclose, rclose, relax = 10e-6, 1e-3, 0.97
+
+
+def get_model(idx, dir):
```
langevin-usgs, line 29: If this is the same (or nearly the same) as what's in `test_par_gwf01.py` is there any way you can import `build_model` from that python file and not reproduce here?  Maybe more trouble than it's worth for now, but might be nice to move in that direction.

PR #1332, opened 2023-08-17
refactor(fmi): introduce FlowModelInterface module/type, use for gwt

w-bonelli:
- factor out `FlowModelInterface.f90`, also used by PRT https://github.com/aprovost-usgs/modflow6/pull/21
- update `gwt1fmi1.f90` to extend `FlowModelInterfaceType`
- pull check for SSM pkg from `fmi_df()` in gwtfmi into `gwt_df()` in gwt, allows reusing parent fmi define routine
- use compact docstrings in gwt and gwtfmi

Tentative &mdash; maybe this is not justified and some duplication in gwt and prt fmi is preferable. Adds ~250 loc here, removes ~1200 with PRT, for net savings of ~1000


langevin-usgs:
@w-bonelli, I like this alot.  I'd say go ahead and make sure the PR is up to date, and then let's bring it on in.

PR #1337, opened 2023-08-29
feat(idm): support list and grid based input for base stress packages

mjreno:


mjreno:
New high level public routines:
- 	idm_df():  define routine (called from mf6core)
- 	idm_rp():  read and repare routine (called from mf6core)
- 	idm_ad():  advance routine (called from NumericalSolution and ExplicitSolution)
- 	idm_da():  deallocate routine (called from mf6core)

Capabilities:
- 	Abstract types for creating static and dynamic loader objects (InputLoadType.f90)
- 	Concrete type for loading traditional (ascii based) input (IdmMf6File.f90)
      - These are used for basic, non-stress packages (e.g. dis, npf, etc)
      - These are also used to load static input (pre period block) from basic stress packages
- 	Concrete types for loading traditional (ascii based) dynamic input
      - Basic stress package grid and list based inputs (StressListInput.f90, StressGridInput.f90)
- 	Extendable:
    - 	Framework is set up to incorporate new static/dynamic loaders.  New static/dynamic loader pairs can be added for additional input sources, e.g. netcdf or some other source.
    - 	Framework is also set up to make it possible to add additional dynamic loaders for traditional inputs.  For example, to add an advanced stress package loader, or add one for each advanced stress package (AsciiInputLoadType.f90)

  Additional detail:
-   Static loading:
    - 	Static loading is effectively being done in this PR in the same way in this PR as how it is being handled in develop.
    - 	Static input in general is copied (e.g. with mem_set_value routine) from the input context to the model context.
-   Dynamic loading:
    - 	Dynamic loading is a new capability with this PR
    - 	IDM allocates and loads based on dfn described parameters, it does not for example aggregate parameters into a 2d bound array
    - 	In general, IDM tries to represent input as provided from the user.  For example, with list based inputs IDM reads and stores cellids.  A cellid will need to be translated to a node (user or reduced) on the model side.
    - 	Timeseries and TAS are managed on the input side:
          - TAS code has been lightly updated for so IDM can use it without providing a DIS instance.  Other users of TAS are not impacted.
          - List input tracks string fields and sets up new links for each period (StructVector.f90)
          - Grid based input also sets up tas links each period
          - No auxiliary multipliers or flow (cell area) calculations are done on the input side- this will need to be managed on the model side.
     - 	While certain parameters may be updated through a copy on the model side (e.g. nbound), many arrays are intended to be directly read from the model side.

mjr-deltares:
Hi @mjreno , nice job, this is a great PR! I think it should go in like this.

mjreno:
Thanks @mjr-deltares, I added a check that can report a more meaningful error when DIS dimensions aren't provided.  I don't plan to add more content unless in response to feedback.

PR #1339, opened 2023-09-01
ci: add gitlab config for automatic denali build

w-bonelli:
- add `gitlab-ci.yml` with denali deploy job, triggered on push to default branch on internal USGS GitLab mirror
- add `update-module.slurm.batch` job to update denali mf6 module
- move hardcoded info to env vars set via git forge (github or gitlab) variables

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,28 @@
+image: ubuntu:20.04
+before_script:
+  # install ssh agent
+  - 'which ssh-agent || ( apt-get update -y && apt-get install openssh-client git -y )'
+  # configure ssh agent
+  - eval $(ssh-agent -s)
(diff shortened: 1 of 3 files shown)


jdhughes-usgs (commented):
Would it be better to put `.gitlab-ci.yml` in a `.gitlab` folder? Don't know if this would work.

w-bonelli:
I'd prefer that too but the [docs](https://docs.gitlab.com/ee/ci/yaml/gitlab_ci_yaml.html) say it needs to be in the root

jdhughes-usgs:
> I'd prefer that too but the [docs](https://docs.gitlab.com/ee/ci/yaml/gitlab_ci_yaml.html) say it needs to be in the root

Ok. Looks good to merge then.

PR #1343, opened 2023-09-06
fix(par): backtracking was out of sync

mjr-deltares:


jdhughes-usgs (approved):
Makes sense if one model backtracks all of them must backtrack.

PR #1344, opened 2023-09-07
fix(gwf3vsc8.f90): fix for the text header in the vsc package's binary output file

### LINKED ISSUE #1342: Bug in exporting viscosity file.
**Describe the bug**
When a viscosity file is written, the description is recorded as '         VISCOSI' instead of '       VISCOSITY'

**To reproduce**
Activate the "  VISCOSITY FILEOUT " option in the Viscosity package and run the model. Then check the description of the data in the file.

**Expected behavior**
The description of the data should be '       VISCOSITY'.

 ** Cause**
In the viscosity package source code, '         VISCOSITY' has 18 characters instead of 16.

        call this%dis%record_array(this%visc, this%iout, iprint, ibinun, &
                                   '         VISCOSITY', cdatafmp, nvaluesp, &
                                   nwidthp, editdesc, dinact)

emorway-usgs:
fixes #1342

### DIFF
--- a/autotest/test_gwf_vsc01.py
+++ b/autotest/test_gwf_vsc01.py
@@ -311,6 +311,18 @@ def eval_results(sim):
             + ex[2]
             + ", but it is not."
         )
+    
+    # Ensure that binary output file is readable (has the correct header)
+    vsc_filerecord = "{}.vsc.bin".format(gwfname)
+    fname = os.path.join(sim.simpath, vsc_filerecord)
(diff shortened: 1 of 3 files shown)


langevin-usgs (approved):
@emorway-usgs, this looks good.  Thanks for the fix.  At some point we should decide if fixes to VSC should be described in the release notes in the stress packages section or the internal flow packages section.  Internal flow packages seems like perhaps a better fit.

PR #1350, opened 2023-09-14
refactor(fmi): reduce unneeded duplication

wpbonelli:
Clean up some rough edges from #1332

- remove unneeded duplication in (de)allocation routines
- deduplicate `read_packagedata()` &mdash; introduce `handle_default_packagedata()` procedure, called from the default case
  - fmi implementation just stores error
  - gwt fmi overrides to allocate/setup advanced package budget
- use compact docstrings in gwt fmi


src/Model/ModelUtilities/FlowModelInterface.f90
```diff
@@ -435,10 +435,10 @@ subroutine initialize_bfr(this)
...
 
   !> @brief Advance the budget file reader
-  !!
-  !! Advance the budget file reader by reading the next chunk
-  !! of information for the current time step and stress period.
-  !!
+  !
+  ! Advance the budget file reader by reading the next chunk
+  ! of information for the current time step and stress period.
+  !
```
emorway-usgs, line 441: From what I can tell, the majority (if not all) of the MF6 code base uses "!!" for lengthier docstrings as opposed to a single "!".  I think it would be best to stay consistent in this.

emorway-usgs (approved):
Thanks for cleaning this up, Wes.

src/Model/ModelUtilities/FlowModelInterface.f90
wpbonelli: think this was caused by running fprettify without the config file arg &mdash; fixed

langevin-usgs (approved):

PR #1356, opened 2023-09-20
docs(mf6io): standardize variable name of the feature IDs in SFR

emorway-usgs:
Initially adjusting SFR only to see if it causes any hiccups with FloPy on ci


emorway-usgs:
@jdhughes-usgs I thought it would be best to test the change for a single advanced package prior to making changes in all 4 advanced packages.  Thus, I think this PR can go in pending your review and I'll submit a PR that has changes for LAK, MAW, and UZF in a separate PR.  Also note that if acceptable, you may also want to accept [PR #1956](https://github.com/modflowpy/flopy/pull/1956) over on the FloPy repo, which will overwrite the SFR dfn file there as well to keep the dfn files in sync with the change made in this PR.

jdhughes-usgs (approved):

langevin-usgs (approved):

PR #1357, opened 2023-09-20
ci(msvs): test msvs build

### LINKED ISSUE #953: Add check to CI that MSVS files are valid
The MSVS files for MODFLOW 6 and the utilities are not checked as part of the CI testing.  Consequently, we may break them without knowing as we add new functionality.

For the Windows CI testing, we should add a workflow that builds mf6 and the utilities using the MSVS project files.

wpbonelli:
- close #953
- fix path to `tsp1.f90` in `mf6core.vfproj`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
         run: python .github/common/fortran_format_check.py
 
   build:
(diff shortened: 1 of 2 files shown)


wpbonelli:
closed in favor of #1365

PR #1358, opened 2023-09-20
feat(csub): add z-displacement support for disu grids

jdhughes-usgs:


src/Model/ModelUtilities/DiscretizationBase.f90
```diff
@@ -1505,4 +1506,37 @@ function get_area(this, node) result(area)
...
+    integer(I4B), intent(in) :: node !< cell node number
+    integer(I4B), intent(in) :: m !< connected cell node number
+    ! -- local
+    real(DP) :: area_node
+    real(DP) :: area_conn
+    ! ------------------------------------------------------------------------------
+    !
+    ! -- calculate the cell area fraction
+    area_node = this%area(node)
+    area_conn = this%con%hwva(m)
```
langevin-usgs, line 1533: I don't think m is the connected cell number.  This must be the connection number.  Should probably replace `m` with a different variable and indicate it is connection number and not a node number.

langevin-usgs (commented):
Looks nice.  I like how the test works for three different grid types.  Small nit to pick in the BaseDisModule.

PR #1359, opened 2023-09-21
docs(mf6io): standardize variable name of the feature ID in the LAK & LKT dfn files

emorway-usgs:
Decided to submit changes associated with standardizing the feature ID variable name to help make the reviews easier.  It turns out there are 20+ files that are altered with this update.


doc/mf6io/mf6ivar/tex/gwf-lak-desc.tex
```diff
@@ -173,7 +173,7 @@
...
+\item \textcolor{blue}{\texttt{rate}---real or character value that defines the extraction rate for the lake outflow. A positive value indicates inflow and a negative value indicates outflow from the lake. RATE only applies to outlets associated with active lakes (STATUS is ACTIVE). A specified RATE is only applied if COUTTYPE for the OUTLETNO is SPECIFIED. If the Options block includes a TIMESERIESFILE entry (see the ``Time-Variable Input'' section), values can be obtained from a time series by entering the time-series name in place of a numeric value. By default, the RATE for each SPECIFIED lake outlet is zero.}
```
langevin-usgs, line 176: Wasn't this change already in develop?  Did you merge instead of rebase?

.github/workflows/ci.yml
```diff
@@ -8,13 +8,37 @@ on:
...
+      - '**.jpg'
+      - '**.jpeg'
+      - '**.png'
+      - '**.bbl'
+      - '**.bib'
+      - 'doc/**.dat'
+      - 'doc/**.ipynb'
+      - 'doc/**.py'
+      - 'doc/**.sh'
+      - 'doc/**.xlsx'
```
langevin-usgs, line 41: What are these changes?

doc/mf6io/mf6ivar/tex/gwf-lak-desc.tex
emorway-usgs: @langevin-usgs Not sure what's going on.  The only intentional thing I've done at this point is create a new branch and commit to it.  It's unclear to me how or when this crept in to the PR.  Maybe my local develop is behind?  I'll try running a rebase, but I definitely have not run any merges.

.github/workflows/ci.yml
emorway-usgs: @langevin-usgs, Initially, these PRs were not triggering the full suite of ci, these are some changes that @wpbonelli and I came up with so that PRs associated with adjustments to the documentation generally would not trigger the full suite of ci; however, if adjustments to a .dfn file are made, the thinking is that the full suite of ci should be run.  We tried coming up with a more reduced way of implementing this logic, but in the end it required what's shown.

langevin-usgs: Okay, that's fine.  Just curious.

doc/mf6io/mf6ivar/tex/gwf-lak-desc.tex
emorway-usgs: I can confirm that #1354 was merged; however, the change that it contained does not appear on the develop branch as of [66231fe](https://github.com/MODFLOW-USGS/modflow6/tree/66231febbe705b12add7c5c7eb047a248acc5c2e)

In the image below, you can see that on line 176, the change in #1354 isn't showing up?  I've never seen this before.  I guess this means the change still needs to be implemented?

change_not_there

langevin-usgs: Maybe you didn't update your fork/develop with the changes in upstream/develop

emorway-usgs: Figured out the issue here.  In #1354 I updated the .dfn file, but then failed to run mf6ivar.py, which means that the change didn't propagate out to the .tex file.  Since if ran mf6ivar.py in this workflow, part of what should have been included in #1354 is included with this PR.

langevin-usgs (approved):

PR #1365, opened 2023-09-23
ci(msvs): workflow to verify msvs project files are complete

langevin-usgs:
Simple hack to compare the files in the source folder with the files listed in the msvs project files.  If they do not match, then issue an assertion error.  If this works out, then could also implement for zone budget and mf5to6.

What do you think @wpbonelli?  Would this be a simpler (albeit less strict) alternative to #1357?


wpbonelli:
Nice idea @langevin-usgs. I'm not sure if an MSVS build in CI will pan out, or whether the overhead will be worth it if it does (e.g. may need not just hpckit but basekit too, adding even more time to oneapi setup on Windows, which is already ~30min before caching)

.github/common/msvs_vfproj_check.py
```diff
@@ -0,0 +1,71 @@
...
+    for f in root.iter('File'):
+        s = f.attrib["RelativePath"]
+        s = s.replace("\\", "/")
+        s = s.replace("../", "")
+        fpath = Path(s)
+        msvs_files.append(fpath)
+    return msvs_files
+
+
+def check_files(name, src_files, msvs_files):
```
wpbonelli, line 30: to avoid the loops this could be something like

```python
s, m = set(src_files), set(msvs_files)
diff = s ^ m
from pprint import pformat
assert not any(diff), (
    f"{name} src files don't match msvs project file\n"
    f"=> symmetric difference:\n{pformat(diff)}\n"
    f"=> src - msvs:\n{pformat(s - m)}\n"
    f"=> msvs - src:\n{pformat(m - s)}"
)
```

wpbonelli (approved):

.github/common/msvs_vfproj_check.py
langevin-usgs: Nice, @wpbonelli.  Change was made.

PR #1366, opened 2023-09-24
feat(idm): update CHD to source static and dynamic list input from the input context

mjreno:


mjreno:
@langevin-usgs I'm thinking about adding CHD updates to this PR if you aren't opposed to it.  Without some initial package there are some missing variables in the dfn2f90 generated code that the BoundaryPackage extensions refer to.

langevin-usgs:
Sure, I say go for it.  This PR is pretty small in it's current form, so the CHD additions make sense here.

PR #1382, opened 2023-10-02
feat(ArrayHandlers): add 2D array expansion interface

wpbonelli:
Support expanding 2D integer & double allocatable arrays. For use by PRT for particle bookkeeping.

*unit testing demo/discussion moved to #1387*


langevin-usgs (commented):
The changes to array handlers looks good.  I think we should probably have a discussion about the use of test-drive and make sure this is the direction we want to head.  It looks promising.

PR #1384, opened 2023-10-02
refactor: generalize routine to disable development features

wpbonelli:
Support disabling development features from anywhere, optionally associated with an input file. Previously only the block parser had this facility. Allows e.g. disabling models under development.

* add `dev_feature()` routine to `SimModule`
* call from block parser, specifying file unit


src/Utilities/BlockParser.f90
```diff
@@ -583,8 +584,7 @@ subroutine DevOpt(this)
     if (IDEVELOPMODE == 0) then
       errmsg = "Invalid keyword '"//trim(this%laststring)// &
                "' detected in block '"//trim(this%blockname)//"'."
-      call store_error(errmsg)
-      call this%StoreErrorUnit()
+      call dev_feature(errmsg, this%iuext)
```
langevin-usgs, line 587: Nice idea to break this out.  I see that we are double-checking IDEVELOPMODE both in parser%DevOpt() and in dev_feature().  Maybe we can move that entirely into dev_feature()?

langevin-usgs (commented):
Nice idea.

src/Utilities/BlockParser.f90
wpbonelli: Removed IDEVELOPMODE usage from BlockParser. Thanks, I overlooked that.

PR #1386, opened 2023-10-04
refactor(tsp): Elevate FMI to generalized transport class

emorway-usgs:
First step in generalized transport refactor: elevate FMI


src/Model/TransportModel/tsp1.f90
```diff
@@ -18,8 +23,530 @@ module TransportModelModule
...
+      write (errmsg, '(a)') &
+        'Initial conditions (IC6) package not specified.'
+      call store_error(errmsg)
+    end if
+    if (indis == 0) then
+      write (errmsg, '(a)') &
+        'Discretization (DIS6 or DISU6) package not specified.'
+      call store_error(errmsg)
+    end if
+    if (inmst == 0) then
```
langevin-usgs, line 371: Should MST be included in TSP?

langevin-usgs (approved):
@emorway-usgs, this looks good.  There was only one thing that stuck out for me and that was mention of MST in the TSP code.  I will approve and merge but this is something that may need tending.

PR #1391, opened 2023-10-10
refactor(tsp): Elevate ADV to generalized transport class

emorway-usgs:


mjr-deltares:
Nice job moving this forward @emorway-usgs!

I see no issues with this PR. I know there are changes in the draft #1237 that will need some further consideration. I like this approach of merging in functionality in conceptual units. My suggestion from parallel/coupling perspective would be to continue like this respecting the following sequence:

1. merge in TSP base and standalone GWE + sufficient test coverage
2. merge in GWE-GWE with interface model coupling + tests
3. extend Virtual Data to include GWE for parallel + tests (reused from 2)

Since the interface models should directly extend their representative model types (GWT or GWE in this case), we will have some duplication that cannot be solved by type extension. Something to discuss along the way, I guess.

PR #1392, opened 2023-10-11
refactor(tsp): Elevate IC to generalized transport class

emorway-usgs:
Relates to #1386 and #1391


langevin-usgs (approved):
Looks good.  Thanks, @emorway-usgs.

PR #1399, opened 2023-10-13
refactor(tsp): Elevate MVT to generalized transport class

emorway-usgs:
Related to:
#1386
#1391 
#1392 
#1396 
#1398


src/Model/TransportModel/tsp1mvt1.f90
```diff
@@ -58,24 +59,21 @@ module GwtMvtModule
...
-                    gwfmodelname2, fmi2)
-! ******************************************************************************
-! mvt_cr -- Create a new initial conditions object
-! ******************************************************************************
-!
-!    SPECIFICATIONS:
-! ------------------------------------------------------------------------------
+  !> @brief Create a new mover transport object
+  !<
+  subroutine mvt_cr(mvt, name_model, inunit, iout, fmi1, eqnsclfac, & ! kluge note: does this need tsplab?
```
langevin-usgs, line 68: would be good to figure out if tsplab is needed and remove the kluge note

emorway-usgs: Great catch, @langevin-usgs. The tsplab class has been removed, this comment should have come out with it. Fixed with 99049d1

langevin-usgs (approved):

PR #1404, opened 2023-10-16
refactor(tsp): Elevate CNC to generalized transport class

emorway-usgs:


emorway-usgs:
Relates to:
#1386
#1391
#1392
#1396
#1398
#1399 
#1400

src/Model/TransportModel/tsp1cnc1.f90
```diff
@@ -241,40 +233,37 @@ subroutine cnc_ck(this)
...
     class(MatrixBaseType), pointer :: matrix_sln
     ! -- local
-! --------------------------------------------------------------------------
     !
-    ! -- return
+    ! -- Return
     return
   end subroutine cnc_fc
 
+  !> @brief Calculate flow associated with constant concentration/tempearture
```
langevin-usgs, line 258:
```suggestion
  !> @brief Calculate flow associated with constant concentration/temperature
```

langevin-usgs (commented):
Looks good, @emorway-usgs.  We might want to think about renaming the TSP object to something other than CNC since this the second C refers to concentration.  Maybe it could be something like SDV (specified dependent variable)?  But we would still want the user to specify CNC for transport, and maybe something else for temperature?

emorway-usgs:
> We might want to think about renaming the TSP object to something other than CNC ...maybe it could be something like SDV (specified dependent variable)? But we would still want the user to specify CNC for transport, and maybe something else for temperature?

@langevin-usgs,  I like `SDV` and will plan to implement that as part of this PR.  

Regarding a potential alternative name for the GWE equivalent of `CNC`, what about `CTP` (constant temperature), or would it be better to simply use `CNT` (only change the last letter)?

langevin-usgs (commented):
This looks good, but I wonder how this is going to play with the new Input Data Model (IDM) routines.  We are still going to want separate DFN files for CNC and CNT, but I'm not sure how this is going to work IDM.  Perhaps you should look at the constant head package for flow and see if you can implement IDM for CNC as part of this PR.  Since IDM is fully implemented now for CHD, WEL, DRN, RIV, GHB, EVT, and RCH, it would make sense it is implemented now for CNC as well.  Any thoughts?

emorway-usgs:
@langevin-usgs, Thanks for bringing the IDM issue to my attention.  Agreed on the separate .dfn files for CNC and CNT.  From a documentation (mf6io.pdf) perspective alone it makes sense to have 2 separate .dfn files.  

Also agree that it makes sense to look into IDM support for SDV now.  I'll have a look at what was done to CHD for supporting IDM and may need to get together with @mjreno for the finer points.

emorway-usgs:
Going in a different direction with CNC.  Going to leave it as part of the GroundWaterTransport model for the time being.  A parallel package, constant temperature ("CNT"), will be introduced with GWE.  After CNT is introduced, can look into creating a new class that contains the common code among CHD, CNC, & CNT that each of these packages could inherit from.

PR #1405, opened 2023-10-17
feat(InputOutput.f90): Introduce padl, a string manipulation function

emorway-usgs:
Part of the generalized transport refactoring is that the dependent variable name varies depending on which type of model is implemented (GWT vs GWE).  To accommodate this in the standard output listing file, `padl` assists with writing well-formatted messages.


src/Utilities/InputOutput.f90
```diff
@@ -1197,6 +1197,22 @@ subroutine get_ijk(nodenumber, nrow, ncol, nlay, irow, icol, ilay)
     !
     return
   end subroutine get_ijk
+  
+  !> @brief Function for string manipulation
+  !<
+  function padl(str, width) result(res)
```
jdhughes-usgs, line 1203: I think a more descriptive name should be used (maybe `str_pad_left`)

emorway-usgs: @jdhughes-usgs, I'm all for descriptive names.  I installed the suggested name with bda8094

PR #1412, opened 2023-10-26
build(meson): add support for NetCDF in meson build system

jdhughes-usgs:
* rename PROFESSIONAL to EXTENDED version
* add function to test if __WITH_NETCDF__ define
* refactor is_pro to is_extended function that evaluates if either WITH_PETSC or WITH_NETCDF are defined
* add -Dextended=true meson option that will eventually replace -Dparallel=true

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,6 +9,7 @@ dependencies:
   - appdirs
   - filelock
   - fprettify
+  - fortran-language-server
   - jupytext
   - matplotlib
   - meson>=1.1.0
--- a/meson.build
+++ b/meson.build
@@ -173,7 +173,7 @@ if is_extended_build
   else
     # For CI testing only; Windows not yet supported
     nc_dir = meson.project_source_root() / '..' / 'ncf' / 'netcdf-fortran-4.6.1' / 'fortran'
-    netcdf = fc.find_library('netcdff', dirs: [ nc_dir ], required : true, static : false)
+    netcdf = fc.find_library('netcdff', dirs: [ nc_dir ], required : false, static : false)
     #nc_incdir = include_directories([ nc_dir ])
   endif
   if netcdf.found()
--- a/meson.build
+++ b/meson.build
@@ -105,14 +105,17 @@ elif fc_id == 'intel-llvm-cl'
 endif
 
 # parallel build options
-is_parallel_build = get_option('parallel')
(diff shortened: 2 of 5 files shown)


environment.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
   - appdirs
   - filelock
   - fprettify
+  - fortran-language-server
```
wpbonelli, line 12: [this](https://github.com/fortran-lang/fortls#fortls-vs-fortran-language-server) makes it sound like `fortls` supersedes `fortran-language-server` but I have not used the latter so not sure.

jdhughes-usgs: The intent was to add the fortran-language-server to the development environment. Will change to fortls.

jdhughes-usgs: Actually it seems the modern fortran plugin for Visual Studio Code does not work with fortls but needs fortran-language-server (which has the same `fortls` exe name). See

https://stackoverflow.com/questions/72153608/fortran-language-server-fortls-error-in-vscode-windows

wpbonelli: Had a quick look at the fortls repo, it seems there are autodiscovery issues. Pointing vscode to it in workspace settings has worked if I recall, but that does not seem like it should be necessary and maybe there are other problems.

Sorry to lead astray, will look deeper in coming weeks

meson.build
```diff
@@ -163,7 +166,23 @@ if is_parallel_build
...
   extra_cmp_args += [ '-D__WITH_MPI__']
   with_mpi = true
+
+  # find netcdf
+  if build_machine.system() != 'windows'
+    netcdf = dependency('netcdf', language : 'fortran', required : false)
+  else
+    # For CI testing only; Windows not yet supported
+    nc_dir = meson.project_source_root() / '..' / 'ncf' / 'netcdf-fortran-4.6.1' / 'fortran'
+    netcdf = fc.find_library('netcdff', dirs: [ nc_dir ], required : true, static : false)
```
mjreno, line 176: required should be set false here to avoid a meson abort

mjreno (approved):

meson.build
jdhughes-usgs: eventually we need to decide if we are going to allow the `extended` version to cherry pick including petsc or netcdf

mjr-deltares (approved):

PR #1413, opened 2023-10-27
feat(gwt1cnc1idm.f90): add IDM support for CNC

emorway-usgs:


src/Model/GroundWaterTransport/gwt1cnc1.f90
```diff
@@ -486,4 +487,33 @@ subroutine cnc_rp_ts(this)
...
+    ! -- dummy variables
+    class(GwtCncType), intent(inout) :: this !< BndExtType object
+    integer(I4B), intent(in) :: col
+    integer(I4B), intent(in) :: row
+    ! -- result
+    real(DP) :: bndval
+    !
+    select case (col)
+    case (1)
+      bndval = this%tspvar(row)
```
mjreno, line 506: This should actually be reporting the scaled quantity when iauxmultcol is greater than 0.  The chd package needs to be updated as well (it was the first and it was an oversight), you could considering that update in this PR.  See ghb/drn/riv for comparison as those should all report the scaled quantity.

mjreno (approved):
Looks great @emorway-usgs.

src/Model/GroundWaterTransport/gwt1cnc1.f90
emorway-usgs: Resolved with 2babaa8.  Need to see how the CI turns out.

emorway-usgs: CI looks to be passing

src/Model/GroundWaterTransport/gwt1cnc1.f90
```diff
@@ -18,9 +22,12 @@ module GwtCncModule
   character(len=LENFTYPE) :: ftype = 'CNC'
   character(len=LENPACKAGENAME) :: text = '             CNC'
   !
-  type, extends(BndType) :: GwtCncType
+  type, extends(BndExtType) :: GwtCncType
+
+    real(DP), dimension(:), pointer, contiguous :: tspvar => null() !< constant head array
```
langevin-usgs, line 27:
```suggestion
    real(DP), dimension(:), pointer, contiguous :: tspvar => null() !< constant concentration array
```

src/Model/GroundWaterTransport/gwt1cnc1.f90
```diff
@@ -115,24 +121,28 @@ subroutine cnc_allocate_arrays(this, nodelist, auxvar)
       this%ratecncin(i) = DZERO
       this%ratecncout(i) = DZERO
     end do
+    ! -- set constant head array input context pointer
+    call mem_setptr(this%tspvar, 'TSPVAR', this%input_mempath)
```
langevin-usgs, line 125: Presumably, you cal this tspvar instead of concentration because you are planning to move this to the general TSP model?

src/Model/GroundWaterTransport/gwt1cnc1.f90
```diff
@@ -160,36 +172,42 @@ subroutine cnc_rp(this)
...
     !
     ! -- Process each entry in the constant concentration cell list
     do i = 1, this%nbound
       node = this%nodelist(i)
-      cb = this%bound(1, i)
+      if (this%iauxmultcol > 0) then
+        cb = this%tspvar(i) * this%auxvar(this%iauxmultcol, i)
+      else
+        cb = this%tspvar(i)
+      end if
```
langevin-usgs, line 210: Shouldn't this be replaced with `cb = conc_mult(i)`

src/Model/GroundWaterTransport/gwt1cnc1.f90
```diff
@@ -199,79 +217,69 @@ subroutine cnc_ad(this)
...
-! ------------------------------------------------------------------------------
+      &"('Specified dependent variable boundary ',i0, &
+      &' conc (',g0,') is less than zero for cell', a)"
     !
     ! -- check stress period data
     do i = 1, this%nbound
       node = this%nodelist(i)
       ! -- accumulate errors
-      if (this%bound(1, i) < DZERO) then
+      if (this%tspvar(i) < DZERO) then
```
langevin-usgs, line 243: I think instead of tspvar, you should be using conc_mult() everywhere instead, but you should verify with @mjreno.

langevin-usgs (changes requested):
Might be a couple of things to fix

src/Model/GroundWaterTransport/gwt1cnc1.f90
emorway-usgs: Yes, the thought behind `tspvar` (short for _transport variable_) is that eventually it will support either concentration or temperature.  We can rename now or later if you think that would be helpful.

src/Model/GroundWaterTransport/gwt1cnc1.f90
mjreno: Thanks @langevin-usgs, that is right.  As conc_mult() was added it should replace every other use of tspvar where the scaled value should be used- and we should use this same pattern broadly for all other package files that define an analogous *_mult() function.

emorway-usgs: addressed with a12332c

src/Model/GroundWaterTransport/gwt1cnc1.f90
emorway-usgs: addressed in a12332c

src/Model/GroundWaterTransport/gwt1cnc1.f90
emorway-usgs: addressed in a12332c

langevin-usgs (approved):
Nice!

PR #1418, opened 2023-10-31
ci: add simple gwfgwf exchange model with sfr mover

jdhughes-usgs:
* fix sfr gwtgwt test to add mover to exchange files

### DIFF
--- a/autotest/simulation.py
+++ b/autotest/simulation.py
@@ -164,10 +164,6 @@ class TestSimulation:
         return
 
     def setup_comparison(self, src, dst, testModel=True):
-        # evaluate if comparison should be made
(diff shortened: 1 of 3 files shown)


jdhughes-usgs:
@wpbonelli I found out with `make_comparision=False` that it does not run the comparision model. I added a `run_comparision` keyword to allow a comparsion model to be run and then `make_comparison` can be use to stop the comparison. Not sure if you can think of a better way to do this.

wpbonelli (approved):
Looks good. I can take a closer look at the test framework and see if it can be simplified after this goes in.

mjr-deltares (approved):

PR #1420, opened 2023-11-03
fix(gwf3chd8.f90): account for auxiliary multiplier value in IDM modifications to CHD

emorway-usgs:
- PR addresses a comment @mjreno left here: https://github.com/MODFLOW-USGS/modflow6/pull/1413#discussion_r1380567761
- Updated docstrings
- Ran fprettify

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3chd8.f90
+++ b/src/Model/GroundWaterFlow/gwf3chd8.f90
@@ -37,6 +37,7 @@ module ChdModule
     procedure :: allocate_arrays => chd_allocate_arrays
     procedure :: define_listlabel
     procedure :: bound_value => chd_bound_value
+    procedure :: head_mult
     ! -- methods for observations
     procedure, public :: bnd_obs_supported => chd_obs_supported
     procedure, public :: bnd_df_obs => chd_df_obs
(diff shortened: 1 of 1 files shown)


src/Model/GroundWaterFlow/gwf3chd8.f90
```diff
@@ -245,56 +223,51 @@ subroutine chd_ck(this)
...
     !
     ! -- check stress period data
     do i = 1, this%nbound
       node = this%nodelist(i)
       bt = this%dis%bot(node)
       ! -- accumulate errors
-      if (this%head(i) < bt .and. this%icelltype(node) /= 0) then
+      if (this%head_mult(i) < bt .and. this%icelltype(node) /= 0) then
         call this%dis%noder_to_string(node, nodestr)
         write (errmsg, fmt=fmtchderr) i, this%head(i), bt, trim(nodestr)
```
mjreno, line 234: The error message should probably report the value that failed, head_mult() in this case.

langevin-usgs (approved):

src/Model/GroundWaterFlow/gwf3chd8.f90
emorway-usgs: Addressed this concern in #1424.

PR #1423, opened 2023-11-03
fix(GeomUtil): fix point_in_polygon, expand unit tests

wpbonelli:
There were a few bugs in the `point_in_polygon()` routine added in #1407 (off-by-n errors and premature termination). The test included in that PR was only a very basic case (unit square) and so did not catch these. Fix the issues and add test cases for other geometries:

* triangle
* irregular pentagon

Tests are also refactored a bit for easier future expansion and nicer messages on failure


langevin-usgs (approved):
Hey @wpbonelli, this looks good to me.  Nice to see the additional tests.

PR #1424, opened 2023-11-03
fix(gwf3chd8.f90): ensure that error message reports the correct value

emorway-usgs:
Accompanies #1420, addresses [this concern](https://github.com/MODFLOW-USGS/modflow6/pull/1420#discussion_r1382100049) from @mjreno.


langevin-usgs:
@mjreno added this to #1421, so will close this.

PR #1425, opened 2023-11-03
refactor(IC): integrate initial conditions pkg with IDM

wpbonelli:
* use IDM to load flow and transport model initial conditions (no options, just grid data)
* rebased after #1421


src/Model/GroundWaterFlow/gwf3.f90
```diff
@@ -309,8 +309,8 @@ subroutine gwf_ar(this)
     class(BndType), pointer :: packobj
 ! ------------------------------------------------------------------------------
     !
-    ! -- Allocate and read modules attached to model
-    if (this%inic > 0) call this%ic%ic_ar(this%x)
+    ! -- Load modules attached to model
+    if (this%inic > 0) call this%ic%ic_load(this%x)
```
langevin-usgs, line 313: Unless there is some reason we need to wait until AR, I think we should call this ic_load method as soon as possible, maybe toward the end of ic_create?

src/Model/GroundWaterFlow/gwf3ic8.f90
```diff
@@ -74,175 +69,77 @@ subroutine ic_ar(this, x)
...
-    ! -- Read options
-    call this%read_options()
-    !
-    ! -- Read data
-    call this%read_data()
+    ! -- read grid data
+    call this%source_griddata()
     !
-    ! -- Assign x equal to strt
+    ! -- assign starting head
```
langevin-usgs, line 79: Maybe we couldn't do this in ic_cr, which is why we are calling ic_load() as part of the AR calls.  Maybe we can keep an ic_ar() routine and assign strt to model%x in it.  I know it is not a "allocate and read" action, but maybe we keep the names consistent for now and consider renaming them later.

src/Model/GroundWaterTransport/gwt1.f90
```diff
@@ -269,7 +269,7 @@ subroutine gwt_ar(this)
     ! -- Allocate and read modules attached to model
     call this%fmi%fmi_ar(this%ibound)
     if (this%inmvt > 0) call this%mvt%mvt_ar()
-    if (this%inic > 0) call this%ic%ic_ar(this%x)
+    if (this%inic > 0) call this%ic%ic_load(this%x)
```
langevin-usgs, line 272: Keep the ar() name now for consistency?

langevin-usgs (commented):
Thank you for doing this, @wpbonelli

src/Model/GroundWaterFlow/gwf3ic8.f90
```diff
@@ -74,175 +69,77 @@ subroutine ic_ar(this, x)
...
+    map => null()
+    if (this%dis%nodes < this%dis%nodesuser) map => this%dis%nodeuser
+    !
+    ! -- set values
+    call mem_set_value(this%strt, 'STRT', this%input_mempath, map, found%strt)
+    !
+    ! -- ensure STRT was found
+    if (.not. found%strt) then
+      write (errmsg, '(a)') 'Error in GRIDDATA block: STRT not found.'
+      call store_error(errmsg)
```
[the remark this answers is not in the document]
wpbonelli: Revised to 

```
call store_error(errmsg, terminate=.false.)
call store_error_filename(this%input_fname)
```

I noticed `store_error` defaults to terminate false while `store_error_filename` defaults true. Should these be consistent or is this deliberate?

src/Model/GroundWaterFlow/gwf3.f90
wpbonelli: After trying this out, it seems this needs to come after define (`*_df`), I think because that's where discretization sources its info, so I left it in AR for now.

src/Model/GroundWaterTransport/gwt1.f90
wpbonelli: restored to `ic_ar`

src/Model/GroundWaterFlow/gwf3ic8.f90
wpbonelli: cf https://github.com/MODFLOW-USGS/modflow6/pull/1425#discussion_r1382401652

langevin-usgs (approved):

PR #1427, opened 2023-11-06
feat(par): add mover functionality to parallel capabilities

mjr-deltares:


langevin-usgs (approved):
Hey @mjr-deltares, this looks really nice, especially considering how difficult it seemed back when we first started talking about it.  I see you've added the parallel test of the xmvr.  Should we think about adding another test in the future for the bidirectional flow case?

mjr-deltares:
Hi @langevin-usgs , good point! Let's try and put it in with this PR. I will set up something, might contact you offline.

mjr-deltares:
@langevin-usgs : I have included that test we talked about. Also sync'ed qavailable such that the MVR flow rates printed to the list file are correct.

langevin-usgs (approved):

PR #1428, opened 2023-11-06
refactor(tsp): Elevate APT to generalized transport class

emorway-usgs:
Relates to:
#1386
#1391
#1392
#1396
#1398
#1399
#1400

- Includes updates in SFT, LKT, MWT, & UZT owing to the changes in APT
- Ran fprettify
- Updated meson
- Ran build_makefiles


src/Model/GroundWaterTransport/gwt1mwt1.f90
```diff
@@ -102,9 +98,12 @@ subroutine mwt_create(packobj, id, ibcnum, inunit, iout, namemodel, pakname, &
     character(len=*), intent(in) :: namemodel
     character(len=*), intent(in) :: pakname
     type(TspFmiType), pointer :: fmi
+    real(DP), intent(in), pointer :: eqnsclfac !< governing equation scale factor
+    character(len=LENVARNAME), intent(in) :: dvt !< For GWT, set to "CONCENTRATION" in TspAptType
+    character(len=LENVARNAME), intent(in) :: dvu !< For GWT, set to "mass" in TspAptType
+    character(len=LENVARNAME), intent(in) :: dvua !< For GWT, set to "M" in TspAptType
```
langevin-usgs, line 104: I think you should change all these character definitions to character(len=*), which means they will come in at whatever size they were passed in with.  That way if we ever change the size, we'd only need to do it in the calling module, instead of in here too.  I'd change it for the other advanced packages too.

src/Model/TransportModel/tsp1apt1.f90
```diff
@@ -76,7 +77,11 @@ module GwtAptModule
     integer(I4B), pointer :: ibudcsv => null() !< unit number for csv budget output file
     integer(I4B), pointer :: ncv => null() !< number of control volumes
     integer(I4B), pointer :: igwfaptpak => null() !< package number of corresponding this package
-    real(DP), dimension(:), pointer, contiguous :: strt => null() !< starting feature concentration
+    integer(I4B), pointer :: idxprepak => null() !< budget-object index that precedes package-specific budget objects
+    integer(I4B), pointer :: idxlastpak => null() !< budget-object index of last package-specific budget object
+    real(DP), dimension(:), pointer, contiguous :: strt => null() !< starting feature concentration (or temperature)
+    real(DP), dimension(:), pointer, contiguous :: ktf => null() !< thermal conductivity between the apt and groundwater cell
```
langevin-usgs, line 83: Something seems off if we need to store thermal conductivity for the generic transport package class.

src/Model/TransportModel/tsp1apt1.f90
```diff
@@ -1068,7 +1039,8 @@ subroutine apt_ot_dv(this, idvsave, idvprint)
         end if
         this%dbuff(n) = c
       end do
-      call ulasav(this%dbuff, '   CONCENTRATION', kstp, kper, pertim, totim, &
+      write (text, '(a)') str_pad_left(this%depvartype, 16)
```
langevin-usgs, line 1042: Is there a way to avoid hard coding this "16"?  Should this be one of our constants?

aprovost-usgs (commented):
Took a quick look through all the files, and though there are multiple files involved, overall the changes seem straightforward and appropriate.  There is some coding and some inline comments that specifically reference gwe/energy/temperature, which maybe goes a little beyond the minimum necessary to elevate gwt for now, but I don't see much harm in that.  I assume many of the inline comments are temporary, and you'll scan for those later and delete them or convert them to "real" comments before release.

src/Model/TransportModel/tsp1apt1.f90
```diff
@@ -2381,27 +2273,27 @@ subroutine apt_fill_budobj(this, x)
...
+    !
     ! -- FROM MOVER
     if (this%idxbudfmvr /= 0) then
       idx = idx + 1
       nlist = this%ncv
       call this%budobj%budterm(idx)%reset(nlist)
-      do n1 = 1, nlist
-        q = this%qmfrommvr(n1)
+      do j = 1, nlist
+        call this%apt_fmvr_term(j, n1, n2, q) ! kluge note: don't really need to do this in apt_fmvr_term now, since no override by uze
```
langevin-usgs, line 2283: Are these changes ready for develop?

src/Model/TransportModel/tsp1apt1.f90
```diff
@@ -2353,23 +2244,24 @@ subroutine apt_fill_budobj(this, x)
...
       call this%apt_accumulate_ccterm(n1, q, ccratin, ccratout)
     end do
-
-    ! -- individual package terms
-    call this%pak_fill_budobj(idx, x, ccratin, ccratout)
-
+    !
+    ! -- skip individual package terms for now and process them last
+    ! -- in case they depend on the other terms (as for uze)
+    idx = this%idxlastpak
```
langevin-usgs, line 2250: Something seems complicated here.

src/Model/TransportModel/tsp1apt1.f90
```diff
@@ -2463,53 +2356,96 @@ subroutine apt_stor_term(this, ientry, n1, n2, rrate, &
...
+    integer(I4B), intent(inout) :: n1
+    integer(I4B), intent(inout) :: n2
+    real(DP), intent(inout), optional :: rrate
+    real(DP), intent(inout), optional :: rhsval
+    real(DP), intent(inout), optional :: hcofval
+    !
+    ! -- Calculate MVR-related terms
+    n1 = ientry
+    n2 = n1
+    if (present(rrate)) rrate = this%qmfrommvr(n1) ! presumably in terms of energy already for heat transport???
```
langevin-usgs, line 2423: Is this ready to go into develop?

src/Model/TransportModel/tsp1apt1.f90
```diff
@@ -76,7 +77,11 @@ module GwtAptModule
     integer(I4B), pointer :: ibudcsv => null() !< unit number for csv budget output file
     integer(I4B), pointer :: ncv => null() !< number of control volumes
     integer(I4B), pointer :: igwfaptpak => null() !< package number of corresponding this package
-    real(DP), dimension(:), pointer, contiguous :: strt => null() !< starting feature concentration
+    integer(I4B), pointer :: idxprepak => null() !< budget-object index that precedes package-specific budget objects
+    integer(I4B), pointer :: idxlastpak => null() !< budget-object index of last package-specific budget object
```
langevin-usgs, line 81: Unclear why we need these.

src/Model/TransportModel/tsp1apt1.f90
emorway-usgs: After discussing this over with @aprovost-usgs, we concluded that it is ready for the develop branch. In essence, several lines were packaged into this subroutine (`apt_fmvr_term`). The "kluge note" has been removed.

src/Model/TransportModel/tsp1apt1.f90
emorway-usgs: These are part of a modification by @aprovost-usgs for the benefit of UZE. In formulating the matrix problem, the UZE balance equation is folded into the corresponding cell balance equation. In the subsequent budget calculations, the heat exchanged between the UZE feature and the cell (by virtue of their coming to thermal equilibrium) is "backed out" as the residual of the UZE budget terms. Therefore, the UZE-package-specific budget calculations, which include this residual calculation as the final step, need to come after the generic, APT-based budget calculations. Although (currently) only UZE should require this reordering, it is simplest to do it that way for all advanced packages and should pose no problem.  `idxprepak` and `idxlastpak` are indices that are maintained so that the package specific calculations can be initially skipped and then done at the end.

src/Model/TransportModel/tsp1apt1.f90
emorway-usgs: Please see response [here](https://github.com/MODFLOW-USGS/modflow6/pull/1428#discussion_r1385581552).

src/Model/TransportModel/tsp1apt1.f90
emorway-usgs: After discussing with @aprovost-usgs, yes, this is ready to go into develop.  We wanted to leave a note for ourselves to double check values when bringing in GWE later.

src/Model/TransportModel/tsp1apt1.f90
emorway-usgs: Great catch, `ktf` should not have been in this...bad code merging.  Fixed in 7345130

src/Model/GroundWaterTransport/gwt1mwt1.f90
emorway-usgs: Good to know that's how that works.  Have changed in 7345130, changes applied to all advanced packages.

src/Model/TransportModel/tsp1apt1.f90
langevin-usgs: Alternatively, UZE could have overridden parent routines and stored these indices on its own.  I'm fine with whatever way you want to go, especially if you think this is general and possibly useful beyond UZE in the future.

aprovost-usgs: @langevin-usgs Agree there are several ways this could be done. The bottom-line requirement is that all budget terms be in place before UZE calculates its residual.

-- The way it's done now, duplication of code is minimized, but it's admittedly a bit klugey to skip the package-specific calls and then comeback to them at the end.

-- Riffing off of the way it's done now, the kluginess could be eliminated by maybe reordering the indexing of the budget terms in APT so the package-specific terms come last. Haven't thought through the possible side effects.

-- As you suggested, UZE could override and handle all the budget calculations, doing them in whatever order it needs, at the expense of some code duplication.

-- Could break out the residual calculation as a separate step at the end. This is actually a more generally applicable approach, since the alternatives above will work only as long as UZE is the only APT package that needs to calculate its residual. So the idea would be to complete (in any order) all the generic and package-specific budget calculations other than the residuals, then at the end call package-specific residual-calculating routines (which would do nothing for most packages). Not sure if we'll ever have another residual-dependent package like UZE, but maybe it would be best to go ahead set it up that way anyway? The cost would be an additional subroutine baked into the APT infrastructure. (It's been a while since I looked deeply at this part of the code, but I believe it would be ok to have multiple residual-dependent packages, since the residual represents an "internal" transfer within each package and shouldn't affect other packages.)

Happy to do it whatever way is best for MF6 in the long run.

PR #1437, opened 2023-11-10
refactor(idm): relocate model package lists to model modules

mjreno:


langevin-usgs (approved):
This looks great, @mjreno.  I see there is still one place where a new model needs to be added (ModelPackageInputModule), but I like where this is located.

PR #1446, opened 2023-11-16
refactor: introduce MathUtilModule, move is_same, add tests

wpbonelli:
The `GenericUtilitiesModule` currently has a function `is_same()` which uses a symmetric relative comparison for floating point equality. Move it to new `MathUtilModule` as `GenericUtilitiesModule` is mostly simulation related and depends on `SimVariablesModule` etc. Modify floating point equality routine and rename to `is_close()`.

Motivation

Relative comparisons [get more sensitive to difference near zero](https://peps.python.org/pep-0485/#behavior-near-zero) holding epsilon equal, and if either value is 0, equality is only reported with epsilon >= 1. This can cause difficulties if it's unknown whether values to compare will be close to 0.

A common approach is to fall back to an absolute comparison like Python's `math.isclose()` and `numpy.isclose()`. Introduce function `is_close()` doing this.

Python stdlib math and numpy use different abstol defaults: `math.isclose()` sets the default absolute tolerance to zero, motivation that there is no universally reasonable default. The new `is_close()` here follows this but I'm not sure that's the best way.

An option is also provided to toggle asymmetric comparison, e.g. with a known reference value to which relative difference should be scaled

Places (near-)zero comparisons are or might be made:

* `utils/mf5to6/src/Preproc/ObsBlock.f90` checks explicitly against zero
* `TimeSeries` and `TimeArraySeries` time comparisons
* coordinate comparisons in `GridSorting.f90`
* `point_in_polygon()` could support fuzzy boundaries 

Reference

* https://peps.python.org/pep-0485/
* https://stackoverflow.com/a/32334103/6514033
* https://numpy.org/doc/stable/reference/generated/numpy.isclose.html
* https://github.com/fortran-lang/stdlib/blob/master/src/stdlib_math_is_close.fypp


langevin-usgs (commented):
@wpbonelli, this looks good to me.  If we switch to `is_close()` do we still need `is_same()`?  And I'm sure it's fine, but maybe in `TestMathUtil` we should explicitly verify that comparison with DNODATA works as this is used in csub?

PR #1464, opened 2023-11-23
refactor: simplify autotest framework

wpbonelli:
Changes include

* fold `TestSimulation` into `TestFramework`
* rename/refactor properties
  * `simpath` -> `workspace`
  * `build_function` -> `build` (and inject in `__init__`)
  * `exfunc` -> `check`
  * `exe_dict` -> `targets`
  * `require_failure` -> `xfail`
  * `action` -> `compare`
  * fold switches `run_comparison`, `make_comparison`, `mf6_regression` into `compare` (optional str), accepts:
    * "auto" picks comparison type based on workspace contents, equivalent to previous "compare" (default)
    * "run_only"  equivalent to previous `run_comparison=True and make_comparison=False`
    * "mf6_regression" equivalent to previous `mf6_regression=True`
    * `None` disables comparison
* standardize test function names, step toward automatic discovery?
  * `build_models` creates models
  * `check_output` checks results
* framework doesn't need to know about test case indices, remove `idxsim` and just use lambda if `build_models()` needs a test case index
* support `pathlib.Path` or `str` paths
* run black/isort, format test description docstrings
* remove unused functions from `common_regression.py`
* remove [`pytest-cases`](https://smarie.github.io/python-pytest-cases/) from tests and as a test dependency
* fix hardcoded gridgen left in `conftest.py` from #1460 

Merge considerations

- requires modflow-devtools >= 1.1 (update `environment.yml` to specify this)
- should probably wait for #1467 and any other open PRs touching autotests, so any conflicts can be resolved here

Future directions

- support multiple comparison/regression models? (e.g. compare current mf6 with previous mf6 and mf2005, or PRT with mp6 and mp7)
- explicit bookkeeping for sim/model workspaces? (replace implicit expectation that test and framework agree)
- autodiscovery of test cases following e.g. [this pattern](https://docs.pytest.org/en/7.1.x/example/nonpython.html#yaml-plugin)?


autotest/common_regression.py
```diff
@@ -160,15 +161,15 @@ def setup_comparison(namefile, dst, remove_existing=True):
                 print("Could not make " + dst)
         # clean directory
         else:
-            print(f"cleaning...{dst}")
+            print("Cleaning ", dst)
```
jdhughes-usgs, line 164: The change is fine but I am wondering if there is a reason to move away from f-strings

autotest/common_regression.py
```diff
@@ -440,10 +444,9 @@ def get_mf6_comparison(src):
         comparison type
 
     """
-    action = None
     # Possible comparison - the order matters
-    optcomp = (
-        "compare",
+    OPTCOMP = (
```
jdhughes-usgs, line 448: I like the change to uppercase for defined data.

autotest/common_regression.py
```diff
@@ -612,60 +619,49 @@ def get_mf6_mshape(disfile):
...
-        lines = f.readlines()
-
-    for line in lines:
-
+    srcdir = os.path.dirname(namefile)
+    mdl_files = []
+    pkg_files = []
+    out_files = []
+    pkg_keys = ["TDIS6", "GWF6", "GWT", "GWF6-GWF6", "GWF-GWT", "IMS6"]
+    mdl_keys = ["GWF6", "GWT"]
```
jdhughes-usgs, line 640: Should `GWT` be `GWT6` like `GWF6`. Also maybe it would be clearer if `mdl_keys` was `model_keys`.

jdhughes-usgs (approved):

autotest/common_regression.py
wpbonelli: I meant to move this to a top-level variable so other modules could import it, will do so if you agree

autotest/common_regression.py
wpbonelli: fixed

wpbonelli:
There are some breaking changes here, so if this merges, minor refactoring will be needed for any unmerged work which includes new tests. Thinking of GWE @emorway-usgs and SNF @langevin-usgs. I am happy to wait for other work to merge and then update tests here, or consult on the new pattern if this goes in first

langevin-usgs (approved):
Looks good.  @wpbonelli, please merge it in if you are satisfied.

Issue #1477, opened 2023-11-30
The latest MF6 won't read multilayer binary initial head file correctly


ougx:
For a DIS grid, `STRT` is read as `DOUBLE1D` which will skip headers of lower layers in the HDS file. `6.4.2` works fine.

mjreno:
Hi @ougx, IC input processing was changed in [this](https://github.com/MODFLOW-USGS/modflow6/commit/1b220d470ccf1f52781b6c23751d9308b7ab9d0a) commit, earlier versions might be ok.  If possible can you provide an example input file that will cause the problem?  Thanks for reporting!

PR #1479, opened 2023-11-30
refactor: update 1D ArrayReaders to read multiple binary headers

mjreno:
Intended to address #1477.

### DIFF
--- a/src/Utilities/ArrayRead/Double1dReader.f90
+++ b/src/Utilities/ArrayRead/Double1dReader.f90
@@ -5,7 +5,7 @@ module Double1dReaderModule
   use BlockParserModule, only: BlockParserType
   use SimVariablesModule, only: errmsg
   use SimModule, only: store_error, store_error_unit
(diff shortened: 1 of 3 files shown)


ougx:
@mjreno thanks for fixing the issue. it seems that the new fix did not process the `IPRN` flag. It could be useful if user wants to check the initial head values.

mjreno:
@ougx there is summary input logging for this array if PRINT_INPUT is specified in mfsim.nam options.  But I agree that IPRN is not supported by this reader as it has been historically- the IPRN setting is read but isn't acted upon.  I'll look into that more- I'm not remembering at the moment if that was intentional.

mjreno:
This PR is being closed as this behavior is not currently supported.  According to supported options in mf6io, the file should be split to contain layer variables OR the layers should be combined into a single array variable (with a single header) in the current file.

IPRN is a separate issue and is currently not supported for the reader that is currently loading IC (and other IDM integrated packages).

PR #1484, opened 2023-12-06
refactor(ListModule): trim dependencies, add defaults, improve docstrings

wpbonelli:
* remove `SimModule` dependency and `sim_message()`/`stop_with_error()` usages
  * minimize dependencies of generic utilities on simulation modules so latter can use former without circularity
  * on programmer error, just print error message and `error stop 1`
* use `associated(x, y)` as default `isEqual` procedure for `ContainsObject`
* more descriptive docstrings, use compact format

### DIFF
--- a/src/Utilities/List.f90
+++ b/src/Utilities/List.f90
@@ -1,12 +1,11 @@
 module ListModule
-  ! -- ListType implements a generic list.
   use KindModule, only: DP, I4B
   use ConstantsModule, only: LINELENGTH
-  use GenericUtilitiesModule, only: sim_message, stop_with_error
   implicit none
   private
-  public :: ListType, ListNodeType, isEqualIface, arePointersEqual
+  public :: ListType, ListNodeType, isEqualIface
 
+  !> @brief A generic heterogeneous doubly-linked list.
   type :: ListType
     ! -- Public members
     type(ListNodeType), pointer, public :: firstNode => null()
@@ -65,8 +64,7 @@ module ListModule
 
 contains
 
-  ! -- Public type-bound procedures for ListType
-
+  !> @brief Append the given item to the list
   subroutine Add(this, objptr)
     ! -- dummy variables
     class(ListType), intent(inout) :: this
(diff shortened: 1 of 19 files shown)


src/Utilities/List.f90
```diff
@@ -285,17 +254,15 @@ subroutine InsertAfter(this, objptr, indx)
         followingNode%prevNode => newNode
         this%nodeCount = this%nodeCount + 1
       else
-        write (line, '(a)') 'Programming error in ListType%insert_after'
-        call sim_message(line)
-        call stop_with_error(1)
+        print *, 'Programming error in ListType%insert_after'
+        error stop 1
```
langevin-usgs, line 258: We might need to check further on the ramifications of this change.  I'm thinking stop_with_error() and its use of call exit() was used intentionally, perhaps for API behavior.

src/Utilities/List.f90
```diff
@@ -304,7 +271,8 @@ subroutine InsertBefore(this, objptr, targetNode)
     type(ListNodeType), pointer :: newNode => null()
     !
     if (.not. associated(targetNode)) then
-      stop 'Programming error, likely in call to ListType%InsertBefore'
+      print *, "Programming error in ListType%InsertBefore"
+      error stop 1
```
langevin-usgs, line 275: Interesting this is in here.  This must be a remnant from a while back.

src/Utilities/List.f90
wpbonelli: Since gnu, intel, and flang [all support](https://flang.llvm.org/docs/Extensions.html) an `exit()` intrinsic subroutine maybe it makes sense to standardize on that directly, and `stop_with_error()` would be unnecessary

One drawback of `error stop` is it only supports a variable status code in f2018, while `exit()` already does

langevin-usgs: Would it make sense to keep `stop_with_error()` in place and use it for all programming errors?  It could remain the same or we could pass it a message.  If we ever needed to change the behavior, we would only need to change it one place?  Or we could get rid of it and just call exit().

wpbonelli: > Would it make sense to keep stop_with_error() in place and use it for all programming errors?  It could remain the same or we could pass it a message. If we ever needed to change the behavior, we would only need to change it one place?

That seems good to me. Maybe it could live in a separate module so generic utilities don't depend on simulation-related things?

langevin-usgs: Sounds good to me!

PR #1486, opened 2023-12-06
feat(idm): subpackage support and advanced package groundwork

mjreno:


mjreno:
This PR will be split into a few smaller requests- the first is #1517.

PR #1492, opened 2023-12-08
chore(tdis.f90): cleanup docstrings in tdis class

emorway-usgs:
- Also, converted what looks like an old MF2K5 style (or earlier?) function into MF6 style.  
- Moves some constants specified in tdis to the Constants module

### DIFF
--- a/src/Timing/tdis.f90
+++ b/src/Timing/tdis.f90
@@ -226,7 +226,7 @@ contains
   !> @brief Reset delt and update timing variables and indicators
   !!
   !! This routine is called when a timestep fails to converge, and so it is
-  !! retried using a smaller time step (deltnew). 
+  !! retried using a smaller time step (deltnew).
   !<
   subroutine tdis_delt_reset(deltnew)
     ! -- modules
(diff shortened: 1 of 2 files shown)


langevin-usgs:
@emorway-usgs, is it possible there is a breaking change in here causing the failure?

emorway-usgs:
@langevin-usgs yes, I tried to remove the goto statement by altering the logic just a bit.  I suspect that's the source of the failure.  I'll try to figure out what happened.

src/Timing/tdis.f90
```diff
@@ -314,7 +289,7 @@ subroutine tdis_set_delt()
       delt = tsmult(kper) * delt
     end if
     !
-    ! -- return
+    ! -- Return
     return
   end subroutine tdis_set_delt
```
emorway-usgs, line 295: There appears to be an commented-out function ('tdis_set_delt_std') in tdis.f90 that I can remove as part o this PR, if you like?  The original line numbers look to be 321-373.  The new line numbers under this PR equate to 296-348

langevin-usgs (approved):

Issue #1500, opened 2023-12-13
Incomplete description of laksetting in LKT package


rbwinst-usgs:
**Describe the bug**
The description of laksetting in LKT package omits EXT-INFLOW as one of the valid values but it is listed as an example.

laksetting—line of information that is parsed into a keyword and values. Keyword values that can be used
to start the LAKSETTING string include: STATUS, CONCENTRATION, RAINFALL, EVAPORATION,
RUNOFF, and AUXILIARY. These settings are used to assign the concentration of associated with the corresponding
flow terms. Concentrations cannot be specified for all flow terms. For example, the Lake Package
supports a “WITHDRAWAL” flow term. If this withdrawal term is active, then water will be withdrawn from
the lake at the calculated concentration of the lake.
STATUS <status>
CONCENTRATION <concentration>
RAINFALL <rainfall>
EVAPORATION <evaporation>
RUNOFF <runoff>
EXT-INFLOW <ext-inflow>
AUXILIARY <auxname> <auxval>

emorway-usgs:
Confirmed that EXT-INFLOW [appears in `lkt_set_stressperiod`](https://github.com/MODFLOW-USGS/modflow6/blob/e2cdd259cb0b8358a2aa3bb5ce1929cdb91587ec/src/Model/GroundWaterTransport/gwt1lkt1.f90#L1125)

PR #1503, opened 2023-12-14
fix(gwt-lkt.dfn): typo

emorway-usgs:
Addresses #1500


langevin-usgs:
@emorway-usgs, if you indicate "Close #1500" then the issue will be automatically closed when the PR is merged.

Issue #1502, opened 2023-12-13
Incorrect description of mwtsetting in MWT package


rbwinst-usgs:
**Describe the bug**
The description of mwtsetting lists STATUS, CONCENTRATION, RAINFALL, EVAPORATION,
RUNOFF, and AUXILIARY as valid keywords. The correct keywords are STATUS, CONCENTRATION, RATE, and AUXILIARY.

mwtsetting—line of information that is parsed into a keyword and values. Keyword values that can be used
to start the MWTSETTING string include: STATUS, CONCENTRATION, RAINFALL, EVAPORATION,
RUNOFF, and AUXILIARY. These settings are used to assign the concentration associated with the corresponding
flow terms. Concentrations cannot be specified for all flow terms. For example, the Multi-Aquifer
Well Package supports a “WITHDRAWAL” flow term. If this withdrawal term is active, then water will be
withdrawn from the well at the calculated concentration of the well.
STATUS <status>
CONCENTRATION <concentration>
RATE <rate>
AUXILIARY <auxname> <auxval>

emorway-usgs:
[Had a look in](https://github.com/MODFLOW-USGS/modflow6/blob/e2cdd259cb0b8358a2aa3bb5ce1929cdb91587ec/src/Model/GroundWaterTransport/gwt1mwt1.f90#L888) `mwt_set_stressperiod` and confirmed that 'RATE' (plus the other usual items: 'STATUS', 'CONCENTRATION', and 'AUXILIARY' found in `apt_set_stressperiod`) are all that is available in `<mwtsetting>`

PR #1504, opened 2023-12-14
fix(gwt-mwt.dfn): typo

emorway-usgs:
Addresses #1502


langevin-usgs (approved):

langevin-usgs:
Thanks for this, @emorway-usgs

PR #1509, opened 2023-12-15
refactor(ArrayHandlers): respect lower bound when resizing

wpbonelli:
* previously array expansion routines assumed lower bound of 1 &mdash; e.g. an array bounded from 0 to 4 and requested to expand by 1 would be shifted to 1-6
* update expansion routines to respect existing lower bound
* add some test cases
* arrays with non-default bounds are likely rarely used (especially due to [potentially inconsistent argument-passing behavior](https://fortran-lang.discourse.group/t/allocatable-arrays-and-lower-bounds/1087/31)), but are occasionally useful (motivated by PRT)

### DIFF
--- a/autotest/TestArrayHandlers.f90
+++ b/autotest/TestArrayHandlers.f90
@@ -1,6 +1,7 @@
 module TestArrayHandlers
   use KindModule, only: I4B, DP, LGP
-  use testdrive, only: error_type, unittest_type, new_unittest, check, test_failed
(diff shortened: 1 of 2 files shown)


wpbonelli:
I added tests for `ExtendPtrArray` here too (though this interface is not currently used), and they pass as independent programs, but fail when run with test-drive, not sure why yet. The tests are skipped pending a resolution.

PR #1517, opened 2023-12-20
extend(idm): update fortran dfns for advanced and subpackages

mjreno:
These updates are forward looking and are intended to support input loading of advanced and subpackages.  Advanced packages are marked as such with the already existing `# package-type advanced-stress-package` tagline in dfn files.  A new tag `# mf6 subpackage <subp>` identifies one subpackage- any number of these can be added to a dfn.  The subpackage identifier follows the naming convention for dfn files themselves, so to add TVK to NPF as a subpackage the entire tagline would be `# mf6 subpackage utl-tvk`.

A subpackage tagline should be added to a package only if both packages are already IDM integrated.  This means that each package does not use a parser to read it's own input file but rather sources it's input from the input context.


mjreno:
Closing as these changes have been added to PR #1581.

PR #1518, opened 2023-12-20
refactor(memory): update mem_reallocate to copy optional interface

mjreno:
Update:
The mem_reallocate interface has been updated to include an optional copy argument.  It has been refactored to copy by default, however, this can be overriden with the optional arg.  The interface will now throw an error if the array size has shrunk and copy is set to TRUE.  The mem_reset interface has been removed.

----------------

The MemoryManager mem_reallocate interface for the most part assumes a size increasing reallocation and then copies existing array values into the new array.  This new interface is meant to reallocate to any size (including 0) and to never copy existing values.   The existing mem_reallocate interface could probably be updated to alternatively provide this same behavior so this PR represents one way forward.

This is needed by IDM when Dynamic list based array data is loaded across periods that do not have an explicit dimension- TVK and TVS are examples.  In these cases, **maxbound** implicitly is the number of reduced nodes, however to reduce memory overhead it might preferred to reallocate these arrays to the size needed for each period.  This interface supports the idea of resetting allocated memory from the previous period.

### DIFF
--- a/src/Model/GroundWaterFlow/gwf3disu8.f90
+++ b/src/Model/GroundWaterFlow/gwf3disu8.f90
@@ -255,8 +255,8 @@ contains
         this%yc(noder) = this%cellxy(2, node)
       end do
     else
-      call mem_reallocate(this%xc, 0, 'XC', this%memoryPath)
-      call mem_reallocate(this%yc, 0, 'YC', this%memoryPath)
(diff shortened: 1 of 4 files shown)


mjr-deltares (approved):
Hi @mjreno , I see no obstacles but I would like to discuss this offline to get the gist of it.

mjreno:
Closing as these changes have been added to PR #1581.

PR #1520, opened 2023-12-22
test(framework): drop support for *.cmp file comparisons

wpbonelli:
* https://github.com/MODFLOW-USGS/modflow6-testmodels/pull/10
* remove support for testmodel `*.cmp` reference solution files
* support for mf2005/etc comparisons preserved
* fix default mf6 regression executable selection
* misc cleanup

### DIFF
--- a/autotest/common_regression.py
+++ b/autotest/common_regression.py
@@ -4,20 +4,14 @@ from pathlib import Path
 from typing import List, Tuple
 from warnings import warn
 
-COMPARE_PATTERNS = (
-    ".cmp",
+COMPARE_PROGRAMS = (
     "mf2005",
-    "mf2005.cmp",
     "mfnwt",
-    "mfnwt.cmp",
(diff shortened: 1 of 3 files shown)


langevin-usgs (commented):
I can't remember the last time I generated a *.cmp file for comparison purposes, but I wonder if we are ready to completely remove that capability.  It is possible that a new testmodel would run successfully with the develop version of mf6 but not with the released version or previous release.  In that case a *.cmp file might be useful.  Granted, we would probably write the new test in the autotest flopy format, so maybe I'm overthinking this. With that said,  we should remove unused code if possible -- just wanted to raise the potential issue.

wpbonelli:
@langevin-usgs if something like a compare file is needed in the future, I think we could use pytest plugins to do snapshot testing and achieve the same thing, e.g. with https://github.com/tophat/syrupy. I am putting together a demo

PR #1521, opened 2023-12-22
feat(MathUtil): add root-finding functions

wpbonelli:
* moved ~~unaltered~~ from / used by PRT ternary solution for DISV grids
* modified style (removed gotos, explicit local variables)
* unit tested briefly


langevin-usgs (approved):
Maybe update fortran code to more modern style?  Otherwise looks good.

mjr-deltares:
Hi @wpbonelli , I know this one is closed and I am a bit behind, but I thought maybe we can still get rid of the implicit statements to make the code a bit safer. I will explain my thoughts offline.

PR #1531, opened 2023-12-30
fix(FlowModelInterface.f90): re-install if statement for proper messaging

emorway-usgs:
In #1332, wherein some code originally appearing in `tsp1fmi1.f90` was moved to `FlowModelInterface.f90`, the line of code pointed to in the screen grabs below didn't survive the move.  As a result, I was getting some messaging printed to console that was throwing me for a loop while debugging GWE.  That is, when some code in an exchange was being defined it was being reported that "FLOWS ARE ASSUMED TO BE ZERO."  However, that's not the case.  This PR seeks to restore the `if` statement indicated in the screen grabs which should lead to better messaging.  I may be missing something, though?  Also interested in see what the ci comes back with.  This change shouldn't impact any of the autotests.

Original code (prior to #1332):
fmi1

New code (after #1332) is missing the line in question:
fmi2

### DIFF
--- a/src/Model/ModelUtilities/FlowModelInterface.f90
+++ b/src/Model/ModelUtilities/FlowModelInterface.f90
@@ -87,15 +87,17 @@ contains
                     &' VERSION 2, 8/17/2023')"
 
     ! --print a message identifying the FMI package.
-    if (this%inunit /= 0) then
-      write (this%iout, fmtfmi) this%inunit
-    else
-      write (this%iout, fmtfmi0)
-      if (this%flows_from_file) then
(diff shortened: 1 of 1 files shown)


wpbonelli (approved):
I don't recall a reason for removing the outer condition, this just seems like a mistake on my end &mdash; sorry for the lost time @emorway-usgs

emorway-usgs:
The ci seems to be failing on a parallel build.  I've relaunched the ci a number of times in case it is related to the occasional hiccup, but it is persisting?  Hard to believe it is related to this PR since I just reinstated an `if` statement.

wpbonelli:
It is unrelated, it's affecting all PRs at the moment

Issue #1491, opened 2023-12-08
Feature toggling


wpbonelli:
Add a [feature toggle](https://en.wikipedia.org/wiki/Feature_toggle) mechanism to allow models/features to live in the `develop` branch before they are ready for inclusion in official releases. This involves a few capabilities:

- [x] disable dev features in release mode
   - #1384
- ~~build-time model selection~~ unnecessary at this time
   - #1607
   - #1603
   - #1532
- [x] filter code/docs/DFNs/examples for distribution
  - #1745
  - #1746

PR #1532, opened 2023-12-30
feat: configure supported models at build time

wpbonelli:
A potential approach for #1491
* Configure supported models & exchanges at build time via meson options
  * Move model sources to independent folders for easier filtering
  * Move model/exg creation to new modules, generated from templates with [`fypp`](https://github.com/aradi/fypp)
  * rename ConnectionBuilder -> ConnectionFactory, move to top level, generate from template
  * rename SimulationCreate -> Simulation
* Generate IDM files with meson 
  * use [`configure_file()`](https://mesonbuild.com/Reference-manual_functions.html#configure_file) before compiling
  * definition files still created with `dfn2f90.py`, moved to `src/IDM`
  * model input file and consolidated selector file generated with `fypp` in `src/`
  * generic framework files still in `src/Utilities/IDM`
  * `dfn2f90.py` accepts input DFNs via config file, default `utils/idmloader/dfns.txt`
  * `dfn2f90.py` accepts an output directory path
  * rename files, modules and routines for basename templating
    * consistent DFN & IDM file naming
    * remove indices from file names

Considerations

* Template processing is also possible with pure python if not worth adding `fypp`
* To keep support for make/MSVS, [`configure_file()`](https://mesonbuild.com/Reference-manual_functions.html#configure_file) is used instead of [`generator()`](https://mesonbuild.com/Generating-sources.html#using-generator) and IDM/templated files are still under version control. Developers not using meson can generate code manually before build time, like is currently done for IDM
* If support for make/MSVS is ever dropped and meson becomes the standard build tool, could switch to `generator()`, preprocess at build time and remove generated files from versioning, but debugging may become trickier


wpbonelli:
https://github.com/MODFLOW-USGS/modflow6/pull/1652#issuecomment-2049988295

PR #1539, opened 2024-01-04
refactor(model): lift properties, use base in virtual model

wpbonelli:
* move `bndlist`, `dis`, `ibound` from `Numerical/ExplicitModel` to `BaseModel`
* allows using `BaseModel` instead of `NumericalModel` in base virtual model type, eventually supporting explicit virtual models, also simpler templating
* update usage in `SimulationCreate`

### DIFF
--- a/src/Distributed/VirtualGwfModel.f90
+++ b/src/Distributed/VirtualGwfModel.f90
@@ -4,7 +4,7 @@ module VirtualGwfModelModule
   use VirtualDataContainerModule, only: VDC_GWFMODEL_TYPE
   use VirtualModelModule
   use SimStagesModule
-  use NumericalModelModule, only: NumericalModelType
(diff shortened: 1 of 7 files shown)


langevin-usgs (commented):
@wpbonelli, I wonder if moving dis/ibound/bndlist to basemodel is the right strategy.  I think the idea for basemodel is to be entirely generic (and it should probably be abstract).  dis/ibound/bndlist are really constructs of a NumericalModel and I'm not sure they are relevant to all future model types.  With that said, however, there are no immediate plans to implement a model that wouldn't need dis/ibound/bndlist.  So I guess the question is whether to move forward with this as is, or rethink the design.  I'm still thinking that we might be able to eliminate ExplicitModel and instead allow a NumericalModel to be solved by IMS or EMS.  I think this would give us and users some nice flexibility.

PR #1543, opened 2024-01-08
refactor(HashTable): update naming, docstrings, format

wpbonelli:
* consistent naming for routines, variables
* use compact format for docstrings, update some language
* add minimal unit test


langevin-usgs (approved):
very nice

PR #1546, opened 2024-01-10
test(List): add minimal unit tests for ListType

wpbonelli:
* add a few exploratory unit tests written during other work

### DIFF
--- a/autotest/TestList.f90
+++ b/autotest/TestList.f90
@@ -1,0 +1,254 @@
+module TestList
+  use KindModule, only: I4B
+  use testdrive, only: error_type, unittest_type, new_unittest, check
+  use ConstantsModule, only: LINELENGTH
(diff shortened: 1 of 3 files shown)


autotest/TestList.f90
```diff
@@ -0,0 +1,254 @@
...
+    call list%Add(p)
+
+    ! check count
+    call check(error, list%Count() == 1, "count should be 1")
+    if (allocated(error)) return
+
+    ! retrieve item
+    p => list%GetItem(1)
+    call check(error, associated(p, n))
+    select type (item => p)
```
langevin-usgs, line 57: I'm not sure i've seen this syntax before.  Item is not declared, but yet it is assigned the correct type.

langevin-usgs (approved):

PR #1548, opened 2024-01-10
fix(idm): reset auxvar 2d arrray in array input loader each period

mjreno:
The Modflow6 input/output guide describes the input aux parameter for EVT and RCH Array-Based Input, which includes the line:

```
If an array is not specified for an auxiliary variable, then a value of zero is assigned.
```

When the input specifies a TAS for the aux parameter it is already reset because the TAS manager is reset.  However, the parameter is not currently explicitly reset to zero when not associated with a TAS.

### DIFF
--- a/src/Utilities/Idm/mf6blockfile/StressGridInput.f90
+++ b/src/Utilities/Idm/mf6blockfile/StressGridInput.f90
@@ -249,18 +249,17 @@ contains
                                   this%mf6_input%mempath, this%sourcename)
     end if
     !
-    ! -- reset input context memory for parameters
     do n = 1, this%nparam
       if (this%param_reads(n)%invar /= 0) then
         !
-        ! -- set definition
-        idt => this%mf6_input%param_dfns(this%idt_idxs(n))
-        !
         ! -- reset read state
         this%param_reads(n)%invar = 0
-
+        !
       end if
     end do
+    !
+    ! -- reset auxvar array each period
+    this%bndctx%auxvar = DZERO
     !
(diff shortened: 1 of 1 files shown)


src/Utilities/Idm/mf6blockfile/StressGridInput.f90
```diff
@@ -249,19 +248,18 @@ subroutine ingrid_reset(this)
...
         !
         ! -- reset read state
         this%param_reads(n)%invar = 0
-
+        !
       end if
     end do
     !
+    ! -- explicitly reset auxvar array each period
+    this%bndctx%auxvar = DZERO
```
langevin-usgs, line 261: We should probably loop through each array element and set to zero, rather than using a single expression.  Maybe it doesn't matter with latest compilers, but these types of assignments have given us stack issues in the past.

langevin-usgs (commented):
This fix seems worthy of a a bullet in the release notes as it may affect existing models and behavior.

mjreno:
@langevin-usgs This fixes an issue in the develop branch but it has not been exposed through an official release- should release notes still be updated?

langevin-usgs:
> @langevin-usgs This fixes an issue in the develop branch but it has not been exposed through an official release- should release notes still be updated?

Yes, we should be updating the release notes as the changes get merged into develop.  The release notes should document changes in develop that will affect users.  In the release notes folder, there is a develop.tex file where descriptions can go.

langevin-usgs (approved):

PR #1561, opened 2024-01-19
fix(par): bug with print_input in exchange

mjr-deltares:
and
- add test for print_input in exchange
- add test for parallel disu coupling


mjr-deltares:
@emorway-usgs : we should probably update your GWE feature branch with these changes

jdhughes-usgs (approved):

PR #1569, opened 2024-01-23
feat(input): add optional mfsim.par input file

mjreno:
Co-authored with @mjr-deltares 

Assign models to partitions with an optional mfsim.par input file:
- Input file defines a partitions block that assigns models (in mfsim.nam model block order) to provided partitions
- Still needs:
  - update docs 
  - option (log_mpi) support
  - validation of provided partition assignments
  -  FloPy support
  - autotest integration


mjr-deltares:
This will appear merged with other work under a new PR

PR #1573, opened 2024-01-24
fix(dist): multiple fixes following review

wpbonelli:
* omit `utils/idmloader/` subdir
* expand distribution checks per above
* rerun `mf6ivar.py` to update mf6io files
* fix typo in `gwf-maw.dfn` (missing "and")
* use latex-friendly quotes in `gwf-rcha.dfn`
* thanks to @rbwinst-usgs for raising issues


langevin-usgs (approved):
Looks good.  Only possible issue is the non-ASCII double quotes mentioned offline.

PR #1576, opened 2024-01-26
fix(chd): fix CHD rate and budget calculation for edge case with XT3D and exchanges

mjr-deltares:


mjr-deltares:
Hi @langevin-usgs , this would be one way to solve that issue with the budget calculation. Let me know what you think.

langevin-usgs (approved):
Tests are all passing, so it looks like the chd calculation can be done this way.  Will need to keep in mind that chd%simvals are not available until after chd_bd.  

We should probably put a short description of this in the release notes (develop.tex) as the fix is beneficial for users.

mjr-deltares:
I will put in a description of this issue in the release notes.

PR #1577, opened 2024-01-26
fix(idm): loading of exchanges with remote models (parallel case)

mjr-deltares:
when multiple exchanges are loaded that connect to the same remote model, NCELLDIM was deallocated for the first and then allocated for the next. mem_deallocate does not remove the entry from the memory list though. Setting a pointer at a later stage then always points to the first entry (which has been set to null at deallocation) simply because it is first in the list. 

This relates to the general idea that our memory manager is not set up for managing 'temporary' items like NCELLDIM. Something that we should probably discuss in more detail.


mjr-deltares:
solution pair-programmed with @mjreno

PR #1581, opened 2024-02-01
feat(idm): add subpackage input load support

mjreno:
Add a subpackage (already idm integrated) to dfn: `# mf6 subpackage <subp>`
- where `<subp>` is, for example, `utl-tvk`
- Add another subpackage to the same dfn by adding a new line: `# mf6 subpackage <subp2>`
- Any number of subpackage types ok but currently restricted to 1 instance each
-  A new internal tag related to the input filename tag contains the mempath (e.g. when `TVK6_FILENAME` is defined also expect `TVK6_MEMPATH` to be defined)

Refactor ascii input loaders:
- add timeseries to input block definition
- encapsulate ascii static load in LoadMf6FileType
- dynamic loader now also controls static load
- add ascii init interface for dynamic loader types
- split ascii dynamic list loader implementation into base and derived types

General:
- use defined constants for string length in input definition types
- remove package load interface
- loader variable renaming for consistency
- rename list and array dynamic loader modules and files


mjreno:
Closing temporarily to focus on netcdf PR.

PR #1582, opened 2024-02-02
fix(par): first version of message cache for MPI traffic

mjr-deltares:
This should deal with memory issue as reported by LSU for large models in parallel runs.


mjr-deltares:
NB: GWT parallel is not working yet because the cache needs to be initialized for every solution separately

PR #1586, opened 2024-02-05
chore(dag_module.f90): cleanup docstrings in DAG library

emorway-usgs:


jdhughes-usgs:
@emorway-usgs I am not sure we should modify the code in `Utilities\Libraries\` since this is code developed by others which is covered by a separate license.

emorway-usgs:
@jdhughes-usgs, @langevin-usgs, Feel free to shelve this.  I'm wondering if at the very least the source code should include a comment at the top identifying where the code came from and possibly even the license under which it is being used?

PR #1597, opened 2024-02-08
refactor: matching dfn and module filenames

wpbonelli:
* drop numbers from filenames so DFNs and corresponding modules match
* changes apply to files in
  * `src/Model/GroundWaterFlow`
  * `src/Model/GroundWaterTransport`
  * `src/Model/TransportModel`
  * `src/Solution/LinearMethods`
* first step toward #1532
* makes explicitly mapping DFN to paired IDM file unnecessary


mjreno (approved):
Looks good to me @wpbonelli

mjreno:
@wpbonelli did you consider renaming generated idm model/sim namefile and exchange fortran definition files so they are also consistent with the corresponding dfn files.

wpbonelli:
Thanks @mjreno for the catch

PR #1598, opened 2024-02-08
fix(mvr): improve water mover performance

langevin-usgs:
* A changed introduced in #1427 resulted in a performance decrease for GWF models with a large number of water movers.  This PR reduces the number of lookups by resetting mapped ids only when the set of movers has changed.
* This PR also resets the release notes for the next release.


mjr-deltares (approved):
thanks @langevin-usgs this looks good

mjr-deltares:
There is a conflict, would have merged otherwise.

NB the performance loss is based on our lookup operation into the memory list. I have a parallel PR #1601  that we should look at together, that could improve the overall performance of the program, for example, for a simulation with a truly large number of (interface) models

PR #1599, opened 2024-02-08
feat(idm): update TVK, TVS and STO packages for IDM

mjreno:
- Dependent on PR #1581
  - will need to merge files including dfn2f90.py, meson/msvs build files and gwf3npf8.f90

- STO period block definition updated to recarray with non-advanced setting type
- Updates show how model code will integrate with input side processing of non-advanced setting (keystring) types


mjreno:
Closing temporarily to focus on netcdf PR.

PR #1600, opened 2024-02-08
chore(timer.f90): cleanup docstrings in Timer utility

emorway-usgs:


src/Utilities/Timer.f90
```diff
@@ -12,183 +13,171 @@ module TimerModule
...
+    real(DP) :: dt
     !
-    IF (IT == 0) THEN
-      call CPU_TIME(t1)
-    ELSE
-      call CPU_TIME(dt)
+    if (it == 0) then
+      call cpu_time(t1)
+    else
+      call cpu_time(dt)
```
jdhughes-usgs, line 175: @langevin-usgs what should the convention be for fortran intrinsics? It might be good use CAPS like we use for parameters (DZERO, etc.).

emorway-usgs: I deliberated on this one, so happy to restore.  See [635cdb7](https://github.com/MODFLOW-USGS/modflow6/pull/1600/commits/635cdb7db296cab37055bd614ec3b9fad0e16652)

jdhughes-usgs:
@emorway-usgs it was a question. Maybe @langevin-usgs has a different opinion.

jdhughes-usgs (approved):

src/Utilities/Timer.f90
langevin-usgs: I don't have a preference.  Will go with whatever is in here now.  Ultimately hoping for a linter to handle this.

langevin-usgs (approved):

PR #1601, opened 2024-02-09
perf(mem): add faster lookup into memory list

mjr-deltares:
Add faster lookup into memory list to improve performance of get_from_memorylist. We use this everywhere and particularly for large numbers of models and exchanges, the simulation has been reported to slow down.

### DIFF
--- a/src/Model/Connection/GwfInterfaceModel.f90
+++ b/src/Model/Connection/GwfInterfaceModel.f90
@@ -69,8 +69,8 @@ contains
     end if
 
     ! create discretization and packages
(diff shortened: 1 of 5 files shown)


mjr-deltares:
Hey @mjreno, I have made some changes to the get_from_memorylist routine to speed things up. Maybe we can discuss offline? I had to modify some IDM related stuff because the old routine had some implicit behavior that has been changed.

mjr-deltares:
Taking this back for now, this approach will not provide the anticipated performance improvement...

PR #1602, opened 2024-02-09
ci: upload test artifacts on failure

wpbonelli:
* if CI tests fail it can be helpful to inspect their outputs
* add `workflow_dispatch` trigger to `large.yml`
* miscellaneous cleanup


langevin-usgs (approved):
great idea

PR #1603, opened 2024-02-09
refactor(dfn2f90): use dfns.txt input file, add options

wpbonelli:
* split off from #1532
* another step toward #1491
* introduce `dfns.txt` specifying DFN files to pair IDM files with, also support YAML file input
* since #1607 paired IDM files are named `<dfn>idm.f90` so no need for explicit mapping
* introduce options `--dfn`, `--outdir`, `--verbose` for `dfn2f90.py`
* co-locate generated IDM files in `src/Idm/` subdirectory
* update `utils/idmloader/IDM.md` per above
* add `pyaml` to `environment.yml`


utils/idmloader/dfns.txt
```diff
@@ -0,0 +1,59 @@
+# Add a new DFN parameter set to MF6
+# by adding a new entry to this list
+# ----------------------------------
+
+# simulation
+sim-nam.dfn
+
```
mjreno, line 7: sim-tdis.dfn should be here?  Also, it is an outlier with respect to the naming pattern- should probably be sim-tdisidm.f90 if possible.

mjreno (approved):
Looks good @wpbonelli

langevin-usgs (approved):

PR #1607, opened 2024-02-10
refactor: matching dfn and module filenames

wpbonelli:
* reopen #1597 
* first step toward #1491
* drop numbers from filenames and update naming as necessary so DFNs and corresponding module files match &mdash; paired component files are named `<dfn name>idm.f90`
* makes explicitly mapping DFN to paired IDM file unnecessary: #1603
* applies to files in
  * `src/Exchange`
  * `src/Model/GroundWaterFlow`
  * `src/Model/GroundWaterTransport`
  * `src/Model/GroundWaterEnergy`
  * `src/Model/SurfaceWaterFlow`
  * `src/Model/TransportModel`
  * `src/Solution/LinearMethods`


autotest/test_swf_dfw_gwf.py
```diff
@@ -58,20 +58,20 @@ def build_models(idx, test):
     add_swf_model(sim)
     add_gwf_model(sim)
 
-    swfgwf_data = [
+    gwfswf_data = [
```
langevin-usgs, line 61: Is there are a reason for changing this?  My thinking is that the name should be swf-gwf to loosely support the idea of upper to lower.  This order also corresponds to the cellids in the swfgwf_data list.  The first tuple there is a SWF reach number and the second is the GWF cellid.

src/Model/GroundWaterFlow/gwf.f90
```diff
@@ -1175,7 +1175,7 @@ end subroutine gwf_da
   !> @brief GroundWater Flow Model Budget Entry
   !!
   !! This subroutine adds a budget entry to the flow budget.  It was added as
-  !! a method for the gwf3 model object so that the exchange object could add its
+  !! a method for the gwf- model object so that the exchange object could add its
```
langevin-usgs, line 1178:
```suggestion
  !! a method for the gwf model object so that the exchange object could add its
```

langevin-usgs (commented):
Hey @wpbonelli, this is great.  I think we should keep the swfgwf name, but let me know if there is reason to change it.

autotest/test_swf_dfw_gwf.py
wpbonelli: Understood. I had two reasons for doing this: consistency with GWT, GWE (and PRT), and it makes [the logic in `src/meson.build`](https://github.com/MODFLOW-USGS/modflow6/pull/1532/files#diff-f28598af2e23aa5d2bc7c72e022ae2c56a33802eb970afffaeca1e40607f97feR27) to programmatically define exchanges slightly simpler if we can assume the order &mdash; but can just check the `src/Exchange` directory instead to see which exchanges are present. I will restore the original swf-gwf namings.

langevin-usgs (approved):

PR #1612, opened 2024-02-12
test: accommodate single-precision comparison programs

wpbonelli:
Double-precision versions of mf2005, mfnwt, mfusg and mflgr have been removed from [version 15 of the executables distribution](https://github.com/MODFLOW-USGS/executables/releases/tag/15.0). I think some autotests either need modification to work with the single-precision programs, or the test framework needs to build double-precision versions with pymake before testing.

Anticipating the former option this PR
* drops double-precision programs from test targets dict
* updates test_gwt_mt3dms_p01 to use single prec mf2005
* could relax comparisons in failing tests, if that is an appropriate solution

...but the above can be walked back if building and testing with double-precision versions is preferred.

Tests are expected to fail here in the meantime.

PR #1614, opened 2024-02-12
test: use v14 of executables distribution

wpbonelli:
* relevant to https://github.com/MODFLOW-USGS/modflow6/pull/1612
* v15 of executables distribution doesn't have double-precision program versions
* temp workaround until mf2005/mfnwt/etc comparisons replaced with mf6 regression


langevin-usgs (approved):
Good approach for now.  I think we are planning to include the double precision versions in next release of the executables.

PR #1620, opened 2024-02-14
docs(release): improve release procedure documentation

wpbonelli:
* correct inaccuracies from earlier versions of the release automation
* describe release notes maintenance procedure
* rename `vx.x.x-template.tex` -> `vx.y.z-template.tex`
* remove unused function from `build_dist.py`
* reformat python scripts in `distribution/`
* remove water mover performance entry from `develop.tex`


langevin-usgs (commented):
very thorough

PR #1633, opened 2024-02-18
fix(sfr): resolve divide by zero issue

### LINKED ISSUE #1585: Sfr fails to deal with downstream inactive reaches correctly.
**Describe the bug**
If a reach is connected to one or more downstream reaches and all the downstream reaches are inactive, MODFLOW terminates with a divide-by-zero error.

**To reproduce**
Run the attached model. It runs correctly for stress period 1 but crashes in stress period 2 with a downstsream reach becomes inactive.

**Expected behavior**
Either of the following
1. MODFLOW terminates with an error message.
2. MODFLOW treats the reach has having no downstream reaches to which flow should be routed.

**Screenshots**
image

**Environment**
 - Operating system (e.g. macOS, Linux, Windows) and version
 - MODFLOW 6 version (if installed via distribution)
 - Compiler toolchain/version (if built from source)
[SfrBug.zip](https://github.com/MODFLOW-USGS/modflow6/files/14170839/SfrBug.zip)

jdhughes-usgs:
* add test that replicates test model provided in issue #1585

closes #1585

### DIFF
--- a/autotest/test_gwf_sfr_inactive.py
+++ b/autotest/test_gwf_sfr_inactive.py
@@ -1,0 +1,247 @@
+# test based on issue 1585 fix
+import flopy
+import numpy as np
+import pytest
+
+from framework import TestFramework
+
+paktest = "sfr"
+cases = ["sfr-1585"]
+
+
+def build_models(idx, test):
+    # static model data
+    # temporal discretization
+    nper = 2
+    tdis_rc = []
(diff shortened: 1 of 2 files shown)


langevin-usgs (approved):

langevin-usgs (commented):
Would be good  to update release notes

PR #1636, opened 2024-02-20
docs: update release notes

jdhughes-usgs:
* add description of the bug fixed by PR #1633
* extend test_gwf_sfr_inactive autotest to reactive sfr reaches after inactivating 3 of 6 reaches

### DIFF
--- a/autotest/test_gwf_sfr_inactive.py
+++ b/autotest/test_gwf_sfr_inactive.py
@@ -1,4 +1,13 @@
-# test based on issue 1585 fix
+# Test based on issue 1585
+# The test has 6 sfr reaches that are not connected to the 
+# groundwater model.
+# 1. initially all reaches are active
+# 2. in the 2nd stress period the last three are inactivated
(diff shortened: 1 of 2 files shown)


autotest/test_gwf_sfr_inactive.py
```diff
@@ -1,4 +1,13 @@
-# test based on issue 1585 fix
+# Test based on issue 1585
```
langevin-usgs, line 1: We've switched to triple quotes for a test description header, such as:

```
"""
This tests reuses the simulation data in test_gwf_ifmod_idomain.py
and runs it in parallel on three cpus with

cpu 0: 'refmodel'
cpu 1: 'leftmodel'
cpu 2: 'rightmodel'

so we can compare the parallel coupling of 'leftmodel' + 'rightmodel'
with a serial 'refmodel'
"""
```

langevin-usgs (commented):
Looks good.

PR #1637, opened 2024-02-20
refactor(discretization): move dis modules to src/Model/Discretization

wpbonelli:
* move `src/Model/GroundWaterFlow/gwf-dis.f90` -> `src/Model/Discretization/Dis.f90`, same for disv/disu
* define "found type" in discretization modules, remove generated IDM file dependency
* move toward clean separation between generic framework and model-scoped code
* future work may introduce model-specific dis types extending these per recent discussion with @langevin-usgs @jdhughes-usgs @mjreno @aprovost-usgs, for now this allows moving forward with #1532 
* another step toward #1491


mjreno (approved):
Looks good @wpbonelli

Issue #1348, opened 2023-09-13
Self-contained example notebooks


wpbonelli:
### Background
The [MODFLOW 6 example problems](https://modflow6-examples.readthedocs.io/en/latest/) are an entry point to MODFLOW and a common first step is likely to locally reproduce example problems, but the notebooks assumes the `common/`](https://github.com/MODFLOW-USGS/modflow6-examples/tree/master/common) and `data` directories are present in the right relative locations, used from notebooks via `sys.path` manipulation. So it is either necessary to clone the repo then running the scripts or notebooks from the `scripts/` or `notebooks/` folders, respectively, manually download the implicit dependencies, or alter the examples to remove them.

### Proposal

The example notebooks could be downloadable from the docs, and could run with no external dependencies besides a Python/flopy environment and the MODFLOW binaries.

- [x] update [RTD site](https://modflow6-examples.readthedocs.io/en/latest/) site: render thumbnail gallery of downloadable notebooks on ReadTheDocs [like flopy](https://github.com/modflowpy/flopy/pull/1794), fix order of example descriptions and notebooks to match each other
- [x] [remove `.ipynb` files from version control](https://github.com/MODFLOW-USGS/modflow6-examples/pull/121) or [stripping output/metadata before committing](https://github.com/EC-USGS/pywatershed/blob/develop/DEVELOPER.md#committing-jupyter-notebooks)
  - https://github.com/MODFLOW-USGS/modflow6-examples/pull/128 
- [x] move general-purpose utilities from the mf6 examples repo's `common/` to [modflow-devtools](https://github.com/MODFLOW-USGS/modflow-devtools) &mdash; if a modeling-related utility is only used from a single notebook, moving it into that notebook &mdash; if a modeling-related utility is used from multiple notebooks, or may be more broadly useful, moving it to flopy &mdash; make mf6 notebooks self-contained and immediately runnable, without cloning the repo, copying files, etc
  - https://github.com/MODFLOW-USGS/modflow-devtools/pull/132
  - https://github.com/MODFLOW-USGS/modflow6-examples/pull/137

Related: [make flopy notebooks self-contained](https://github.com/modflowpy/flopy/issues/1872)

wpbonelli:
if https://github.com/MODFLOW-USGS/modflow-devtools/issues/134 materializes example scripts can be updated to use it, but otherwise this is done &mdash; the examples now only require modflow programs and a suitable python environment, no implicit filesystem dependencies

PR #1645, opened 2024-02-23
docs: fix typo in supplemental info

### LINKED ISSUE #1644: Typographical error in supplemental information
**Describe the bug**
In the supplemental information, Figure 3-2A has a typographical error. On the upper right side of the figure "ZDRN + DDN" should be "ZDRN + DDRN".

wpbonelli:
* fix 'ZDRN + DDN' -> 'ZDRN + DDRN'
* same fix in `DRN-Discharge-Scaling.ipynb`
* close #1644

The notebooks in `doc/SuppTechInfo/python` escaped my notice in #1348 &mdash; should they be included in the docs site @langevin-usgs @jdhughes-usgs?


langevin-usgs (approved):
The notebooks there are just there for us developers, in case we need to make a change to one of those figures in the supp tech guide.  I'm not sure we need to spend any time on them.

PR #1649, opened 2024-02-26
refactor(ts): skip search for repeated boundary time series file

ScottBoyce:
This should be done now. I ended up keeping the search loop and instead skip it when the time series file is for a boundary condition. My reasoning is that I spent all my time evaluating its application in the CHD and GHB packages and not their auxiliary variables. Just to be safe, I will let it do that for the aux time series files but skip for the boundary time series files.

I ran through each of Richard Winston's examples and get the same answer with a substantial speedup. In particular, the million time series file example reduced the runtime from 19 hours to 3 minutes!

commit message:

The subroutine read_value_or_time_series checks to see if a time series file (TS) has been previously included in the time series manager (tsManager) before adding it.

The check compares against the package name and the input parser row and column (ii, jj) location in the input file. If the package name and location match, then the TS is not added to the tsManager.

This check is never true for boundary (BND) type TS, so a conditional is added to only perform the search for for non-BND TS types.

This results in a substantial performance increase when there are a large number of time series files.

### DIFF
--- a/src/Utilities/TimeSeries/TimeSeriesManager.f90
+++ b/src/Utilities/TimeSeries/TimeSeriesManager.f90
@@ -580,22 +580,25 @@ contains
         r = timeseries%GetValue(totimsav, totim, &
                                 tsManager%extendTsToEndOfSimulation)
         bndElem = r
-        ! Look to see if this array element already has a time series
-        ! linked to it.  If not, make a link to it.
-        nlinks = tsManager%CountLinks(auxOrBnd)
         found = .false.
-        searchlinks: do i = 1, nlinks
-          tslTemp => tsManager%GetLink(auxOrBnd, i)
-          if (tslTemp%PackageName == pkgName) then
-            ! -- Check ii, jj against iRow, jCol stored in link
-            if (tslTemp%IRow == ii .and. tslTemp%JCol == jj) then
-              ! -- This array element is already linked to a time series.
-              tsLink => tslTemp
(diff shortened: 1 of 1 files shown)


langevin-usgs:
I looked into this more and think the check for AUX time series can also be removed.  There are two different ways a time series link can be created.  The first is from the routine modified in this PR (`read_value_or_time_series`).  This routine is used by the standard boundary packages (wel, drn, riv, ...) to establish a link between the time series specified in the time series file and the array inside the model package.  For the standard boundary packages, all of the links are cleared in the _RP routine before the links are created, so there is no need for a check.

There is also a separate routine called `read_value_or_time_series_adv` for advanced packages.  This routine also has this same check, and it appears it is necessary in this case.  The reason is that with the SETTINGS approach in the advanced package, a user could specify the same value (and thus the same link) multiple times.  See the example below.  Right now, this input is allowed.  For the first entry, a time series link would be established in the conc auxiliary variable for the first position in the auxiliary array.  And then upon reading the second entry, the previous link would be found and removed, and a new link would be created for the first position in the array.

```
BEGIN period  1
  1  AUXILIARY  conc  conc_ts1
  1  AUXILIARY  conc  conc_ts2
END period  1
```

So based on this, it seems like the search and check for an existing link in `read_value_or_time_series` is no longer needed, and is likely the remnant from earlier versions of the time series implementation.

I updated the PR

PR #1650, opened 2024-02-26
refactor: move hyeff function to separate module

wpbonelli:
* move function for effective horizontal hydraulic conductivity out of `gwf-npf.f90` so gwt-dsp and gwe-cnd don't depend on gwf-npf &mdash; toward eliminating dependencies between models
* introduce `HGeoUtilModule` for standalone hydrogeologic functions
* split out from #1532
* part of #1491

### DIFF
--- a/src/Utilities/HGeoUtil.f90
+++ b/src/Utilities/HGeoUtil.f90
@@ -1,0 +1,110 @@
+!> @brief General-purpose hydrogeologic functions.
+module HGeoUtilModule
+  use KindModule, only: DP, I4B
+  use ConstantsModule, only: DZERO, DONE
+
+  implicit none
+  private
+  public :: hyeff
+
+contains
+
+  !> @brief Calculate the effective horizontal hydraulic conductivity from an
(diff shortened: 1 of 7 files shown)


src/Utilities/HGeoUtil.f90
```diff
@@ -0,0 +1,110 @@
...
+  implicit none
+  private
+  public :: hyeff
+
+contains
+
+  !> @brief Calculate the effective horizontal hydraulic conductivity from an
+  !! ellipse using a specified direction (unit vector vg1, vg2, vg3)
+  !!
+  !! k11 is the hydraulic conductivity of the major ellipse axis
```
langevin-usgs, line 15: We should talk to @aprovost-usgs and come up with more general names for this routine and the argument variables.  These are generic ellipsoid properties that apply to hydraulic conductivity or solute dispersion or energy dispersion.  Good idea to move the routine out of NPF as it doesn't belong there any more.  We can do the renaming later in order to keep this moving.

aprovost-usgs: Good idea, I'll think about that

PR #1652, opened 2024-02-29
refactor: introduce model and exchange factories

wpbonelli:
* factor `Model/ExchangeFactory` modules out of `SimulationCreate` module
* add noop virtual PRT and SWF models and exchanges &mdash; allows cleaner templates, and virtual models/exchanges will be added anyway when SWF and PRT support parallel
* prep for templating for build-time model selection (see comments `TODO BEGIN/END TEMPLATING`)
* rename exchange create routines for templating
* move `typename` and `filename` up to `BaseExchangeType`
* deduplicate model registration logic, this...
  ```fortran
    model => null() ! can be null for remote models
    if (model_ranks(n) == proc_id) then
      im = im + 1
      write (iout, '(4x,2a,i0,a)') 'GWF6', ' model ', &
        n, ' will be created'
      call gwf_cr(fname, n, model_names(n))
      model => GetNumericalModelFromList(basemodellist, im)
      model_loc_idx(n) = im
    end if
    call add_virtual_gwf_model(n, model_names(n), model)
  ``` 
  ...was previously repeated for each model type, factor out a shared routine accepting proc pointers
* consolidate add/create routines in virtual model types 
* carved out of #1532
* step toward #1491
* after this, we can introduce [fypp](https://github.com/aradi/fypp), add templates for `Model/Exchange/ConnectionFactory`, and update meson files to generate source from templates


wpbonelli:
closing after offline chat with @langevin-usgs @jdhughes-usgs @mjr-deltares, we don't need to pursue templating at this time

PR #1655, opened 2024-03-04
refactor(npf): remove mfnwt and mfusg development conductance options

jdhughes-usgs:
* remove DEV_MODFLOWNWT_UPSTREAM_WEIGHTING option


jdhughes-usgs:
> I think I got to the bottom of the issue regarding why we could not replace the quadratic saturation call (in hcond) with satn and satm. Looks like we are recalculating conductance as part of CQ in the qcalc subroutine, so the quadratic saturation call will give a different answer with the latest head iterates than the satn/satm values calculated before the solve. Seems like we may have a small conductance inconsistency in CQ compared to CF/FC.

So I guess we leave as is? Or we could have an optional boolean indicating if the sat should be recalulated that would be false by default and True in call from CQ.

langevin-usgs:
> > I think I got to the bottom of the issue regarding why we could not replace the quadratic saturation call (in hcond) with satn and satm. Looks like we are recalculating conductance as part of CQ in the qcalc subroutine, so the quadratic saturation call will give a different answer with the latest head iterates than the satn/satm values calculated before the solve. Seems like we may have a small conductance inconsistency in CQ compared to CF/FC.
> 
> So I guess we leave as is? Or we could have an optional boolean indicating if the sat should be recalulated that would be false by default and True in call from CQ. Leaving as is might be the best so we can wait to see how things unroll with Richards.

I think we should replace that quadratic saturation call with satn and satm, and then modify Eric's problem to pass.  I think it's related to a bigger question.  If we are solving Ax=b as part of our picard loop, do we recalculate A with the converged head solution prior to doing the budget?  My feeling is that we shouldn't (we don't for the non-newton case).  I ran a test where sat was recalculated prior to the qcalc calls (essentially recalculating A with the latest heads), and there were a bunch of test failures.

src/Model/ModelUtilities/GwfConductanceUtils.f90
```diff
@@ -0,0 +1,402 @@
...
+      !
+      ! -- if both cells are non-convertible then use condsat
+    elseif (ictn == 0 .and. ictm == 0) then
+      condnm = condsat
+      !
+      ! -- At least one of the cells is convertible and using the
+      !    newton-raphson conductance formulation
+    else if (inwtup == 1) then
+      if (hn > hm) then
+        condnm = satn
```
langevin-usgs, line 82:
```suggestion
        sat_up = satn
```

src/Model/ModelUtilities/GwfConductanceUtils.f90
```diff
@@ -0,0 +1,402 @@
...
+      else
+        condnm = satm
```
langevin-usgs, line 84:
```suggestion
        sat_up = satm
```

src/Model/ModelUtilities/GwfConductanceUtils.f90
```diff
@@ -0,0 +1,402 @@
...
+      end if
+      !
+      ! -- multiply condsat by condnm factor
+      condnm = condnm * condsat
```
langevin-usgs, line 88:
```suggestion
      condnm = sat_up * condsat
```

langevin-usgs (approved):

PR #1657, opened 2024-03-05
fix(buy): coupled GWF models with BUY require interface model

mjr-deltares:
We do not have BUY supported in the classic GwfGwfExchange formulation, an interface model based approach is required.


mjr-deltares:
reviewed by @langevin-usgs

PR #1663, opened 2024-03-07
test(gwe): New autotest of gwe-gwe exchange

emorway-usgs:
- Redo of #1659 (the ci kept 'hanging' despite passing locally)
- Further tests gwe-gwe exchanges against Carslaw & Jaeger analytical solution, but also facilitates additional parallel testing of GWE (that autotest will be possible once this autotest passes on CI.


emorway-usgs:
No need for this after discovering that `plt.show()` was the culprit in #1659

PR #1672, opened 2024-03-14
ci(win/par): add test workflow for parallel mf6 on Windows

wpbonelli:
CI test mf6 parallel on Windows. Step toward a Windows parallel distribution.

* use [cygwin approach](https://petsc.org/main/install/windows/#cygwin-gnu-compilers-on-microsoft-windows) to build petsc with `ifort` and `cl`
* factor out actions for parallel testing on windows/non-windows
* factor out scripts for windows steps needing special treatment
* cache oneAPI and PETSc, installing/building is slow

Parallel autotests pass, but there are a couple remaining (but non-blocking) issues

* PETSc `make check` reports some errors, but they don't seem to invalidate the build
* `meson test` fails intermittently on "Parallel simulation test - 2 cores" on Windows, for now it's skipped

@mjr-deltares should these be solved before merging or should we come back to them?

### DIFF
--- a/.github/actions/test-par-win/action.yml
+++ b/.github/actions/test-par-win/action.yml
@@ -71,6 +71,15 @@ runs:
         path: petsc
         key: petsc-${{ runner.os }}-${{ steps.get-date.outputs.date }}
 
+    - name: Configure environment
+      shell: bash
+      run: |
+        PETSC_DIR=$GITHUB_WORKSPACE/petsc
+        PETSC_ARCH=arch-mswin-c-opt
+        echo "PETSC_DIR=$PETSC_DIR" >> $GITHUB_ENV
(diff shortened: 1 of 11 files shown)


mjr-deltares:
Nice progress @wpbonelli ! I was wondering, maybe we should do the meson tests like this:

`test(..., is_parallel : false)`

By default it runs them in parallel and the errors seemed to be file system related on Windows. (https://mesonbuild.com/Unit-tests.html#parallelism)

mjr-deltares (approved):

mjr-deltares:
For completeness (and possible spectators), you mentioned:

"PETSc make check reports some errors, but they don't seem to invalidate the build"

We agreed to keep moving forward and study this as a separate issue. The PR can go in as is.

wpbonelli:
also for posterity, doing meson tests in serial solves the intermittent failure, thanks @mjr-deltares !

PR #1673, opened 2024-03-14
feat(sfw): implement conveyance concept

langevin-usgs:
* Modify the mannings equation to use a conveyance term, which is calculated from the cross section package
* Add a test-drive test for the swf cross section utilities


autotest/test_swf_dfw_bowl.py
```diff
@@ -243,14 +243,23 @@ def check_output(idx, test):
...
             2.00975535,
             1.00975489,
             0.00928387,
         ]
 
-    assert np.allclose(stage_all[-1].flatten(), stage_answer, atol=1.0e-5)
+    # print out the answer in a form that can be 
+    # dropped into this script, if necessary, as the answer
+    for v in stage_all[-1].flatten():
+        print(f"{2*"    "}{v:.8f},")
```
mwtoews, line 256: This raises: `SyntaxError: f-string: expecting '}'`. How did this get past CI?

wpbonelli: The CI test env currently defaults to python 3.12, which [relaxed some restrictions on f-strings](https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings). We might CI test the oldest Python we want to support, for now 3.9

PR #1678, opened 2024-03-17
fix(gwf-lak): fix and test lakes set to inactive

### LINKED ISSUE #1674: STATUS INACTIVE in LAK package doesn't seem to work
Hi,

I've slightly modified one of your test cases (ex-gwf-lak-p02) in order to test the LAK package:

[__test.zip](https://github.com/MODFLOW-USGS/modflow6/files/14612412/__test.zip)

Lak 1 is deactivated in stress period 2 using STATUS INACTIVE.
In my understanding this should turn it off entirely for that period.
But it still outputs budgets for that period both in ex-gwf-lak-p02.cbc (SAVE_FLOWS) and __TEST.LAK.CBC (BUDGET FILEOUT).
Modflow version is 6.4.4

Best regards
Andreas

langevin-usgs:
As mentioned in #1674 setting lake status to inactive does not seem to work properly.  The intent of this PR is to fix this issue and test to make sure that when a lake is inactive that all output is set as intended.

- [x] Add test
- [x] Ensure inactive lake does not affect groundwater flow
- [x] Ensure LAK-GWF flows are zero
- [x] Ensure lake stage is DHNOFLO (in both the binary lake stage file and the gwf list file) for an inactive lake
- [x] Ensure all lake budget terms are zero (including rainfail, evap, etc.) if a lake is inactive
- [x] Check SFR to see if it has a similar issue; fixed on #1729 
- [x] Close #1674

The stage table written to the listing file will now look like the following when the lake is inactive:

```
 LAK PACKAGE (LAK_0) STAGES FOR EACH CONTROL VOLUME   PERIOD      2   STEP        1
 -----------------------------------------------------------------------------------
 NAME                   NUMBER      STAGE     SURFACE AREA WETTED AREA     VOLUME   
 -----------------------------------------------------------------------------------
 LAKE1                    1       1.00000E+30  1.00000E+30  1.00000E+30  1.00000E+30
 -----------------------------------------------------------------------------------
```

Couple of things to note:
* The last stage value remains in memory when a lake is inactivated, and it seems like this in-memory stage is used as the initial stage when a lake is turned back on.  Not sure if we want to do anything about this, but there is some question as to what we should do with the lake storage term.
* The qleak variable can probably be eliminated as it is simply equal to -simvals.  Qleak wasn't being reset to zero when a lake was inactivated.
* It looks like we were still solving for lake stage even if a lake was inactive.  I added the logic to skip out of the solve loop when a lake was inactive.
* I removed the old `lak_cfupdate` remnant as it seems like it is time for this to go.
* SFR has been updated to make sure inactive reaches work properly.  We should also test MAW.
* We should add a test for these inactive features and make sure they work properly with transport.


jdhughes-usgs (approved):

langevin-usgs:
Now that #1729 has gone in, the format of the lake dependent variable table has changed to:

```
 LAK PACKAGE (LAK_0) STAGES FOR EACH CONTROL VOLUME   PERIOD      2   STEP        1
 -----------------------------------------------------------------------------------
 NAME                   NUMBER      STAGE     SURFACE AREA WETTED AREA     VOLUME   
 -----------------------------------------------------------------------------------
 LAKE1                    1        INACTIVE     INACTIVE     INACTIVE     INACTIVE  
 -----------------------------------------------------------------------------------
```

The budget table shows no flows for period 2 when the lake is inactive

```
 LAK_0 PACKAGE - SUMMARY OF FLOWS FOR EACH CONTROL VOLUME   PERIOD      2   STEP        1
 ---------------------------------------------------------------------------------------------------------------------------------------------------------
                                                                                                                                                PERCENT   
   NUMBER       GWF        RAINFALL   EVAPORATION     RUNOFF     EXT INFLOW   WITHDRAWAL  EXT OUTFLOW    STORAGE      CONSTANT     IN - OUT    DIFFERENCE 
 ---------------------------------------------------------------------------------------------------------------------------------------------------------
     1       0.0000       0.0000       0.0000       0.0000       0.0000       0.0000       0.0000       0.0000       0.0000       0.0000       0.0000     
 ---------------------------------------------------------------------------------------------------------------------------------------------------------
```

PR #1685, opened 2024-03-26
ci(parallel): don't use -n auto on windows to avoid oversubscription

wpbonelli:
* runner VM [has 4 cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories), most parallel tests are 2-way split, so use `-n 2`
* hoping to reduce parallel CI runtime


wpbonelli:
The test runtime hovers around 30min with `-n auto`, `-n 2` and in serial alike, so this probably isn't worth it. But it's definitely worth warming the oneapi cache nightly like we do elsewhere, which should hold runtime to ~40m total

PR #1687, opened 2024-03-27
ci(parallel): nightly cache warming for mf6 parallel oneapi components

wpbonelli:
* followup to #1685
* installs can take 30+ min and not handled by setup-fortran, so need to cache manually
* factor out an action for mf6 parallel oneapi install


langevin-usgs (approved):
The Windows-2022 parallel job took 1.5 hours to complete.  Presumably this will speedup once the compiler is cached?

wpbonelli:
Yes- once the oneapi components are cached by the scheduled job they are available to PR branches. previously each PR branch had to create its own cache

PR #1695, opened 2024-04-02
Move from mamba to pixi

Hofer-Julian:


Hofer-Julian:
Closing, since CI seems to be optimized for PRs from forks

PR #1696, opened 2024-04-02
Move from mamba to pixi

Hofer-Julian:
Pixi is an alternative to conda, made by https://prefix.dev which are also behind mamba.
We use it at Deltares for a bit over half a year now and are really satisfied with it. The parallel MODFLOW 6 demo uses it as well.
Full disclosure: I will stop working at Deltares in one month and start working for prefix starting from July.

The biggest difference between pixi and mamba, is the presence of a lock file and task support.
Lock files ensure reproducibility between CI and programmers. They also make it possible to check out an old commit with the correct dependencies.
Task support makes it much easier to reproduce the work happening in the CI.

This PR is as minimal as possible in order to make reviewing easier. It doesn't remove `environment.yml` yet and doesn't touch the release workflows.
Some workflows are not triggered by PRs so we should check if they actually work after merging.

Lock files also mean that you don't automatically get the newest dependencies.
Because of that, I've added a cron task which opens an update PR once per month.

Pixi is fully supported by JetBrains, VSCode support is [WIP](https://github.com/microsoft/vscode-python/pull/22968).
I found that everything works by just starting VSCode with `pixi run code .`

(Potential) follow-ups:
- Move release workflows
- Remove `environment.yml`
- Use petsc from conda forge for linux and macOS
- Add more pixi tasks


wpbonelli (approved):
To summarize some offline chatting with @Hofer-Julian

* We thought probably best to keep the conda environment and related portions of dev docs until contributors have had time to get used to pixi. This can be "opt-in" for now.
* Julian has now updated the release workflows. I will follow up on this PR with a few more tweaks, as for some reason I don't have permission to push directly to it (maybe because Julian's fork is from Martijn's, not the upstream?)

PR #1702, opened 2024-04-03
refactor(export_array): testing

langevin-usgs:
* add export array testing to autotest/test_gwf_npf01_75x75.py
* update tex files

There is still an unresolved issue in that export_array only works for "layered" variables, but if we add "layered true" to disv1d variables, then the io guide says "layered" for these variables.  Also, for the swf-dis2d grid, if we include support for export_array, then we get nrow files created, each with ncol values in it.  So there is some work left to do in supporting disv1d, dis2d, and other variables, such as gwf-dis-top, that might be considered "layered".


mjreno (approved):
Looks good @langevin-usgs

PR #1703, opened 2024-04-03
ci: various pixi refactoring

wpbonelli:
* followup to #1696
* add version to `pixi.toml`
* include `doc/Doxyfile` and `pixi.toml` in `update_version.py`
* consolidate tasks
* refactor custom install as pypi dependencies
* remove branch filters from push trigger in `commit.yml`
* update dev docs: keep conda language, tidying
* ignore pixi beta feature warning in ci
* restore vscode config to avoid breaking contributor workflow &mdash; when vscode python supports pixi I think we can update the tasks to use `"command": "${command:python.interpreterPath}"`


.vscode/build_vscode.py
```diff
@@ -21,8 +21,22 @@
 if args.action == "rebuild" and os.path.isdir(builddir):
     shutil.rmtree(builddir)
 
+if args.buildtype == "release":
```
Hofer-Julian, line 24: Was something broken here?
Or is there a different reason why you reverted that?

pixi.toml
```diff
@@ -70,3 +32,32 @@ requests = "*"
 scipy = "*"
 shapely = "*"
 xmipy = "*"
+
+[pypi-dependencies]
+flopy = { git = "https://github.com/modflowpy/flopy.git" }
+modflowapi = { git = "https://github.com/MODFLOW-USGS/modflowapi.git" }
+modflow-devtools = { git = "https://github.com/MODFLOW-USGS/modflow-devtools.git" }
```
Hofer-Julian, line 39: I already had that change locally.
There are two reasons why I didn't submit it in the end:
- This is a change in behavior, the git commit will be pinned in the lock file
- Our use of flopy is a bit special. We modify it with `update-flopy`. Let's keep that in mind in case problems with flopy appear.

Hofer-Julian (commented):
Looks like a nice cleanup to me. Added a few comments to keep in mind :)

pixi.toml
wpbonelli: > This is a change in behavior, the git commit will be pinned in the lock file

good point, I hadn't thought this through all the way &mdash; anyone developing flopy alongside mf6 will need to check in an updated pixi lock file to pull in any changes to flopy that are not generated components

.vscode/build_vscode.py
wpbonelli: this was to make adopting pixi optional for vscode users for now.

pixi.toml
langevin-usgs: Seems to me like we are going to want to develop using the most up to date version of flopy.

wpbonelli: Yes- I think there is still a case to keep the version locked if we want to keep going in the direction of deterministic CI? We can still develop against the latest flopy, it would just be explicit in mf6 version history when a change is made w.r.t. flopy compatibility. could make debugging easier at the cost of a busier commit log.

PR #1709, opened 2024-04-08
fix(prt): set tmax properly before particle solve

wpbonelli:
Fix two issues:
* maximum tracking time was set incorrectly, particles were solved beyond the end of the time step for multi-step simulations &mdash; max time within the tracking loop should be end of time step unless it's the last time step in the simulation, in which case it should be particle stop time
* if, on recording a particle datum, the tracking time fell exactly on the boundary between time steps, the datum was assigned to the subsequent time step and stress period, while MODPATH 7 assigns it the previous &mdash; factor out a routine in the base `MethodType` to save particle data, with some extra logic for an inclusive upper bound on the time step like MP7

### DIFF
--- a/src/Solution/ParticleTracker/Method.f90
+++ b/src/Solution/ParticleTracker/Method.f90
@@ -154,7 +154,7 @@ contains
 
   !> @brief Save a particle's current state.
   subroutine save(this, particle, reason)
-    use TdisModule, only: kper, kstp, nper, nstp, totimc, delt
+    use TdisModule, only: kper, kstp, totimc
     ! dummy
     class(MethodType), intent(inout) :: this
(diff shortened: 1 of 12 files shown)


src/Solution/ParticleTracker/Method.f90
```diff
@@ -151,6 +152,38 @@ subroutine pass(this, particle)
     call pstop(1, "pass must be overridden")
   end subroutine pass
 
+  !> @brief Save a particle's current state.
```
wpbonelli, line 155: @aprovost-usgs I factored out a `save()` routine here to handle the special treatment for period/timestep in case of times right on the boundary between time steps, per offline discussion

PR #1713, opened 2024-04-10
ci: set pixi-setup frozen to true

jdhughes-usgs:


wpbonelli:
I'm confused about pixi `--locked` vs `--frozen`. I thought `--frozen` would install exactly what's in the lockfile. But the lockfile has flopy [4c44cb0d9d6ef02c092938dad943cdb8285f82a7](https://github.com/MODFLOW-USGS/modflow6/blob/f5edfadf38d88869e2319d86a95d750c8f1e5333/pixi.lock#L233) and [here](https://github.com/MODFLOW-USGS/modflow6/actions/runs/8636233278/job/23676200439?pr=1713#step:6:88) we get flopy [9e87acddcf0740aac491826ef86fd45ae2304dac](https://github.com/modflowpy/flopy/commit/9e87acddcf0740aac491826ef86fd45ae2304dac), which is currently the tip of develop.

mjr-deltares:
@jdhughes-usgs parallel Ubuntu action is failing here, will look into that

PR #1719, opened 2024-04-11
fix(tsp-apt): pair the user-specified concentration for finf with rej. infiltration

emorway-usgs:
- When using the mover transport (MVT) package with UZF and UZT, rejected infiltration was paired with the calculated concentration of the UZF object rather than the user-specified concentration assigned to the infiltration.
- black
- fprettify
- Thanks @langevin-usgs for the elegant fortran solution that is free from "special case" handling!


src/Model/TransportModel/tsp-apt.f90
```diff
@@ -635,6 +635,19 @@ function apt_check_valid(this, itemno) result(ierr)
     end if
   end function apt_check_valid
 
+  !> @brief Advanced package transport utility function
+  !!
+  !! Set the concentration (or temperature) to be used by either MVT or MVE
+  !<
+  function get_mvtconc(this)
```
langevin-usgs, line 642: Do we want to call this get_mover_depvar instead (here and elsewhere in the changes) so that it works for both concentration and temperature?  Just a thought.

langevin-usgs (approved):
Glad this worked out!  Just one minor comment for your consideration.

src/Model/TransportModel/tsp-apt.f90
emorway-usgs: Good idea @langevin-usgs, updated function name with 8ef02be

langevin-usgs (approved):

PR #1722, opened 2024-04-11
fix(test): numpy resize not reliable in test_gwf_ats_lak01.py

langevin-usgs:
* test works in mf6 CI but not in flopy CI
* allocate budget array instead


wpbonelli:
Not that there is any issue with this solution, but I wonder if `np.resize()` would work &mdash; the latter copies the array, while `ndarray.resize` works inplace. I don't know why there is another ref to the underlying array in the flopy CI, but not here.

langevin-usgs:
I first went with the np.resize() function but then went with zeros and was able to simplify a bit.  Thanks for checking.

PR #1725, opened 2024-04-15
check(petsc): run all autotests with PETSc solver

mjr-deltares:


mjr-deltares:
easier to check locally

PR #1732, opened 2024-04-16
fix(idm): stress package loaders allocate duplicate iper/ionper variables

mjreno:
Issue introduced in PR #1610.
  - This PR moved iper/ionper to dynamic loader base class
  - There are 2 contexts (loaders) for stress data- an input style specific block loader (e.g. list, array) and the generic dynamic loader that manages the load and allocates the input style loader.  These were both allocating iper/ionper via the base class.
  - Fix is to move iper/ionper back into the higher level dynamic loader.


mjr-deltares:
Hi @Manangka this is related to the problem you find in your work with #1701

langevin-usgs (approved):

Issue #915, opened 2022-03-01
Does UZT route mass from GWSEEP?


RyanConway91:
Is it possible that GWSEEP sent to MOVER from UZF is not getting used in transport?
In my current model, UZFs are adding 0 mass to LAKs and SFRs in a transport simulation. However, I am using GWSEEP and looking at color flood of the simulated concentrations it seems like the UZF cells should be picking up a lot of mass from GWSEEP. 
If I print flows into the transport list file, I only see rejected infiltration mass printed. However, I know from the observation utility that the UZFs are routing a lot of flow to MVR.
 
![image](https://user-images.githubusercontent.com/25184697/156238511-79030cdb-1ac7-4f15-b3e6-f7d03b5cf9f4.png)

I am thinking (based on some previous issue discussion) that it might make sense to switch from UZF to drains and route the flow that way. Any obvious issues with that approach?

langevin-usgs:
I think this is one of the main reasons (better support for MVR and MVT) that we are recommending drains for this.

emorway-usgs:
Also, [this example](https://modflow6-examples.readthedocs.io/en/master/_notebooks/ex-gwf-drn-p01.html) employs both UZF and DRN, in case seeing a working example is helpful.  Note that it doesn't set up the associated transport, but that wouldn't be too difficult.

PR #1743, opened 2024-04-17
perf(UZF): add deprecation warning for SIMULATE_GWSEEP

emorway-usgs:
- Relates to #915 and #1719 
- It is not recommended to use `SIMULATE_GWSEEP` because of the potential for errors with `MVT` and `MVE`
- A better approach is to use the `AUXDEPTHNAME` option available within the `DRN` package.  An example showing how to use this option with a comparison to groundwater seepage calculated by the UZF package is provided in [MODFLOW 6 examples GitHub repository](https://github.com/MODFLOW-USGS/modflow6-examples).  The example is titled "[Drain Package Drainage Discharge Scaling Option Problem 1](https://modflow6-examples.readthedocs.io/en/latest/_examples/ex-gwf-drn-p01.html)" 
- As of right now, only a deprecation warning is introduced.  The functionality still remains for the time being with a warning message to the user.
- fprettify


langevin-usgs (approved):
nice

PR #1540, opened 2024-01-05
feat: filter components in mf6ivar and dist scripts

wpbonelli:
* final piece of #1491
* configurable models/exchanges in release workflow `release.yml`
* configurable models/exchanges in distribution scripts and `mf6ivar.py`
* alphabetize `mf6ivar.md`, remove old manual ordering
* apply some unapplied updates to tex/dat/md files

PR #1745, opened 2024-04-18
refactor(mf6ivar): discover dfns dynamically

wpbonelli:
* look up dfns in `doc/mf6io/mf6ivar/dfn/` instead of hardcoded list
* preserve top-level component ordering: sim, exg, sln, models, utl
* order packages alphabetically within each component section
* add argparse CLI allowing model selection, default to all
* carved out from #1540 
* step toward #1491


langevin-usgs (approved):
thanks for the cleanup, @wpbonelli

PR #1756, opened 2024-04-23
feat(PRT): localz option for particle release point package

wpbonelli:
Optional support for local z coordinates. This allows positioning particles based on the water table. The motivations here are 1) modpath 7 parity and 2) to provide the same capabilities whether prt is running alongside a gwf model or as a post-processor. Previously dynamic positioning was possible only if prt runs as a post-processor.

Motivated by https://github.com/MODFLOW-USGS/modflow6-examples/pull/171.


aprovost-usgs (approved):
looks good

PR #1758, opened 2024-04-23
fix(test_gwe_drycell_cond0): minor alteration to gwe autotest

emorway-usgs:
test_gwe_drycell_cnd0.py was reportedly having issues passing on an M1 chip with numbers in 1e-12 range.  This fix cured the `assert` failure.


langevin-usgs (commented):
Should we follow our normal convention of passing in a value for atol into np.allclose() instead of rounding?

emorway-usgs:
> Should we follow our normal convention of passing in a value for atol into `np.allclose()` instead of rounding?

Because the check is confirming an inequality, I'm not sure `np.allclose()` is useable here.

PR #1765, opened 2024-04-26
feat (memorymananger) Create iterators for the ListType and MemoryListType

Manangka:
In this PR Iterators are added for the MemoryListType and the ListType.
Iterators are a way of iterating though a container while 'remembering' where you are in the iteration loop.

This is very useful for containers that don't provide random access e.g. lists, hashtables, etc

It is also an agnostic way of going through a containers items. This make it possible to write loops that are not depended on the container type. One of the benefitrs being that it makes it easier to switch from container type without triggering an avalanche of changes.

This commit is made in preparation of switching the internal data container from a list to a hashtable.
It should make it possible to simply switch the container types without many changes needed.


src/Utilities/Iterator.f90
```diff
@@ -0,0 +1,37 @@
...
+    procedure(value_if), deferred :: value
+
+  end type IteratorType
+
+  abstract interface
+
+    function has_next_if(this) result(res)
+      import IteratorType
+      class(IteratorType) :: this
+      type(logical) :: res
```
mjr-deltares, line 21: Although debatable, we generally work with kind LGP for logicals, so: logical(LGP) :: ...

src/Utilities/ListIterator.f90
```diff
@@ -0,0 +1,89 @@
...
+    module procedure Constructor
+  end interface ListIteratorType
+
+contains
+
+  !> @brief Constructor to create a ListIterator
+  !!
+  !<
+  function Constructor(first_node) result(iterator)
+    ! -- dummy
```
mjr-deltares, line 34: I would remove the dummy comments

mjr-deltares (approved):
Looks really good! Please have a look at my one or two style suggestions, and probably pull in develop once more. Then I think it's ready to go in.

PR #1769, opened 2024-05-03
feat(par): add HPC configuration file for parallel

mjr-deltares:


src/Utilities/Idm/SourceLoad.F90
```diff
@@ -168,6 +171,13 @@ subroutine load_simnam()
       !
       ! -- open namfile and load to input context
       call input_load(simfile, mf6_input, simfile, iout)
+      !
+      ! -- load optional HPC configuration file
+      if (filein_fname(hpc6_filename, 'HPC6_FILENAME', mf6_input%mempath, &
+                       simfile)) then
+        hpc_input = getModflowInput('HPC6', 'UTL', 'HPC', 'UTL', 'HPC', simfile)
```
mjreno, line 178: Line 178 can remove the optional argument "simfile" as it is used to distinguish an array based package input set.

src/Utilities/Idm/SourceLoad.F90
```diff
@@ -168,6 +171,13 @@ subroutine load_simnam()
...
+        call input_load(hpc6_filename, hpc_input, hpc6_filename, iout)
```
mjreno, line 179: In this line the 3rd parameter "hpc6_filname" (second occurrence) should instead be "simfile".  This second filename argument is used to set an error associated with opening the file as that is where the filename originated from.

langevin-usgs (approved):
This is looks really clean, @mjr-deltares.  Really like how it cleans up simulation_create.  

At some point, it might be nice to add some words to described the concept of "distributed simulation".  The one thing about the word "distributed" is that it sounds like the simulation is distributed across multiple cpus, but it seems that "distributed" indicates the manner of distribution, which may be non-distributed.  I think the "distributed" terminology is fine, but just wondering if there are alternatives.  Maybe cpu_layout?  Not sure that is any better.  Either way, containerizing it is really nice.

Manangka (approved):

PR #1771, opened 2024-05-03
build: modify meson build file to select the correct C/C++ runtime

jdhughes-dev:


jdhughes-usgs:
This is what meson does on Windows with MT and MTd libraries

image

jdhughes-dev:
Superseded by #2299

PR #1780, opened 2024-05-08
fix(par): fix backtracking and residual norm calculation for certain cases in parallel

### LINKED ISSUE #1779: Residual L2 norm calculation incorrect for parallel simulation with Backtracking and/or PTC
After analyzing #1778 it appears that the residual calculation in the outer iteration loop (used for PTC and Backtracking logic) is not adequate. This can affect results in parallel simulations with PTC or Backtracking enabled, e.g. when complexity equals 'COMPLEX'. The fix is to always update the PETSc internal matrix object before calculating the residual.

Other possible issue is that application of the backtracking x update should be a globally synchronized decision. This was not the case yet. For the reported model that didn't cause any issues, but it is related to parallel backtracking and will therefore be included in this fix.

mjr-deltares:
- update petsc matrix before multiplication
- backtracking now properly synchronized

this closes #1779

### DIFF
--- a/src/Solution/NumericalSolution.f90
+++ b/src/Solution/NumericalSolution.f90
@@ -2796,9 +2796,9 @@ contains
     ! -- dummy variables
     class(NumericalSolutionType), intent(inout) :: this !< NumericalSolutionType instance
     integer(I4B), intent(inout) :: bt_flag !< backtracking flag (1) backtracking performed (0) backtracking not performed
-    
+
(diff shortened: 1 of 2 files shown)


mjr-deltares:
Hi @jdhughes-usgs , looks like this is going to be green. Feel free to merge if you agree with the changes, so you can try it out.

jdhughes-usgs (approved):

PR #1781, opened 2024-05-08
fix(mf6io): synchronize mf6io with descriptions in forthcoming SuppTechInfo chapter

emorway-usgs:
For clarity, the new chapter in SuppTechInfo (PR not quite ready) that will accompany the official release of GWE refers to the new conduction package as the Conduction and Dispersion Package, which helps make clear all the parameters that are included in the package.  In heat transport, conduction is likely the dominant process and therefore takes the lead in the package name.  However, mechanical dispersion may still play a secondary role.  So that users know where to go to enter dispersion related parameters, the package name is being updated.


langevin-usgs (approved):
Me gusta

doc/mf6io/gwe/cnd.tex
aprovost-usgs, line 1: XT3D is on by default and typically desired, but not sure I would say CND is "based on" the XT3D formulation.  We could unhook XT3D and you'd still be able to do (probably inaccurate) GWE simulations. That said, I realize this wording parallels what's in the DSP description.  If we decided to change it here, then presumably we'd also change it there.

I'm not sure the tortuosity issue merits mention right up front. (There are bigger assumptions and limitations we tackle in supptechinfo but not mf6io.) Might it be better to just add notes to the parameter descriptions for the thermal conductivities of water and solid?  Those are the parameters the user would adjust for tortuosity.

Would be happy to discuss.

aprovost-usgs (commented):
Thanks for preparing this, Eric.  I'd be happy to discuss my comments.

Issue #128, opened 2019-04-02
binary exchange data


jdhughes-usgs:
evaluate option to specify exchange data as binary data

    ```
    BEGIN EXCHANGEDATA
        <cellidm1> <cellidm2> <ihc> <cl1> <cl2> <hwva> [<aux(naux)>]
        <cellidm1> <cellidm2> <ihc> <cl1> <cl2> <hwva> [<aux(naux)>]
        ...
    END EXCHANGEDATA
    ```
Support this option

    ```
    BEGIN EXCHANGEDATA
        OPEN/CLOSE exchange.exc (BINARY)
    END EXCHANGEDATA
    ```

langevin-usgs:
@mjreno, is this supported now by using IDM (Input Data Model) loading routines for exchanges (#1505)?

mjreno:
@langevin-usgs This hasn't been tested yet but should be close if not already done. I'll find or generate a suitable input and test and then will update this issue.

langevin-usgs:
@mjreno, I suspect this is supported now.  If so, can you close this PR?

mjreno:
@langevin-usgs apologies for losing track of this one, will verify and close.

langevin-usgs:
> @langevin-usgs apologies for losing track of this one, will verify and close.

No probem!  Was just looking through these issues and think some might be fixed now because of IDM.

mjreno:
@langevin-usgs just thinking about this a bit- the struct array input reader already can read similar binary inputs.  The current use case is for stress period data, however, so currently the control record is read and binary read interface invoked only in the dynamic loader.  The change needed would be update the static loader to do the same.

If this sounds right to you it wouldn't be a lot of work on the MODFLOW 6 side- would FloPy also need to be updated to generate binary exchange list style input?

mjreno:
Closing as this has been addressed with #1784 and #1783.

PR #1783, opened 2024-05-12
feat(idm): support binary for regular, numeric list input blocks

mjreno:
Support binary for any recarray style input block which is regular and contains no text columns
  - Includes, for example, DISV cell2d if ncvert is constant
  - Includes exchange exchagedata block if BOUNDNAMES is not optioned (#128)
  - Not currently supported by FloPy- how to test in ci for now?


langevin-usgs (approved):
Hey @mjreno, this looks good.  Too bad about EXCHANGEDATA being explicitly mentioned, but maybe we can think about how to make that generic in the future.  Any thoughts on adding a simple test?  Or perhaps we can do that next after this is in, and also update the release notes.

Maybe in the future, we can think about support for binary text columns.  Seems like that might be possible with IDM.

mjreno:
@langevin-usgs I do have some thoughts about how to remove the EXCHANGEDATA reference.  It might be straightforward and can look at it before or after a merge.  I agree that it seems like text should be possible and would like to look at that more- I think also IDM could read an irregular array if the binary file can be generated.  

As far as testing, I would like to update a couple of existing tests so that they read some existing input blocks from binary.  I have the binary files but not sure of the best way to temporarily add.  Maybe @wpbonelli will have an idea.  I'll add a release note.

PR #1802, opened 2024-05-15
style(mf6core): corrected minor spelling errors in comments

vivekbedekar:


src/mf6core.f90
```diff
@@ -330,7 +330,7 @@ subroutine simulation_df()
     ! -- synchronize
     call run_ctrl%at_stage(STG_AFT_EXG_DF)
     !
-    ! -- when needed, this is were the interface models are
+    ! -- when needed, this is where the interface models are
     ! created and added to the numerical solutions
```
emorway-usgs, line 334:
```suggestion
    !     created and added to the numerical solutions
```

emorway-usgs: A minor stylistic suggestion: Throughout the code, multiline comments are typically justified such that the text on the second (and higher) line has the same left justification as the text appearing on the first line of the comment (i.e., the '--' does not count as part of the left justification)

langevin-usgs (approved):

PR #1803, opened 2024-05-15
refactor(prt): cleanup, refactoring, optimization in dis/disv methods

wpbonelli:
* factor out routines for cleaner `MethodDisType` and `MethodDisvType`
* optimize cell definition load: merge loops when setting cell faceflows
* rename `destroy()` -> `deallocate()` in method types
* misc cleanup, improve docstrings


wpbonelli:
I need to fix a problem in the `setup-fortran` action causing the macos-14/gcc-13 failure. In the meantime I think this need not block this PR or others

PR #1815, opened 2024-05-16
fix(BudgetTerm.f90): ensure that the size of an array is >0 before deallocating

emorway-usgs:
- Fix offered in BudgetTerm was discovered using models that included only a single `SFT`/`SFE` reach.  Previously, `GWF` models with a single `SFR` reach would work.  Only transport models would fail.
- 2 new autotests offered that each use only a single reach
  - The first autotest is for a GWT model (`test_gwt_uztmvt2x1`). This test ensures that in addition to a single reach, a single `MVT` connection between `UZT` and `SFT` works as expected
  - The 2nd autotest is for a GWE model (`test_gwe_drycell_cnd3`). This test checks that conduction between `SFE` and a dry host cell is working as expected in a single-reach model 
- Extended an existing single-reach SFR autotest to include SFT.  Prior to this fix, the added GWT model with a single SFT reach would fail because of the bug.
- Brief description of bug offered in release notes
- fprettify
- black


christianlangevin:
I remember glancing at this a while back.  I think I understand that you are intending to not write any flows for the single reach case and instead remove the `flowjaface` record entirely.  I wonder if this is the best way to handle it.  Perhaps it would be better to still include the header but just not write any records?  I'm thinking that this may break other downstream processors, like zone budget.  Do we have a similar problem for LAK/MAW/UZF?  Can you remind me why this is a problem for SFT but not SFR (I think we discussed this once before).

emorway-usgs:
I will need to refamiliarize myself with the issue.  As I recall from reading my previous notes, I think the issue had something to do with the `flowjaface` budget term in single reach models (meaning its not needed).  The transport model was either (1) not expecting this budget term and it was present, thereby throwing off the indexing or (2) was expecting this budget term and it was missing by virtue of there being no connected reaches.

PR #1823, opened 2024-05-17
feat(idm): update STO packages for IDM

mjreno:
Note:  The Idm STO loader initializes storage to "TRANSIENT", which corresponds to package steady state flag (iss) of 0 .  Initializing the input side string to empty causes a failure in tests like test_gwf_wel01.py, which have one empty period block in the storage input.  If the test isn't valid (or we shouldn't support such inputs), the test will be updated and the input side storage string will be initialized to an empty string.  This will have the effect of forcing the first defined period block of a storage input to be not empty.


langevin-usgs (approved):
Looks good.

PR #1831, opened 2024-05-20
test: mark some parallel tests developmode

wpbonelli:
Add the `developmode` marker to parallel tests that reuse models from tests that use dev options


mjr-deltares:
@wpbonelli when you have some time, can you explain me why we are marking these tests as developmode

wpbonelli:
@mjr-deltares they are based on other test cases that use dev options. For instance `test_gwe_split_analyt.py` uses `dev_interfacemodel_on`

mjr-deltares:
OK, got it. Let's have a look **after** the release to see how to avoid that from happening.

PR #1866, opened 2024-06-11
fix(inputoutput): handle line ending skipped by getline for cr or lf

vivekbedekar:


src/Utilities/InputOutput.f90
```diff
@@ -1972,6 +1974,28 @@ subroutine get_line(lun, line, iostat)
         exit
       end if
     end do
+    !
+    ! -- look for undetected end-of-record with isolated CR or LF
+    linesize = len(line)
+    crlfcheck: do i = 1, linesize
+      if (line(i:i) .eq. cr .or. line(i:i) .eq. lf) then
```
langevin-usgs, line 1981: I must be missing something fundamental here.  The outermost if statement looks for CR or LF in line(i:i).  Then, the inner if statements again check for CR or LF in line(i:i).  How does this check for an isolated CR or LF?  I'm sure you tested this so you know it works -- I just can't see why.

vivekbedekar: The outer if statement identifies an isolated CR or LF. The inner if statement only prints the appropriate message for either CR or LF. Both CR and LF are not expected to be present in the middle of the line [line(i:i)]. If, for any reason, CR and LF appear in the middle of a line, that is still a bug.

langevin-usgs:
@vivekbedekar, thanks for this.  I will go ahead and merge.  Can you please add a release notes description on this PR as a comment and we will add it to the release notes.  Maybe something like: "For ASCII input files erroneously containing a mix of line endings, MODFLOW would sometimes proceed with unexpected results.  The program was corrected to stop with an error message if an input line contained both carriage returns and line feeds."  Ideally you would add this to the release notes yourself (under ./doc/ReleaseNotes/develop.tex) and include in the PR, but this is a latex file and I'm not sure you are set up to work with latex yet.

langevin-usgs:
> a mix of line endings, MODFLOW would sometimes proceed with unexpected results. The program was corrected to stop wit



> @vivekbedekar, thanks for this. I will go ahead and merge. Can you please add a release notes description on this PR as a comment and we will add it to the release notes. Maybe something like: "For ASCII input files erroneously containing a mix of line endings, MODFLOW would sometimes proceed with unexpected results. The program was corrected to stop with an error message if an input line contained both carriage returns and line feeds." Ideally you would add this to the release notes yourself (under ./doc/ReleaseNotes/develop.tex) and include in the PR, but this is a latex file and I'm not sure you are set up to work with latex yet.

Sorry, I just noticed that you did update the release notes in this PR.  Sorry I missed that.  But it might be good to update with a better description of the change.  Can you look at my proposal and revise as appropriate?  Thanks.

Issue #1753, opened 2024-04-21
Missing CSV_INNER_OUTPUT columns for single models in parallel


tandreasr:
Hi again,
while testing the new parallel capabilites I realized that the CSV file(s) created by using the above mentioned option lack(s) all columns related to single models (MODELNAME_inner_dvmax etc.)
Instead they only contain the convergence summary for all models processed by the corresponding MPI rank.
Do you see any chance to implement that output the very same way as IMS does within a "foreseeable" future.
I think that sometimes it's quite important to obtain an overview about which models have convergence issues
(not just the worst one listed in solution_inner_d?max_model).
Regards
Andreas

tandreasr:
One more addition to this issue - actually this is a different topic connected to the CSV in parallel.
Obviously something is still wrong with the columns "solution_inner_dvmax_model" and "solution_inner_drmax_model".
In the example CSV below they are uninitialized in the first couple of rows and always 1 afterwards.
Instead they should contain the model number associated with the MPI rank.

![Bild_2024-04-22_083124867](https://github.com/MODFLOW-USGS/modflow6/assets/13199371/3af5f643-7aea-4903-b566-8ebe05598d0d)

mjr-deltares:
Hi @tandreasr , thanks for reporting this back. Will look into this to have it fixed.

mjr-deltares:
HI @tandreasr , I am looking at your issue and maybe you can provide a minimal example? I was a bit confused by:

_Instead they only contain the convergence summary for all those models processed by the corresponding MPI rank.
...
Meaning to add those columns to the CSV belonging to the MPI rank for all models processed by that rank)_

As far as your question about keeping output the same. We stayed clear so far from synchronizing over all processes to gather output in a single file where possible, hence the *.p<rank>.csv format. Also, IMS has an additionally a solver parameter (alpha/omega) that we do not print when using the parallel PETSc solver. Other than that, we could definitely look into improving the parallel csv output. I did reproduce the 'uninitialized' issue you mentioned in your second comment. Will have a look into that.

mjr-deltares:
I do think that I understand your point now, you are talking about these columns missing from the csv:

![image](https://github.com/MODFLOW-USGS/modflow6/assets/45555666/ef283c1a-3661-44e0-b87c-bfea2d8ffcab)

Will look into that too.

mjr-deltares:
Hi @tandreasr , this should be fixed now. Please reopen if not.

[...]

PR #1869, opened 2024-06-12
feat(par): add model columns to convergence csv for completeness in parallel case

mjr-deltares:


mjr-deltares:
this is follow up from a closed issue, see #1753

PR #1883, opened 2024-06-18
ci(pr-template): modify pr template

emorway-usgs:
add reminder to run black on new and/or modified autotests


wpbonelli (approved):
maybe worth switching to ruff here as elsewhere eventually

PR #1888, opened 2024-06-20
fix(prt): make extended tracking opt-in

wpbonelli:
Previously the PRT model's default behavior was to track particles until termination, as with MODPATH 7 stop time option 2 (extend). Under extended tracking, the program may not halt if particles enter a cycle in the flow system. This PR changes the default to the equivalent of MP7 stop time option 1 (final), terminating at simulation end unless a new Particle Release Point (PRP) package keyword option `EXTEND` is provided. This is meant to provide a stronger guarantee that the program halts under default settings.

The extend option is stored as an integer rather than logical in case we need to enumerate more alternatives in future.

This PR also avoids unnecessary iterations in `MethodSubcellPollock/Ternary`, and changes the default visibility of `ParticleStore` members to private, as per style conventions.

___

- [x] Ran `black` on new and modified autotests


langevin-usgs (approved):
@wpbonelli, thanks for doing this.  I think this will save some trouble.  One thing that that jumps out at me is if there is a clearer keyword to use for this option. Certainly "EXTEND" is fine and is clearly described.  Maybe "EXTEND_TRACKING" or something with TIME in it?  

Also wonder if it would be worth explicitly mentioning that if the extend option is specified that particles will continue to be tracked using the flows from the last time step.  This is probably obvious but I think it's good to mention.  For a transient simulation with a changing flow field, flows from the last time step may not be a good indicator of conditions past the end of the simulation.

wpbonelli:
Thanks @langevin-usgs, renamed `EXTEND` -> `EXTEND_TRACKING` and mentioned that the flows from the final time step are used for extended tracking.

PR #1891, opened 2024-06-21
Build the sphinx documentation on windows

Manangka:
This commit enables building the sphinx documentation on windows.
The make.bat needed to do so was missing and is added in this commit.

To build the documentation from a fresh git clone you need to run these commands:
```
pixi install
pixi run update-flopy
pixi run get-exes
pixi run setup builddir
pixi run build builddir
pixi run benchmark
pixi run sphinx
```

I've also cleaned the pixi.toml file. I added the git dependencies on flopy, pymake, modflowapi and modflow-devtools as pypi-dependencies. This change made it possible to remove the install task.

Furthermore I added the rtd environment to the pixi.toml. When building sphinx you now don't have to manually pip install 
the requirements.rtd.txt. Instead the task will automatically use the rtd environment

Finally i updated the github pipeline files. The newly generated pixi.lock is using version 5 which needs a newer pixi version. I also removed the `pixi run install` as this is no longer required

- [ ] Closed issue #xxxx
- [ ] Ran `black` on new and modified autotests
- [ ] Updated [definition files](/MODFLOW-USGS/modflow6/tree/develop/doc/mf6io/mf6ivar)
- [ ] Updated [develop.tex](/MODFLOW-USGS/modflow6/doc/ReleaseNotes/develop.tex) with a plain-language description of the bug fix, change, feature; required for changes that may affect users
- [ ] Updated [input and output guide](/MODFLOW-USGS/modflow6/doc/mf6io)


pixi.toml
```diff
@@ -17,7 +17,7 @@ matplotlib = "*"
 meson = "1.3.0"
 networkx = "*"
 ninja = "*"
-numpy = "*"
+numpy = "<2.0.0"    # modflowapi has a resriction on the numpy version
```
wpbonelli, line 20: we need to release a new modflowapi supporting numpy2, thanks for the reminder &mdash; then this can be unpinned

pixi.toml
```diff
@@ -34,19 +34,35 @@ shapely = "*"
...
-install-pymake = "pip install --no-build-isolation --no-deps --disable-pip-version-check git+https://github.com/modflowpy/pymake.git"
-install-modflowapi = "pip install --no-build-isolation --no-deps --disable-pip-version-check git+https://github.com/MODFLOW-USGS/modflowapi.git"
-install-modflow-devtools = "pip install --no-build-isolation --no-deps --disable-pip-version-check git+https://github.com/MODFLOW-USGS/modflow-devtools.git"
-install = { depends_on = [
-    "install-flopy",
-    "install-pymake",
-    "install-modflowapi",
-    "install-modflow-devtools",
-] }
+[pypi-dependencies]
```
wpbonelli, line 37: I agree it would be ideal to use `pypi-dependencies` instead of custom tasks, and originally did so in #1703, but after discussing with @jdhughes-usgs @langevin-usgs we went back to custom tasks in #1728 to get rolling CI updates for internal developer deps without needing to update the `pixi.lock` file.

wpbonelli (changes requested):
Just a note re: the custom install task. All else looks great to me.

pixi.toml
Manangka: Ah I see. That's something I overlooked. I will revert back to using tasks

Manangka: I'm wondering if this is actually bad or not. The update-flopy task still works. 

The normal way of working would be:
- `pixi run install`
- some changes happen in flopy that you need
- `pixi run update-flopy`
- repeat above step when newer flopy is needed

These steps don't change when we switch to using `pypi-dependencies`

It is true that when we run the `update-flopy` we should update the lock file as well but it is not necessary.
This is because in the CI pipeline we call the  `update-flopy` task causing the locked version to lose its meaning anyway

wpbonelli: The name of the `update_flopy.py` script/`update-flopy` task is a bit misleading, it's not actually upgrading the flopy package, just regenerating mf6 components from DFNs.

The idea is if we use `pypi-dependencies` we just need to update the lockfile frequently as flopy/modflowapi/pymake/devtools tend to move faster than external dependencies and mf6 and flopy are often developed in sync. With a task we always get the latest in CI which is convenient for development if a little harder to debug.

Manangka: Alright. Thanks for this explanation.
I now get why the install step is required and restored that part.

wpbonelli (approved):

PR #1895, opened 2024-06-24
docs(styleguide) Add formatting and explanations to the Style guide 

Manangka:
In this PR the styleguide has been extended

- The rules have grouped in different sections to which can be linked
- Examples have been added to most of the rules to clarify the intent

Checklist of items for pull request

- [ ] Closed issue #xxxx
- [ ] Referenced issue or pull request #xxxx
- [ ] Added new test or modified an existing test
- [ ] Ran `black` on new and modified autotests
- [ ] Formatted new and modified Fortran source files with `fprettify`
- [ ] Added doxygen comments to new and modified procedures
- [ ] Updated meson files, makefiles, and Visual Studio project files for new source files
- [ ] Updated [definition files](/MODFLOW-USGS/modflow6/tree/develop/doc/mf6io/mf6ivar)
- [ ] Updated [develop.tex](/MODFLOW-USGS/modflow6/doc/ReleaseNotes/develop.tex) with a plain-language description of the bug fix, change, feature; required for changes that may affect users
- [ ] Updated [input and output guide](/MODFLOW-USGS/modflow6/doc/mf6io)
- [ ] Removed checklist items not relevant to this pull request

For additional information see [instructions for contributing](/MODFLOW-USGS/modflow6/.github/CONTRIBUTING.md) and [instructions for developing](/MODFLOW-USGS/modflow6/.github/DEVELOPER.md).

PR #1896, opened 2024-06-24
refactor(sfr): implement submodules for SFR flow methods

jdhughes-usgs:


src/Model/GroundWaterFlow/submodules/sm-gwf-sfr-constant.f90
```diff
@@ -0,0 +1,45 @@
+submodule(SfrModule) sm_gwf_sfr_constant
+contains
+
+  !> @brief Method for solving for standard reaches
+    !!
+    !! Method to solve the continuity equation for a SFR package
+    !! reach using the standard approach.
```
langevin-usgs, line 7: Is this routine description correct?  Seems like it should describe how we solve for constant stage reaches?

src/meson.build
```diff
@@ -162,6 +162,8 @@ modflow_sources = files(
     'Model' / 'GroundWaterFlow' / 'gwf-uzf.f90',
     'Model' / 'GroundWaterFlow' / 'gwf-vsc.f90',
     'Model' / 'GroundWaterFlow' / 'gwf-wel.f90',
+    'Model' / 'GroundWaterFlow' / 'submodules' / 'sm-gwf-sfr-steady.f90',
```
langevin-usgs, line 165: If these are already in a folder named submodules, do they need an "sm" prefix in the file name?  Maybe we can drop the "sm"?

src/Model/GroundWaterFlow/submodules/sm-gwf-sfr-steady.f90
```diff
@@ -0,0 +1,305 @@
+submodule(SfrModule) sm_gwf_sfr_steady
```
langevin-usgs, line 1: Just throwing this out there for discussion -- if the submodule is part of SfrModule, then does the submodule name need to include "gwf_sfr" in it?  (as a separate note, we probably need to rename SfrModule to GwfSfrModule at some point to be consistent with other parts of the code).

In following with our other naming conventions, should the submodule name, then be SteadySubmodule since it by definition determined to part of gwf_sfr?

src/Model/GroundWaterFlow/submodules/sm-gwf-sfr-constant.f90
```diff
@@ -0,0 +1,45 @@
...
+    !!
+  !<
+  module subroutine sfr_calc_constant(this, n, d1, hgwf, qgwf, qd)
+    ! -- dummy variables
```
langevin-usgs, line 11: Minor annoying comment -- if we are ditching the "--" in comments, maybe start off fresh with these two new submodule files?

langevin-usgs (commented):
I like the submodules, and after reading about them, they do offer nice compiling benefits as well as better organization.  I added a few annoying comments about naming in case we want to fine tune that a bit, but going with what's here now is also fine by me.

Since we are moving into submodules, we should probably add a section to the style guide under development in #1895 at some point.

PR #1906, opened 2024-06-27
refactor(autotest): restructure netcdf tests to remove duplicate code

mjreno:

- [x] Ran `black` on new and modified autotests


langevin-usgs:
I like where this is headed.  An alternative way to do this, which comes with its own complications, would be to leave the original test files alone, and then modify the resulting simulation using some form of the following:

```
sim, dummy = build(idx, test)
gwf = sim.gwf[0]
dis = gwf.dis
dis.export_array_netcdf = True
```

Just a thought.

langevin-usgs (approved):

PR #1926, opened 2024-07-05
fix(meson.build): separate parallel/extended dependencies

wpbonelli:
Separate logic for parallel and extended builds. Previously a netcdf dependency was declared even for a parallel-only (not extended) build. This can cause issues if meson detects a netcdf that has been improperly installed or incompletely setup as seems to now be the case on Mac CI runner images

___


mjr-deltares:
Thanks @wpbonelli , for dealing with this!

PR #1945, opened 2024-07-12
MSVS File Update and Fixes

ScottBoyce:
This PR contains two commits. One that updates the visual studio solution and project files associated with the mf6.exe and mf6.dll and the other updates the same files for zonebudget and MF5to6. They all still use **ifort** as the fortran compiler. 

This will not invoke any tests, but I ran `pixi run autotest` on my local machine. All tests pass except for the following:
```
FAILED test_gwf_libmf6_evt01.py::test_mf6model[0-libgwf_evt01] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_libmf6_ghb01.py::test_mf6model[0-libgwf_ghb01] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_libmf6_ifmod01.py::test_mf6model[0-libgwf_ifmod01] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_libmf6_ifmod02.py::test_mf6model[0-libgwf_ifmod02] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_libmf6_ifmod03.py::test_mf6model[0-libgwf_ifmod03] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_libmf6_rch01.py::test_mf6model[0-libgwf_rch01] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_libmf6_rch02.py::test_mf6model[0-libgwf_rch02] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_libmf6_riv01.py::test_mf6model[0-libgwf_riv01] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_libmf6_riv02.py::test_mf6model[0-libgwf_riv02] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_libmf6_sto01.py::test_mf6model[0-libgwf_sto01] - UnboundLocalError: local variable 'buff' referenced before assignment
FAILED test_gwf_tdis.py::test_tdis_tsmult[1.0] - FileNotFoundError: Could not find module 'C:\Data\dev\mf6-dev\modflow6\bin\libmf6.dll' (or one of its dependencies). Try using the full path with constructor syntax.
FAILED test_gwf_tdis.py::test_tdis_tsmult[1.2] - FileNotFoundError: Could not find module 'C:\Data\dev\mf6-dev\modflow6\bin\libmf6.dll' (or one of its dependencies). Try using the full path with constructor syntax.
ERROR test_gwf_libmf6_rch02.py::test_mf6model[0-libgwf_rch02] - shutil.Error: [('C:\\Users\\seboyce\\AppData\\Local\\Temp...
```

I am unsure as to the cause of the libmf6.dll errors as the file is placed in the bin directory. It could also be that these tests have not been run with VS made dll targets in a while (or some quark of Windows). I will keep looking into it, but the original version of the dll project files would not compile at all for me initially.

The bulk of the changes are described in the git commit log. The major changes are that win32 is removed and all projects are setup to compile x64. The sln/vfproj files are updated to VS2022 format, which is backward compatible with VS2019.

The major fix is that mfbmi needed several modification to successfully compile with the current version of ifort in oneAPI (2021.12). In particular, mf6bmi required ignoring the LIBCMT, LIBCMTD, and LIBIFCOREMT (not originally there) and mf6 no longer required ignoring LIBCMT.

The project for  Zonebudget now produces "zbud6.exe" as its output rather than "zonebudget.exe"

The Zonebudget and MF5to6 now produce their executables in the bin folder,
rather then using a post-build "copy" event.

Added /Qdiag-disable:10448 to suppress warning about ifort being deprecated.

The intermediate directory was changed from  
`$(ProjectName)/$(Platform)/$(ConfigurationName)`  
to  
`$(ProjectName)_$(Platform)_$(ConfigurationName)`  
to simply the structure from having multiply directories that contain nothing but a single directory.

### DIFF
--- a/msvs/mf6.sln
+++ b/msvs/mf6.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30907.101
+# Visual Studio Version 17
+VisualStudioVersion = 17.10.35004.147
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{6989167D-11E4-40FE-8C1A-2192A86A7E90}") = "mf6", "mf6.vfproj", "{369C1E2E-A513-4E83-A076-923534E30304}"
 	ProjectSection(ProjectDependencies) = postProject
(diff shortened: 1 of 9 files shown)


wpbonelli:
Maybe worth checking if the same test failures occur in a CI environment? Adding an MSVS build should be easy as we now have an [action](https://github.com/MODFLOW-USGS/modflow6/blob/develop/.github/actions/setup-par-oneapi/action.yml) to install base/hpc kits. Then I think we can just run `devenv` [like this](https://github.com/sezero/quakespasm/pull/69/files#diff-01d393d5fc96ff0e19c736f6b445df76a62887b81071928a950e936379746f2dR17).

xref #1357 which was abandoned due to the oneapi installation overhead. Seems more justifiable now that we are already doing it for the Windows parallel test.

mjr-deltares:
I could reproduce the errors locally with the msvs build of the library. They were related to the Intel runtime not being found. If I change the project runtime setting to use static linking (Multithreaded), and then the issue is resolved. Not sure if that got changed somehow along the way? Maybe installing the runtime through a redistributable would fix it too.

I have used an older intel: Compiling with Intel® Fortran Compiler Classic 2021.11.0 [Intel(R) 64]...

Hope this helps you forward with your issue.

Issue #1944, opened 2024-07-12
GWE Observations: Temperature Observation Type Not Found 


MickeyRush:
Maybe I'm jumping the gun on GWE, but it looks like Temperature observations are not recognized:

ERROR REPORT:

  1. Observation type not found: TEMPERATURE

UNIT ERROR REPORT:

  1. Error occurred while reading file
     'gwe.obs'

My gwe.obs file is of the format:
BEGIN options
END options

BEGIN continuous  FILEOUT  gwe_obs.csv
 WELLNO_13_11.0  TEMPERATURE  1 37 15
  ...
END continuous FILEOUT gwe_obs.csv

emorway-usgs:
You're not jumping the gun.  I'll have a look.  Would you mind sharing your GWE name file?

MickeyRush:
Thanks - here it is:

# File generated by Flopy version 3.7.0 on 07/12/2024 at 16:36:04.
BEGIN options
END options

BEGIN packages
  DIS6  ccg_gwe.dis  dis
  IC6  ccg_gwe.ic  ic
  OC6  ccg_gwe.oc  oc
  OBS6  ccg_gwe.obs  obs_gwe
  ADV6  ccg_gwe.adv  adv
  CND6  ccg_gwe.cnd  cnd
  EST6  ccg_gwe.est  est
  SSM6  ccg_gwe.ssm  ssm
  ESL6  ccg_gwe.esl  esl
END packages

emorway-usgs:
@MickeyRush, This issue should be fixed with #1950.  Do you have the ability to compile fortran?  If not, I can get you a working copy of the exe, just email me at my usgs email emorway at usgs dot gov

wpbonelli:
@emorway-usgs note that when the PR is merged it will become available later that day in the [nightly build](https://github.com/MODFLOW-USGS/modflow6-nightly-build)

MickeyRush:
Thanks @emorway-usgs and @wpbonelli I will grab the linux executable when the PR is merged.

emorway-usgs:
Fixed via #1950.  @MickeyRush, Thanks for helping us identify this bug.  Per @wpbonelli's previous post on this thread, an executable should be available this evening at the link Wes provided.  

If you spot some other issues with GWE, letting us know about 'em here is appreciated.

wpbonelli:
the nightly build is broken at the moment, working on a fix...

MickeyRush:
[...]

PR #1950, opened 2024-07-15
fix(tsp-obs): the OBS utility in GWE not supporting TEMPERATURE keyword

emorway-usgs:
The mf6io guide appropriately included a section on using the OBS utility with a GWE model; however, support in the MF6 source code was lacking for a TEMPERATURE `obstype`.  This issue was originally pointed out in #1944 

- [x] Fixes issue #1944

### DIFF
--- a/autotest/test_gwe_obs.py
+++ b/autotest/test_gwe_obs.py
@@ -64,11 +64,13 @@ scheme = "UPSTREAM"
 # scheme = "TVD"
 
 cases = [
-    "horiz-obs",  # 4-cell model, horizontally connected, check adv & cnd
+    "hrz-obs",  # 4-cell model, horizontally connected, check adv & cnd
+    "hrz-badobs",  # Specify a 'concentration observation in a gwe model - should fail
     "vert-obs",  # 4-cell model, vertically connected, check adv & cnd
 ]
 
 conn_types = (
+    "horizontal",
     "horizontal",
     "vertical",
 )
@@ -83,6 +85,14 @@ dis_data = {
         "strthd": np.ones(4, dtype=float),
     },
     conn_types[1]: {
+        "nrow": 1,
+        "ncol": 4,
+        "nlay": 1,
+        "top": np.ones(4, dtype=float),
+        "bot": np.zeros(4, dtype=float),
+        "strthd": np.ones(4, dtype=float),
+    },
+    conn_types[2]: {
         "nrow": 1,
         "ncol": 1,
         "nlay": 4,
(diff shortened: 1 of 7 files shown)


langevin-usgs (commented):
I see that `TEMPERATURE` is added as an observation type to tsp-obs, but I wonder if we need a way for the concrete models (GWE, GWT) to add these through a subclass.  FLOWJA can stay in the parent because it is available to both model types.  Otherwise, what happens when someone specifies CONCENTRATION for a GWE model or TEMPERATURE for a GWT model?  Okay with kicking this down the road for now, but this might be a good opportunity to think this through.

emorway-usgs:
> I see that `TEMPERATURE` is added as an observation type to tsp-obs, but I wonder if we need a way for the concrete models (GWE, GWT) to add these through a subclass. FLOWJA can stay in the parent because it is available to both model types. Otherwise, what happens when someone specifies CONCENTRATION for a GWE model or TEMPERATURE for a GWT model? Okay with kicking this down the road for now, but this might be a good opportunity to think this through.

Thanks for the nudge to do this better @langevin-usgs.  I'm hopeful that [aae590c](https://github.com/MODFLOW-USGS/modflow6/pull/1950/commits/aae590caec16cc704deb7135c58b67e6c879a9ef) addresses these concerns.  Initially I had worked up some error checking in response to this comment (which I'm pasting here in case I need to come back to it):
```
  !> @brief Check observation type against name of dependent variable
  !<
  subroutine check_obs_id(this)
    ! -- modules
    use InputOutputModule, only: upcase
    ! -- dummy
    class(TspObsType), intent(inout) :: this
    ! -- local
    integer(I4B) :: i
    character(len=LENVARNAME) :: obs_str
    character(len=LINELENGTH) :: ermsg
    character(len=*), parameter :: fmterrmsg = &
      &"(4x, 'Obs type:',x,a,x,'does not match model type:',x,a)" 
    !
    ! -- Ensure that obs type matches model type
    do i = 1, this%npakobs
      obs_str = trim(adjustl(this%pakobs(i)%obsrv%obstypeid))
      call upcase(obs_str)
      if (obs_str /= 'FLOW-JA-FACE') then
        if (obs_str /= this%depvartype) then
          ! -- Report mismatch in observation types
          write(ermsg, fmterrmsg) obs_str, &
                                  trim(adjustl(this%filtyp))
          call store_error(ermsg)
        end if
      end if
    end do
    !
    ! -- Return
    return
  end subroutine check_obs_id
```
However, I think the current offering is a bit sleeker with less code-bloat.  I discovered there is already some error checking built in which the slightly-expanded new autotest (`test_gwe_obs.py`) now leverages.  To make use of the existing error checking, the call to StoreObsType (line 100-ish in `tsp-obs.f90`) simply needed the _dynamic_ dependent variable name passed in rather than  a hard-coded name (`concentration` or `temperature`)

PR #1962, opened 2024-07-25
fix(bmi-ems): ExplicitModelSolution for PRT fixed for XMI/API calls

langevin-usgs:
During development of the particle tracking (PRT) and surface water flow (SWF) models, a new `ExplicitSolutionType` was added to mf6 for models that do not need to be solved as part of a matrix solution.  Instead, the `ExplicitSolutionType` simply asks each model assigned to it to solve itself.  Prior to the changes in the PR, the bmi-xmi extension could only access `NumericalSolutionType` objects, and so PRT did not work with the XMI-level calls through the API.

This PR changes the `getSolution()` function in `mf6bmiUtil.f90` to return a parent `BaseSolutionType` object, rather than the concrete `NumericalSolutionType` object which it did previously.  This means that `getSolution` now returns either a numerical or explicit solution.  To get this to work properly, the `BaseSolutionType` was supplemented with deferred interfaces to `prepareSolve`, `solve`, and `finalizeSolve`.  One could make an argument that the `BaseSolutionType` should not have any reference to prepareSolve, solve, and finalizeSolve, but since they are foundational for XMI, it seems justified in this case.

With this PR, the [modflowapi](https://github.com/MODFLOW-USGS/modflowapi) should be able to be updated to work with PRT and any other explicit models/solutions that we add in the future.

### DIFF
--- a/autotest/test_prt_libmf6_budget.py
+++ b/autotest/test_prt_libmf6_budget.py
@@ -1,0 +1,109 @@
+"""
+This test runs the simulations in test_prt_budget.py, but uses the api
+to run the PRT test.
+"""
+
+from pathlib import Path
+
+import flopy
+import matplotlib.cm as cm
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import pytest
+from framework import TestFramework
+from modflowapi import ModflowApi
+
+simname = "prt_libmf6"
+cases = [simname]
+
+from test_prt_budget import (HorizontalCase, build_prt_sim, build_mp7_sim, 
+                            check_output)
+
+
+def build_models(idx, test):
+    # build MODFLOW 6 files
+    ws = test.workspace
+    name = cases[idx]
+
+    gwf_sim = HorizontalCase.get_gwf_sim(
+        test.name, test.workspace, test.targets["mf6"]
+    )
+    prt_sim = build_prt_sim(
+        test.name, test.workspace, test.workspace / "prt", test.targets["libmf6"]
+    )
+    mp7_sim = build_mp7_sim(
+        test.name,
+        test.workspace / "mp7",
+        test.targets["mp7"],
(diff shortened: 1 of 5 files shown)


mjr-deltares:
This is a very nice extension!

autotest/test_prt_libmf6_budget.py
```diff
@@ -0,0 +1,110 @@
...
+    # model time loop
+    idx = 0
+    while current_time < end_time:
+        # get dt and prepare for non-linear iterations
+        dt = mf6.get_time_step()
+        mf6.prepare_time_step(dt)
+
+        # convergence loop
+        kiter = 0
+        mf6.prepare_solve()
```
mjr-deltares, line 79: The reason to have prepare_solve, solve, finalize_solve is so you can call solve multiple times and manipulate in between. Is that something you would want to support with this Explicit Solution? And if so, should it maybe also be a test? In this particular test I guess you could change everything with a call to update.

mjr-deltares (approved):
Just that one comment about the test case, but that should not hold this PR back

autotest/test_prt_libmf6_budget.py
langevin-usgs: Very good point.  This is a work in progress, and we are not quite ready to be able to call solve multiple times yet for PRT.  Your comment did also point out that PRT won't work quite right with ATS yet (particles wouldn't get reset if a time step needs to be retried with a shorter length and output is written during the solve).  This has triggered  discussions for how to support PRT/ATS and PRT with multiple solves.  For now, this new support for the explicit solution makes it possible for modflowapi to run any of the PRT models (without any manipulation between the different solve steps).  The current version of modflowapi run_model always uses the lowest level XMI calls (preparesolve, solve, finalizesolve) instead of preparetimestep, dotimestep, finalizetimestep or or update.

PR #1966, opened 2024-07-29
ci: add jobs to build/test with gfortran in debug mode

wpbonelli:
Add jobs to the gfortran test matrix in `ci.yml` to build and test in debug mode. The intent is to catch situations like the one fixed by #1959. To avoid too much CI burden this PR only adds `macos-14` and `ubuntu-22.04` jobs. This seems to add ~15min to total runtime.

___

- [x] Referenced issue or pull request #1959

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,7 +265,7 @@ jobs:
         working-directory: modflow6
         run: |
           setupargs=""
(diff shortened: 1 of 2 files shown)


langevin-usgs (approved):
Looks good, @wpbonelli.  I think this will be another nice safeguard.

PR #1980, opened 2024-08-06
ci: reclaim disk space on ubuntu for intel test job

wpbonelli:
* The Intel test job is running out of disk space on Ubuntu. Delete some unnecessary things baked into the runner image.
* Copy `libmf6.so` with the mf6 executable into `~/.local/bin/modflow` before running example tests, previously we neglected to do this so the shared object installed from the executables distribution was being used.
* Skip `ex-gwe-radial` when running example tests with Intel Fortran for now, pending reduction in memory use. See comment below.

---


wpbonelli:
I think there may be something else going on since tests are now failing consistently here: https://github.com/MODFLOW-USGS/modflow6/actions/runs/10290838684/job/28482332549?pr=1980#step:17:48

Next I will try walking back the pixi update

wpbonelli:
reverting the pixi update changes nothing, the failure occurs in the same place

wpbonelli:
The problem was that the development build of `libmf6.so` was not copied into `~/.local/bin/modflow` with the mf6 executable before running the example tests.

wpbonelli:
We have more failures in the example tests. Given the 143 exit code, I am starting to think these are related to CPU/memory use as reported by many other projects in https://github.com/actions/runner-images/issues/6680.

I will try limiting xdist to fewer cores than are available.

wpbonelli:
Skipping `ex-gwe-radial` avoids the failure. This example is very memory intensive &mdash; this seems to be the cause of the crash.

PR #1993, opened 2024-08-15
fix(mf6bmiutil): workaround ARM mac char array assignment issue

wpbonelli:
Assignments from the result of `string_to_char_array` fail on ARM mac with GNU Fortran when the LHS and RHS are of different lengths. This seems like a compiler bug? Confirmed with gfortran 12-14. Work around it by introducing a subroutine to do the conversion.

@mjr-deltares and I have discussed and determined as a next step to create a minimal reproduction and post on the Fortran discourse forum. I will add a link here once that's done.

This should resolve the modflowapi macOS CI failures.

___

### DIFF
--- a/srcbmi/mf6bmi.f90
+++ b/srcbmi/mf6bmi.f90
@@ -48,7 +48,7 @@ contains
     integer(kind=c_int) :: bmi_status !< BMI status code
     ! -- local variables
 
-    name = string_to_char_array('MODFLOW 6', 9)
+    call string_to_char_array_2('MODFLOW 6', 9, name)
     bmi_status = BMI_SUCCESS
 
   end function bmi_get_component_name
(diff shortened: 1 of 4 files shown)


scharlton2:
@wpbonelli Did you guys try:

changing line 343 from:
character(kind=c_char), intent(out) :: mf_version(BMI_LENVERSION)
to
character(kind=c_char), intent(inout) :: mf_version(BMI_LENVERSION)

wpbonelli:
@scharlton2 that works! Why does this work? Should these have been `inout` all along?

scharlton2:
@wpbonelli I don't know for sure but it made sense to me as a C++ developer.

wpbonelli:
Ok. Unless anyone else has objections I'll put this solution in instead of the subroutine hack.

Thanks @scharlton2

langevin-usgs:
Sounds good to me.

PR #1493, opened 2023-12-09
feat(GWE): Introduce Groundwater Energy Transport

emorway-usgs:
- Adopting a more staged approach compared to #1237 (which will eventually be closed once this PR is adopted)
- Does not include the advanced transport packages, nor energy source loading, those will come in stages to follow
- Currently includes 6 autotests as part of this PR, more to follow so long as the code compiles.
- Documentation (.tex) builds.
- Ran black
- Ran fprettify
- If this PR eventually passes and is incorporated, its adoption will spawn a couple of new "maintenance" issues.  Currently, there is a non-trivial amount of duplicated code in:
    - ./src/Model/Connection/GweGweConnection.f90 (compare with GwtGwtConnection.f90)
    - ./src/Model/Connection/GweInterface.f90 (compare with GwtInterface.f90)
    - ./src/Exchange/GweGweExchange.f90 (compare with GwtGwtExchange.f90)
    - ./src/Exchange/GwfGweExchange.f90 (compare with GwfGwtExchange.f90)
    - ./src/Model/Connection/GweGweConnection.f90 (compare with GwtGwtConnection.f90)
    - ./src/Model/Connection/GweInterfaceModel.f90 (compare with GwtInterfaceModel.f90)
    Perhaps there would be a way to generalize what's going on here and reduce duplication?
- This PR trades on the changes that "generalized" a base transport class reflected in the following PRs:
    - #1386
    - #1391
    - #1392
    - #1396
    - #1398
    - #1399
    - #1400
    - #1428
    - #1445

Let the PR fun begin...

PR #1999, opened 2024-08-21
fix(GweInputData): large GWE models were throwing stackoverflow error

emorway-usgs:
The following two lines:
`this%gwerhos = gwerhos`
`this%gwecps = gwecps`

should have used `=>` instead:

`this%gwerhos => gwerhos`
`this%gwecps => gwecps`

That discovery led to a number of other fixes since the variables in question were pointers and not copies of data.  While in `GweInputData.f90`, updated it to follow latest style guide.  Will adjust other GWE src files in the future.

- [x] Referenced issue or pull request #1493


mjr-deltares:
Hi @emorway-usgs , that last commit looks a bit different from what we discussed last week. Maybe have a call so I can explain my thoughts?

mjr-deltares (approved):
Looks good. I will make a note of the validation not being done in parallel but this can go in.

PR #2030, opened 2024-10-07
fix(par): PETSc Fortran API has changed in v3.22.0

mjr-deltares:
From their release notes https://petsc.org/release/changes/322/:

* Change the PETSc Fortran API so that non-array values, v, passed to PETSc routines expecting arrays must be cast with [v] in the calling sequence

Checklist of items for pull request

- [x] Replaced section above with description of pull request
- [x] Formatted new and modified Fortran source files with `fprettify`
- [x] Removed checklist items not relevant to this pull request

PR #2031, opened 2024-10-08
refactor(dfns): add in_record attr to utl-ts and utl-tas vars

wpbonelli:
Add `in_record true` to several record fields in `utl-ts.dfn` and `utl-tas.dfn`. Motivated by https://github.com/modflowpy/flopy/pull/2317. FloPy can infer this, evidently, but a less clever (and ideally much simpler) implementation may require them explicitly. IMO it makes sense to do this anyway for consistency's sake, at least while the DFN specification is flat and record membership cannot be inferred directly from the spec's structure.

This is not necessarily comprehensive, further work may reveal the same (or similar) updates are needed elsewhere.

### DIFF
--- a/doc/mf6io/mf6ivar/dfn/utl-tas.dfn
+++ b/doc/mf6io/mf6ivar/dfn/utl-tas.dfn
@@ -19,6 +19,7 @@ type keyword
 shape
 reader urword
 optional false
+in_record true
 longname
 description xxx
 
@@ -29,6 +30,7 @@ shape any1d
 tagged false
 reader urword
 optional false
+in_record true
 longname
 description Name by which a package references a particular time-array series. The name must be unique among all time-array series used in a package.
 
@@ -48,6 +50,7 @@ type keyword
 shape
 reader urword
 optional false
+in_record true
 longname
 description  xxx
 
@@ -59,6 +62,7 @@ shape
 tagged false
 reader urword
 optional false
+in_record true
 longname
 description Interpolation method, which is either STEPWISE or LINEAR.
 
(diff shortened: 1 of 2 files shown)


wpbonelli:
CI failures are unrelated and should be fixed by #2030

martclanor:
Hi @wpbonelli, I'm curious, is there a reason why `method` and `interpolation_method_single` are excluded here?

wpbonelli:
@martclanor it was probably an oversight.. IIRC, I just did this for consistency's sake. Flopy needed `in_record` on certain variables and could infer it for others. We have a tentative new DFN format which, being hierarchical TOML, has no more need of such an attribute. We are in the process of rewriting everything that uses DFNs (code generation etc) to consume TOML, at which point we will replace the old DFNs, so this will soon be irrelevant.

Are you running into this somehow?

martclanor:
Ah okay, understood. Yes, I noticed this as an edge case while trying to extract the block structure from the DFN files.

But in that case, I'll ignore this for now until the version with the toml is released. Thanks for the quick response!

wpbonelli:
No prob! More info here if you want to give the TOML a spin: https://modflow-devtools.readthedocs.io/en/latest/md/dfn.html

There are a few top-level entries which describe the component (info which was previously in "#"-comment lines at the top of the DFN) but if you exclude those keys, what remains are the blocks.

Let us know if you have any feedback on the new format. I'd be curious to know what you're using DFNs for as we assumed they are mostly internal.

martclanor:
I am using the DFN's for this project: https://github.com/martclanor/vscode-mf6-syntax

Thanks for sharing this! I will try this conversion script then parse the toml instead.

wpbonelli:
Wow. Nice! Thanks for this.

PR #2056, opened 2024-11-12
chore: reformat Python code with line length = 88

mwtoews:
This PR reformats Python code with a maximum line length of 88 characters to better represent the code. Note that a related repo [MODFLOW-USGS/modflow6-examples](https://github.com/MODFLOW-USGS/modflow6-examples) already uses the default line length of 88.

See https://github.com/modflowpy/flopy/pull/2362 for a related PR with rational and discussion.

This PR was automated using these Ruff commands (including quick fixes to [ISC001](https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/)):
```bash
ruff format
ruff check --select ISC001 --fix
ruff format
```

### DIFF
--- a/autotest/build_exes.py
+++ b/autotest/build_exes.py
@@ -23,9 +23,7 @@ def test_meson_build(bin_path):
 
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(
-        "Rebuild local development version of MODFLOW 6"
-    )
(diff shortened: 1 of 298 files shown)


wpbonelli (approved):
I guess we can eventually look into removing ignored rules here too

mwtoews:
A follow-up PR could look at relaxing some ignored rules, as well as adding a `.git-blame-ignore-revs` file.

PR #2060, opened 2024-11-17
chore: add codespell configuration, check python source files

mwtoews:
This PR has two primary aims:

1. Add a codespell configuration, so that users can type `codespell` to check everything and show suggested changes. This takes 1.5 seconds on my laptop.
2. `python3 .github/common/check_spelling.py` now checks Python source files. This takes 12.5 seconds on the same laptop.

Other changes:

- Fix a small handful of typos from Python source files.
- Expand the list of ignored words. Some case-sensitive ignores are enforced, others that don't raise errors are removed.
- Improve use of [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run) usage in  `check_spelling.py`

### DIFF
--- a/.codespell.ignore
+++ b/.codespell.ignore
@@ -1,4 +1,4 @@
-alph
+Alph
 wel
 nam
 gage
@@ -7,10 +7,11 @@ delt
 lke
 ist
 inout
+fo
+nd
 ot
-initialx
-initialy
-initialz
+initialY
+localy
 iterm
 dum
 shft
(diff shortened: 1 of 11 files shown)


wpbonelli (approved):
Is there any reason for `check_spelling.py` now if running codespell directly covers all the same cases and is faster?

I think I just copy/pasted this script from `check_format.py` which wraps fprettify. IIRC we had to work around some things like exclude patterns not working. but it would be nice to get rid of the scripts if possible

mwtoews:
In my opinion, `check_spelling.py` doesn't seem to be needed anymore.

PR #2062, opened 2024-11-17
refactor: resolve ruff check F821 for undefined name

mwtoews:
This is a refactor to resolve [F821](https://docs.astral.sh/ruff/rules/undefined-name/) for undefined name.


mwtoews:
There is one outstanding issue with `autotest/test_gwf_lak_wetlakbedarea01.py` that I cannot fix. It was introduced with e81061d64c02d191bc1e29c994aaf293111a52c5. Does @emorway-usgs  have any suggestions?

emorway-usgs (approved):
Not sure what my thinking was in the `test_gwf_lak_wetlakbedarea01.py` script. The variable `b` cannot be negative (which would trigger execution of the `else`) because cell tops can never go below cell bottoms.  The `else` statement was deleted.  After deleting the questionable python, resolved a merge conflict.  CI running now - sorry I didn't see your request to review this PR sooner, @mwtoews

Issue #2048, opened 2024-11-06
Add check for `AUXMULTNAME` option in `UZF` to `GWE`


emorway-usgs:
Recent edits to the draft GWE manuscript highlighted the need for a new requirement in `GWE`.  Specifically, when UZE is in use, the model needs to ensure that the `AUXMULTNAME` option in UZF is not activated.  Multiple UZF/UZE objects within a given grid cell is a potential problem for GWE.

emorway-usgs:
Addressed with #2073

PR #2073, opened 2024-11-26
fix(gwe-uze): stop simulation when area of uze object not equal to cell area

emorway-usgs:
Based of how UZE is formulated, only one UZE object per grid cell is permitted.  If more than one UZE object is discovered in a grid cell, GWE will now exit with an error message.  To fix, users will need to remove the AUXMULTNAME keyword from the OPTIONS block of the corresponding UZF input file.

- [x] Fixes issue #2048 
- [x] Added new test
- [x] Ran `ruff` on new python script in autotest subdirectories.
- [x] Added doxygen comments to new procedures


emorway-usgs:
review comments included: 
- reducing extent of `IDOMAIN` such that reduced node numbers were different then absolute node number
- ensure that the new `dup_chk` vector is sized to `NNODES` and not merely `MAXLIST`
- rename subroutines
- supplement error message with ID of cell where more than one UZF object exists

emorway-usgs:
@langevin-usgs, some more work remains on this.  The auxmultcol-adjusted area of each UZF object isn't set until the _ad() routines.  So the current check (as of 52e6d7e) does not do the area comparison correctly.

src/Model/GroundWaterEnergy/gwe-uze.f90
```diff
@@ -265,6 +270,19 @@ subroutine find_uze_package(this)
     this%idxbudtheq = this%flowbudptr%nbudterm + 1
   end subroutine find_uze_package
 
+  !> @brief Check to ensure auxiliary areas equal respective cell areas
+  !<
+  subroutine uze_chk_aux_area(this)
```
langevin-usgs, line 275: Do we still need uze_chk_aux_area and uzarea_chk?  I'd be inclined to condense these into one routine called uze_ad_chk.

src/Model/GroundWaterEnergy/gwe-uze.f90
```diff
@@ -1293,6 +1311,61 @@ subroutine uze_bd_obs(this, obstypeid, jj, v, found)
...
+
+    ! cycle through uze objects, stop at first occurrence of more than one
+    ! uze object in a cell
+    do n = 1, nuz
+      igwfnode = this%flowbudptr%budterm(idxbudgwf)%id2(n)
+      carea = this%dis%area(igwfnode)
+      uzarea = this%flowbudptr%budterm(idxbudgwf)%auxvar(1, n)
+      ! compare areas
+      if (.not. is_close(carea, uzarea)) then
+        call this%print_uz_err(igwfnode)
```
langevin-usgs, line 1343: Instead of calling `store_error` and `StoreErrorUnit` from this routine, let's keep submitting all the errors, and then after this loop, call `count_errors` and then call `store_error` and `StoreErrorUnit` if any errors were encountered.  Maybe also rename `print_uz_err` to `area_error(igwnode)`.

src/Model/GroundWaterEnergy/gwe-uze.f90
```diff
@@ -156,10 +159,10 @@ subroutine find_uze_package(this)
     ! -- dummy
     class(GweUzeType) :: this
     ! -- local
-    character(len=LINELENGTH) :: errmsg
     class(BndType), pointer :: packobj
     integer(I4B) :: ip, icount
     integer(I4B) :: nbudterm
+    character(len=LINELENGTH) :: errmsg
```
langevin-usgs, line 165: I think the pattern is to import errmsg as `use SimVariablesModule, only: errmsg` at the top so it can be reused throughout the module.

langevin-usgs (approved):
Looks much better.  Did this not work if the check was done at the end of rp?  Just asking because this will require a loop through all nodes at the start of every time step (instead of stress period).  Probably not a big deal.  Other minor suggestions.

src/Model/GroundWaterEnergy/gwe-uze.f90
emorway-usgs: Addressed in c25a3a2

src/Model/GroundWaterEnergy/gwe-uze.f90
emorway-usgs: thanks for catching this, addressed in 189d4e6

src/Model/GroundWaterEnergy/gwe-uze.f90
emorway-usgs: Consolidation is always better, thanks for catching that.  addressed in 1ebcad4

emorway-usgs:
> Did this not work if the check was done at the end of rp? Just asking because this will require a loop through all nodes at the start of every time step (instead of stress period). 

The check doesn't fail at the end of `_rp()` like it should. It appears that the auxiliary variable containing the UZE area hasn't been updated by the end of `_rp()`.  However, it has been updated by the time `_ad()` routines have been called and the new test problems fail as designed.

langevin-usgs (approved):

PR #2087, opened 2024-12-03
ci(plot): add plot option to conftest

langevin-usgs:
Many of the autotest scripts have code for plotting results.  This PR establishes a convention for passing in a --plot command line option to pytest, which will cause a `plot_output()` function to be called.  

### DIFF
--- a/autotest/conftest.py
+++ b/autotest/conftest.py
@@ -100,6 +100,11 @@ def original_regression(request) -> bool:
     return request.config.getoption("--original-regression")
 
 
+@pytest.fixture
+def plot(request) -> bool:
+    return request.config.getoption("--plot")
+
+
 @pytest.fixture(scope="session")
 def markers(pytestconfig) -> str:
     return pytestconfig.getoption("-m")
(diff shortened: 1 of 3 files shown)


wpbonelli (approved):
I remember now that the examples are where we've done this most extensively https://github.com/MODFLOW-USGS/modflow6-examples/blob/develop/autotest/conftest.py

maybe these could be provided by devtools, at least write/run/plot which are generally applicable.

langevin-usgs:
Good idea to look into supporting the full suite of write/run/plot.  It would be nice to have full control over these.  For example, it may be nice to manually modify an input file and then be able to check output and plot.  A trick for the modflow6 repo will be how to handle the paths, as the models are written/tested in a pytest temp directory.

mjr-deltares:
This is a nice development, should we at some point ask people to roll this out into 'their' tests, I certainly would like to do that.

wpbonelli:
@mjr-deltares agreed, yesterday we discussed encouraging people to implement this pattern in their own tests

@langevin-usgs we currently have the `--keep` option to save a temp dir in a user-specified location &mdash; to support the use case you describe, maybe we could consider a `--load` option which instead of writing the simulation, tries to load the simulation from a specified location?

PR #2123, opened 2025-01-01
refactor(tsp-apt): remove gwe-specific code from apt and distribute to advanced gwe packages

emorway-usgs:
The variables `ktf`, thermal conductivity of the of the interface between the aquifer cell and the apt feature, for example a stream reach, and `rfeatthk`, a real value that defines the thickness of the streambed material through which conduction occurs, are specific to `GWE` and shouldn't have been added to `APT` in the first place.  This PR removes both variables from APT and distributes them to `SFE`, `LKE`, and `MWE` where they are needed.  Three pre-existing GWE autotests were already in place to help ensure that the changes didn't break anything (`test_gwe_sfe_strmbedcond.py`, `test_gwe_lke_conduction.py`, and `test_gwe_mwe_conduction.py`).

- [x] Relates to PR #1493
- [x] Fixed a typo in the SFE And LKE [definition files](/MODFLOW-USGS/modflow6/tree/develop/doc/mf6io/mf6ivar)

### DIFF
--- a/doc/mf6io/mf6ivar/dfn/gwe-lke.dfn
+++ b/doc/mf6io/mf6ivar/dfn/gwe-lke.dfn
@@ -294,7 +294,7 @@ tagged false
 in_record true
 reader urword
 longname boundary thermal conductivity
-description is the thermal conductivity of the interface between the aquifer cell and the lake.  
+description is the thermal conductivity of the material between the aquifer cell and the lake. The thickness of the material is defined by the variable RBTHCND.
 
 block packagedata
 name rbthcnd
--- a/doc/mf6io/mf6ivar/dfn/gwe-lke.dfn
+++ b/doc/mf6io/mf6ivar/dfn/gwe-lke.dfn
@@ -294,7 +294,7 @@ tagged false
 in_record true
 reader urword
 longname boundary thermal conductivity
-description is the thermal conductivity of the of the interface between the aquifer cell and the lake.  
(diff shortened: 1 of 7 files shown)


doc/mf6io/mf6ivar/dfn/gwe-lke.dfn
```diff
@@ -294,7 +294,7 @@ tagged false
 in_record true
 reader urword
 longname boundary thermal conductivity
-description is the thermal conductivity of the of the interface between the aquifer cell and the lake.  
+description is the thermal conductivity of the interface between the aquifer cell and the lake.
```
langevin-usgs, line 297: I wonder if we can think of something other than interface (material?) for this description.  There is an associated width of this material, correct?  How is that width determined?  Is it a half cell thickness in either the vertical or horizontal direction depending on the connection type?

langevin-usgs (commented):
Nice work getting these variables out of the tsp model.  Perhaps at some point we can come up with a way to remove all this duplicate code, but for now I think this is good.  I had one comment about the term "interface".  Wonder if we need a better term for this as there should be a thickness to this interface that we might want to mention somewhere.

doc/mf6io/mf6ivar/dfn/gwe-lke.dfn
emorway-usgs: In response, tweaked the explanations with f914f2b

PR #2149, opened 2025-01-17
fix(exg-gwf*): check for equal idomain 

wpbonelli:
Check that models connected to a flow model by an exchange have an idomain identical to the flow model's idomain. This is an easy error to make, and (up to now) hard to detect.

I only added a test for PRT for now. The logic is identical in GWT and GWE. I wondered if there are suitable autotests for those models to add a test case to? Or maybe this behavior should have its own autotest?

#2144 facilitates adding the same check for flow model interface. I figure that can be done in a separate PR since it probably deserves a separate release note?

___

### DIFF
--- a/src/Exchange/exg-gwfgwe.f90
+++ b/src/Exchange/exg-gwfgwe.f90
@@ -181,6 +181,9 @@ contains
   subroutine exg_ar(this)
     ! -- modules
     use MemoryManagerModule, only: mem_checkin
+    use DisModule, only: DisType
+    use DisvModule, only: DisvType
+    use DisuModule, only: DisuType
     ! -- dummy
     class(GwfGweExchangeType) :: this
     ! -- local
@@ -193,6 +196,11 @@ contains
       & ',a,'.&
       &  GWF Model has ', i0, ' user nodes and ', i0, ' reduced nodes.&
       &  GWE Model has ', i0, ' user nodes and ', i0, ' reduced nodes.&
+      &  Ensure discretization packages, including IDOMAIN, are identical.')"
+    character(len=*), parameter :: fmtidomerr = &
+      "('GWF and GWE Models do not have the same discretization for exchange&
+      & ',a,'.&
+      &  GWF Model and GWE Model have different IDOMAIN arrays.&
       &  Ensure discretization packages, including IDOMAIN, are identical.')"
     !
     ! -- set gwfmodel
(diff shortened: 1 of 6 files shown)


src/Exchange/exg-gwfgwe.f90
```diff
@@ -220,6 +228,34 @@ subroutine exg_ar(this)
       call store_error(errmsg, terminate=.TRUE.)
     end if
     !
+    ! -- Make sure idomains are identical
+    select type (gwfdis => gwfmodel%dis)
```
wpbonelli, line 232: All the `select type` statements are unfortunate, but idomain is defined separately for each dis type since it's the real grid shape instead of flat

langevin-usgs: Ahh, bummer.  I guess now that all discretizations have IDOMAIN, we could move it to disbasetype and make it a 1D int vector, but that is for another time, I think.

langevin-usgs (commented):
Looks good.  It's a convoluted check, as you mention, but perhaps a refactor of idomain into disbasetype can be done at some point to clean this up.

langevin-usgs (approved):

src/Exchange/exg-gwfgwe.f90
mjr-deltares: If I am understanding correctly, we are trying to assess the equality of grid. Maybe we can have a `function is_equal(this, other_dis) result(are_equal)` to delegate the check to the particular Discretization module?

mjr-deltares: PS: happy to help and chat more about that option

wpbonelli: @mjr-deltares thanks we should definitely chat on this with @langevin-usgs 

It may be relevant to #2153 as well. In that PR, I first planned to load the full flow model dis into fmi, then an `is_equal` function would be quite natural to use. But the dis types have some strong expectations about how they are initialized and where they get their data from. So in the near term, I was thinking of loading only the minimum necessary info to do the same checks. Hopefully we can come up with a nicer long term solution.

langevin-usgs: This is a good point.  In this new check, we are saying the grids are the same if the number of nodes and idomain is the same; it is easy to accidentally leave out idomain when using flopy to construct dis.  But certainly, if tops and bottoms were also different, then that would not be detected here, and problems would be very difficult to find.  A discretization-based `is_equal()` (and maybe `is_same()`?) seems like a nice way to ensure grid equality.

mjr-deltares: OK, let's talk tomorrow and plan a meeting. Btw I had no intention stalling this PR. This should probably go in and then we refactor when we have a plan.

PR #2151, opened 2025-01-18
docs(prt-fmi): clarify only required when post-processing

wpbonelli:
The first paragraph in the prt fmi section could be read to indicate that the fmi file is always required. Add some clarification. Also update a test to make sure the head/budget file paths provided via fmi are ignored when the simulation uses an exchange

### DIFF
--- a/autotest/test_prt_exg.py
+++ b/autotest/test_prt_exg.py
@@ -108,15 +108,17 @@ def build_mf6_sim(idx, test):
         trackcsv_filerecord=[prt_track_csv_file],
     )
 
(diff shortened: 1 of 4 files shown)


langevin-usgs:
This is a nice clarification. Should we add this same description to the GWT and GWE fmi descriptions also?

wpbonelli:
> Should we add this same description to the GWT and GWE fmi descriptions also?

Added to GWT and GWE

langevin-usgs (approved):

PR #2153, opened 2025-01-20
feat(fmi): support binary grid file

wpbonelli:
Support an optional `GWFGRID` entry in FMI for the flow model's binary grid file. If provided, check the grids have the same size and idomain like we do in EXG as of #2149. This makes for the same error-checking in the exg (coupled) and fmi (post-processing) cases.

@langevin-usgs suggested making the grid file entry optional for some time (1-2 releases), then beginning to require it. At which point the reader could be modified to load the full dis, and we could consolidate these haphazard checks into a robust grid comparison as suggested by @mjr-deltares.

The grb file reader in zonebudget is moved to a new `GridFileReader` module in the main src dir. The reader now loads headers and builds an index at init time, after which variables can be read as necessary.

___


doc/ReleaseNotes/develop.tex
```diff
@@ -20,7 +20,8 @@
 \textbf{\underline{BUG FIXES AND OTHER CHANGES TO EXISTING FUNCTIONALITY}} \\
 \underline{BASIC FUNCTIONALITY}
 \begin{itemize}
-	\item Previously, GWF Model exchanges with other model types (GWT, GWE, PRT) verified that the flow model and the coupled model had the same number of active nodes, but did not check that the models' IDOMAIN arrays selected precisely the same set. Exchanges will now check that model IDOMAIN arrays are identical.
+	\item GWT, GWE and PRT Models require a grid identical to the grid of the corresponding GWF Model. Previously, GWT, GWE and PRT exchanges performed a minimal check that the grids have the same number of active nodes. Exchanges will now also perform an additional check that grid IDOMAIN arrays are identical.
```
wpbonelli, line 23: I tweaked the note from #2149 since the motivation is the same

src/Utilities/GridFileReader.f90
```diff
@@ -0,0 +1,521 @@
...
+        end do
+        ishp = size(this%shp)
+        call ExpandArray(this%shp, increment=ndim)
+        this%shp(ishp + 1:ishp + ndim) = shp
+        call this%shp_idx%add(key, ishp + 1)
+      end if
+    end do
+  end subroutine read_header_body
+
+  !> @brief Skim the file and build an index. Internal use only.
```
wpbonelli, line 204: I guess to get the offsets, this could just multiply shape by byte count instead of reading the file.

src/Utilities/GridFileReader.f90
```diff
@@ -0,0 +1,521 @@
...
+      v(2) = this%read_int("NROW")
+      v(3) = this%read_int("NCOL")
+    case ("DISV")
+      allocate (v(2))
+      v(1) = this%read_int("NLAY")
+      v(2) = this%read_int("NCPL")
+    case ("DISU")
+      allocate (v(1))
+      v(1) = this%read_int("NODES")
+    end select
```
wpbonelli, line 327: need to support the other grid types too

src/Utilities/GridFileReader.f90
```diff
@@ -0,0 +1,521 @@
...
+      call urword(line, lloc, istart, istop, 1, ival, rval, 0, 0)
+      key = line(istart:istop)
+      call upcase(key)
+      nvars = nvars + 1
+      this%keys(nvars) = key
+
+      ! type
+      call urword(line, lloc, istart, istop, 1, ival, rval, 0, 0)
+      dtype = line(istart:istop)
+      call upcase(dtype)
```
langevin-usgs: fyi -- urword has an argument (ncode) that will convert to upper case, so the call here to upcase is not necessary.

langevin-usgs (approved):
Looks good.

Issue #2142, opened 2025-01-16
Missing zero-order decay for the solid phase in GWE


cneyens:
Chapter 10 of the Supplemental Technical Information for MODFLOW 6 (v6.6.0) describes the Groundwater Energy Model. Equations 10-1 to 10-3 show that zero-order decay for both the solid and aqueous phase can be simulated. This is also explicitly stated in the accompanying text. In the IO guide however, only aqueous decay is described, which is indeed only used for decay in the aqueous phase and not both the aqueous and solid phase (as can be be verified by comparing to an analytical solution). 

Is the solid-phase zero-order decay input missing from the input guide or is this functionality not implemented yet (in contrast to what's described in the Supplemental guide)?

emorway-usgs:
Thanks for bringing this to our attention, @cneyens.  We're taking a closer look.  When you say, "as can be verified by comparing to an analytical solution," did you set up a test?

cneyens:
I made a single-cell GWE model with only zero-order production as heat source (see attached notebook).

[gwe_decay.zip](https://github.com/user-attachments/files/18441323/gwe_decay.zip)

emorway-usgs:
Working on fix, @cneyens. Thanks for the example, it'll definitely help ensure the code is working properly.

emorway-usgs:
Thanks again for pointing out this issue, @cneyens, we appreciate your patience as we worked on the fix in #2155.  The next [nightly build](https://github.com/MODFLOW-USGS/modflow6-nightly-build/releases/) should be good to go.

PR #2155, opened 2025-01-21
fix(gwe-est): add support for energy decay in the solid phase

emorway-usgs:
Support for specifying energy decay in the solid phase was missing despite it being described in the Supplemental Technical Information document.  The best I can figure is that during development it was initially skipped over with the intention of coming back to it - but then was forgotten about. 

- [x] Addresses issue #2142 
- [x] Functionality should have been included with pull request #1493 

### DIFF
--- a/autotest/test_gwe_decay01.py
+++ b/autotest/test_gwe_decay01.py
@@ -128,6 +128,13 @@ def build_models(idx, test):
 
 def check_output(idx, test):
     print("evaluating results...")
+    msg = (
+        "Differences detected between the simulated results for zeroth-order "
+        "energy decay and the expected solution for decay specified in "
+    )
+    msg0 = msg + "the aqueous phase."
+    msg1 = msg + "the solid phase."
+    msg2 = msg + "both the aqueous and solid phases."
 
     # read transport results from GWE model
     name = cases[idx]
@@ -152,13 +159,13 @@ def check_output(idx, test):
     print("temperature evaluation: " + str(temp_analy_w))
 
     if "aqe" in name:
-        assert temp_ts[-1, 1] == temp_analy_w[-1]
+        assert np.isclose(temp_ts[-1, 1], temp_analy_w[-1], atol=1e-10), msg0
 
     if "sld" in name:
-        assert temp_ts[-1, 1] == temp_analy_s[-1]
+        assert np.isclose(temp_ts[-1, 1], temp_analy_s[-1], atol=1e-10), msg1
 
     if "both" in name:
-        assert temp_ts[-1, 1] == temp_analy_ws[-1]
+        assert np.isclose(temp_ts[-1, 1], temp_analy_ws[-1], atol=1e-10), msg2
 
 
 # - No need to change any code below
--- a/autotest/test_gwe_decay01.py
+++ b/autotest/test_gwe_decay01.py
@@ -1,0 +1,177 @@
+"""
+Test problem for decay of energy in EST package of GWE.  Uses a single-cell
+model.  Test contributed by Cas Neyens.
+"""
+
+# Imports
+
+import flopy
+import numpy as np
+import pytest
+from framework import TestFramework
+
+# Base simulation and model name and workspace
+
+cases = ["decay-aqe", "decay-sld", "decay-both"]
+
+# Model units
+length_units = "meters"
+time_units = "seconds"
+
+rho_w = 1000  # kg/m^3
+rho_s = 2500  # kg/m^3
+n = 0.2  # -
+gamma_w = -1000  # J/s/m^3, arbitrary value for zero-order aqueous heat production
+gamma_s = -0.1  # J/s/kg, arbitrary value for zero-order solid heat production
+c_w = 4000  # J/kg/degC
+c_s = 1000  # J/kg/decC
+T0 = 0  # degC
+
+nrow = 1
+ncol = 1
+nlay = 1
+delr = 1  # m
+delc = 1  # m
+top = 1  # m
+botm = 0  # m
+
+perlen = 86400  # s
+nstp = 20
+
+parameters = {
+    # aqueous
+    "decay-aqe": {
+        "zero_order_decay_water": True,
+        "zero_order_decay_solid": False,
+        "decay_water": gamma_w,
+        "decay_solid": gamma_s,
+    },
+    # solid
+    "decay-sld": {
+        "zero_order_decay_water": False,
+        "zero_order_decay_solid": True,
+        "decay_water": gamma_w,
+        "decay_solid": gamma_s,
+    },
+    # combined
+    "decay-both": {
+        "zero_order_decay_water": True,
+        "zero_order_decay_solid": True,
+        "decay_water": gamma_w,
+        "decay_solid": gamma_s,
+    },
+}
+
+
+def temp_z_decay(t, rho_w, rho_s, c_w, c_s, gamma_w, gamma_s, n, T0):
+    t = np.atleast_1d(t)
+    coeff = (-gamma_w * n - gamma_s * (1 - n) * rho_s) / (
+        rho_w * c_w * n + rho_s * c_s * (1 - n)
+    )
+    return coeff * t + T0
+
+
+def build_models(idx, test):
+    # Base MF6 GWF model type
+    ws = test.workspace
+    name = cases[idx]
+    gwename = "gwe-" + name
+
+    decay_kwargs = parameters[name]
+
+    print(f"Building MF6 model...{name}")
+
+    sim = flopy.mf6.MFSimulation(
+        sim_name="heat",
+        sim_ws=ws,
+        exe_name="mf6",
+        version="mf6",
+    )
+    tdis = flopy.mf6.ModflowTdis(sim, nper=1, perioddata=[(perlen, nstp, 1.0)])
+    ims = flopy.mf6.ModflowIms(
+        sim, complexity="SIMPLE", inner_dvclose=0.001
+    )  # T can not become negative in this model
+
+    gwe = flopy.mf6.ModflowGwe(
+        sim,
+        modelname=gwename,
+        save_flows=True,
+        model_nam_file=f"{gwename}.nam",
+    )
+    dis = flopy.mf6.ModflowGwedis(
+        gwe, nrow=nrow, ncol=ncol, nlay=nlay, delr=delr, delc=delc, top=top, botm=botm
+    )
+    ic = flopy.mf6.ModflowGweic(gwe, strt=T0)
+    est = flopy.mf6.ModflowGweest(
+        gwe,
+        zero_order_decay_water=decay_kwargs["zero_order_decay_water"],
+        zero_order_decay_solid=decay_kwargs["zero_order_decay_solid"],
+        density_water=rho_w,
+        density_solid=rho_s,
+        heat_capacity_water=c_w,
+        heat_capacity_solid=c_s,
+        porosity=n,
+        decay_water=decay_kwargs["decay_water"],
+        decay_solid=decay_kwargs["decay_solid"],
+    )
+
+    oc = flopy.mf6.ModflowGweoc(
+        gwe,
+        budget_filerecord=f"{gwe.name}.bud",
+        temperature_filerecord=f"{gwe.name}.ucn",
+        printrecord=[("BUDGET", "ALL"), ("TEMPERATURE", "ALL")],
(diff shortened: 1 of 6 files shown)


autotest/test_gwe_decay01.py
aprovost-usgs, line 1: A nice cross-check might be to instead use ESL to add/remove heat to/from the model at the same total rate as in the current setup and make sure you get the same results for temperature.  (Addition/removal of an amount of heat from the aquifer should be the same as production/decay of the same amount of heat distributed between the water and solid because the water and solid equilibrate anyway)

doc/mf6io/mf6ivar/dfn/gwe-est.dfn
aprovost-usgs, line 1: Is it worth clarifying that the decay in each phase is a rate per amount of that phase (not per volume of aquifer), or is that obvious enough?  A similar question applies to straight-up "decay" (sorption aside) in GWT, which is only in the water and is (presumably?) a rate per volume of water (not per volume of aquifer)

src/Model/GroundWaterEnergy/gwe-est.f90
aprovost-usgs, line 1: Is it worth renaming subroutine est_fc_dcy to est_fc_dcy_water (similar for est_cq_dcy)?

aprovost-usgs, line 1: In est_cq_dcy, remove part of comment that says solid still needs to be added

aprovost-usgs, line 1: Remove temporary comment notes about first-order decay from est_fc_dcy, est_cq_dcy, and est_cq_dcy_solid.  If desired, could replace them with formal comments saying that first-order decay of energy is not supported.  Also looks like there's a reference to first-order decay in read_options

aprovost-usgs, line 1: Remove temporary comment note about negative temps from est_fc_dcy.  If desired, could replace it with a formal comment about the fact that negative temps are currently not checked for or prevented

doc/mf6io/mf6ivar/dfn/gwe-est.dfn
aprovost-usgs: "The dimensions of decay for zero-order decay in the aqueous phase is energy per length cubed per time." -- Suggest modifying to "The dimensions of zero-order decay in the aqueous phase are energy per length cubed per time." 
 Similar suggestion for the analogous description of decay_solid.

Related question: based on the code and the test problem, the units of decay in the solid are energy per mass (of solid) per time, not "energy per length cubed per time"

src/Model/GroundWaterEnergy/gwe-est.f90
aprovost-usgs: Suggest modifying
      if (.not. lname(2) .and. .not. lname(3)) then
        write (errmsg, '(a)') 'Zero order decay in either the aqueous &
          &or solid phase is active but zero-order rate coefficients, &
          &either in the aqueous or solid phase, is not specified.  &
          &Either DECAY_WATER or DECAY_SOLID must be specified in the &
          &griddata block.'
to this
      if (.not. lname(2) .and. .not. lname(3)) then
        write (errmsg, '(a)') 'Zero order decay in either the aqueous &
          &or solid phase is active but the corresponding zero-order &
          &rate coefficient is not specified. Either DECAY_WATER or &
          &DECAY_SOLID must be specified in the griddata block.'

(shortened a bit, and corrected "coefficients ... is" to "coefficient ... is")

aprovost-usgs, line 1: I believe logical operators associate from left to right, so the following

      elseif (this%idcy == 2 .and. this%idcysrc == 1 .or. &
              this%idcysrc == 3) then ! zero order decay aqueous phase

which would evaluate the "and" first, then the "or", should be

      elseif (this%idcy == 2 .and. (this%idcysrc == 1 .or. &
              this%idcysrc == 3)) then ! zero order decay aqueous phase

Likewise,

      elseif (this%idcy == 2 .and. this%idcysrc == 2 .or. &
              this%idcysrc == 3) then ! zero order decay in the solid phase

should be

      elseif (this%idcy == 2 .and. (this%idcysrc == 2 .or. &
              this%idcysrc == 3)) then ! zero order decay in the solid phase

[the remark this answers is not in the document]
vivekbedekar: Reminds me of handling this in MT3D to avoid negative concentrations. GWT and GWE may need different treatment of zeroth-order decay.

autotest/test_gwe_decay01.py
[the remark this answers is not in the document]
cneyens: @emorway-usgs asked me via email to reply. Yes, you can mention me by name here.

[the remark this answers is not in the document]
aprovost-usgs: @cneyens Thanks Cas

src/Model/GroundWaterEnergy/gwe-est.f90
emorway-usgs: Fixed with ae887bd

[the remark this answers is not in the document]
emorway-usgs: Removed function `get_zero_order_decay` with a853d93.  Also cleaned up some of the comments

emorway-usgs: Fix for this included with 8e3021e

emorway-usgs: Fixed in 1f51717

emorway-usgs: Fixed in 6a93a3e

emorway-usgs: Fixed with f5c2c94

doc/mf6io/mf6ivar/dfn/gwe-est.dfn
emorway-usgs: Fixed in 2dc749c

autotest/test_gwe_decay01.py
emorway-usgs: New test added with 179b7de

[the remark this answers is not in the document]
emorway-usgs: @aprovost-usgs and I discussed the existing test on a Teams call.  Added a second test as well.

autotest/test_gwe_decay02.py
aprovost-usgs, line 1: The withdrawal rate in ESL is hardwired to -400.  That's the correct number for this particular setup, but maybe consider calculating it from
n * gamma_w + (1 - n) * gamma_s * rho_s
to make it easier for someone to play with the example in the future

doc/mf6io/mf6ivar/dfn/gwe-est.dfn
aprovost-usgs, line 1: After "rate per mass" consider adding "(not volume)" just to be super explicit?

aprovost-usgs, line 1: After "energy per length cubed" consider adding "(volume of water)" to be more specific?  (The corresponding description for the solid is nice in that regard because it explicitly mentions "mass of solid")

aprovost-usgs (commented):
just some minor comments for your consideration

doc/mf6io/mf6ivar/dfn/gwe-est.dfn
emorway-usgs: A change reflecting this suggestion applied in c52ed0e

emorway-usgs: A change reflecting this suggestion applied in c52ed0e

autotest/test_gwe_decay02.py
aprovost-usgs, line 1: Regarding

esl_spd = {
0: [[(0, 0, 0), abs(esl_amt)]],
}

do you want -esl_amt instead of abs(esl_amt) in the general case? I think the esl rate (negative for removal of heat) will be the negative of the total decay rate (positive for removal of heat)

emorway-usgs: Great idea, regrettable I didn't think of that..

Fixed with eb62e82

src/Model/GroundWaterEnergy/gwe-est.f90
```diff
@@ -262,37 +266,75 @@ subroutine est_fc_dcy(this, nodes, cold, cnew, nja, matrix_sln, &
       vcell = this%dis%area(n) * (this%dis%top(n) - this%dis%bot(n))
       swtpdt = this%fmi%gwfsat(n)
       !
-      ! -- add decay rate terms to accumulators
-      idiag = this%dis%con%ia(n)
-      if (this%idcy == 1) then
+      ! -- add zero-order decay rate terms to accumulators
+      if (this%idcy == 2 .and. (this%idcysrc == 1 .or. &
```
langevin-usgs, line 270: One thing you might consider is using an enumeration instead of integers for this.  I think we have all been transitioning that way, and it would make this statement much easier to digest.

```
if (this%idcy == DECAY_ZERO .and. (this%idcysrc == DECAY_AQUEOUS_ONLY .or. this%idcysrc == DECAY_AQUEOUS_SOLD))
```

src/Model/GroundWaterEnergy/gwe-est.f90
```diff
@@ -636,9 +766,22 @@ subroutine read_options(this)
         case ('FIRST_ORDER_DECAY')
           this%idcy = 1
           write (this%iout, fmtidcy1)
-        case ('ZERO_ORDER_DECAY')
+        case ('ZERO_ORDER_DECAY_WATER')
```
langevin-usgs, line 769: Should this be ZERO_ORDER_DECAY_AQUEOUS to be consistent elsewhere?

langevin-usgs (approved):
Looks good to me, only 2 suggestions to consider.  

1.  Would seem nice to use enumerations for `idcy` and `idcysrc`.
2. Should we rename "water" to "aqueous" to be consistent elsewhere?

aprovost-usgs:
@langevin-usgs I like your suggestion to be consistent, but the situation is a little muddy for me with respect to "aqueous" vs "water".  In GWT, aqueous is used throughout.  That makes sense because it's concerned with dissolved species -- "solutions" -- and "aqueous" is a common descriptor for that.  But in GWE we're dealing with thermal energy.  If it was just pure water being heated and cooled, "aqueous" would seem out of place.  But in general it can be an aqueous solution that's being heat and cooled, so referring to it as the "aqueous phase" makes sense from that standpoint.  That said, it seems to put a lot of (too much?) emphasis on the composition of the water.  Another consideration is that we've already gone down the "water" path with properties in GWE, like DENSITY_WATER and HEAT_CAPACITY_WATER.  That's a lot of hand wringing over a small point but just wanted to offer those perspectives

langevin-usgs:
> @langevin-usgs I like your suggestion to be consistent, but the situation is a little muddy for me with respect to "aqueous" vs "water". In GWT, aqueous is used throughout. That makes sense because it's concerned with dissolved species -- "solutions" -- and "aqueous" is a common descriptor for that. But in GWE we're dealing with thermal energy. If it was just pure water being heated and cooled, "aqueous" would seem out of place. But in general it can be an aqueous solution that's being heat and cooled, so referring to it as the "aqueous phase" makes sense from that standpoint. That said, it seems to put a lot of (too much?) emphasis on the composition of the water. Another consideration is that we've already gone down the "water" path with properties in GWE, like DENSITY_WATER and HEAT_CAPACITY_WATER. That's a lot of hand wringing over a small point but just wanted to offer those perspectives

Okay, sounds good.  Just wanted to make sure we weren't missing an opportunity to be consistent.  Based on your comment and seeing that "aqueous" is often used in the context of water as a solvent, it might be best to defer this for now and maybe think on how to come up with clear terminology.

src/Model/GroundWaterEnergy/gwe-est.f90
emorway-usgs: Suggested change implemented with b7a2330

src/Model/GroundWaterEnergy/gwe-est.f90
emorway-usgs: Based on the [follow-up](https://github.com/MODFLOW-USGS/modflow6/pull/2155#issuecomment-2627459652) comment below, I'm going to leave this as-is for the time being.

PR #2158, opened 2025-01-22
refactor(dis*): flatten idomain and move to base type

wpbonelli:
Store idomain as a 1D array in DisBaseType, rather than the proper grid shape in the concrete types. This deduplicates the variable and simplifes grid equivalence checks added in #2149 and planned in #2153.

___

### DIFF
--- a/src/Exchange/exg-gwfgwe.f90
+++ b/src/Exchange/exg-gwfgwe.f90
@@ -229,32 +229,10 @@ contains
     end if
     !
     ! -- Make sure idomains are identical
-    select type (gwfdis => gwfmodel%dis)
-    type is (DisType)
-      select type (gwedis => gwemodel%dis)
-      type is (DisType)
-        if (.not. all(gwfdis%idomain == gwedis%idomain)) then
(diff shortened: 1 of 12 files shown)


langevin-usgs (approved):
Sadly, this seems to have required more work than I would have thought.  I guess because of the mem_set_value, you need to allocate a local idomain multidimensional array and then copy those values into the 1d disbase member.  And the changes also propagate into the converter.

mjr-deltares:
A quick note: this will have consequences for anyone using IDOMAIN in its current shapes through the API. Personally, I have code that relies on the current (multi-dim) arrays, for example, to render VTK files. This will break then. I am not trying to block this, because I can easily adapt, just wanted to share the heads up.

wpbonelli:
After chatting with @mjr-deltares we thought this should wait, at least until a non-bugfix release, since it will break API usages. Maybe also consider alternatives.

@langevin-usgs

wpbonelli:
closing as low priority, maybe reopen if revisited someday

PR #2160, opened 2025-01-23
feat(swf-pcp): add precipitation package for CHF and OLF models

langevin-usgs:
Add a precipitation package to the channel flow and overland flow surface water models.  The new package allows specification of a precipitation rate in length units per time and multiples that value by the maximum area (for a channel) and by the cell area for the overland model to come up with a volumetric flow rate.  Precipitation rates must be zero or greater or the program terminates with an error.


jdhughes-usgs:
@langevin-usgs does it multiply the recharge by the maximum surface area and take et out based on the calculate surface area?

langevin-usgs:
> @langevin-usgs does it multiply the recharge by the maximum surface area and take et out based on the calculate surface area?
For the channel model, it does use the water surface area (which is dependent on stage) to calculate a volumetric flow rate.  For the overland flow model, it just uses the cell area.  My current thought is to only allow positive precipitation rates in the package, and require evapotranspiration to be represented using a separate evap package, which I'm working on next.  The evap package will also use the calculated water surface area to calculate the volumetric flow rate.  Let me know if you have suggestions for this behavior.

emorway-usgs:
If a cell in an OLF model also hosts a CHF reach, and both are using the precip package to apply water, is the surface area associated with the CHF deducted from the OLF surface area to avoid double counting of the precip?

langevin-usgs:
> If a cell in an OLF model also hosts a CHF reach, and both are using the precip package to apply water, is the surface area associated with the CHF deducted from the OLF surface area to avoid double counting of the precip?

@emorway-usgs, the next commit on this PR will apply rainfall over the max area of the channel (not the water surface area, which is dependent on water depth) so that the value is unchanging.  In the situation you mention, the user would have to adjust for possible double counting in the input data if channel reaches are large, though channel to overland flow connections are not available yet so this may evolve.  These may be issues to think about for the atmospheric forcing module.

PR #2164, opened 2025-01-25
fix(mf5to6,docs): fix doubled whitespace in 'internal factor'

martclanor:
This PR is a quick fix to what I believe is a typo error (doubled whitespace) in `INTERNAL FACTOR` which is also present in the mf6 data files in FloPy.


langevin-usgs:
Hi @martclanor, I don't think this is an error.  Our readers will work with any number of spaces.  Unless there is reason to make this change, I think our preference is to leave as is.  Feel free to reopen if we're missing something.

martclanor:
Yes, it is indeed not an error and I understand the decision. I just came across this while trying to build a syntax highlighter (also not a real issue there).

PR #2167, opened 2025-01-27
refactor(get-wetted-station): simplify and speed up calculation

langevin-usgs:
The routine for calculating the wetted station length was simplified based on a suggestion by @aprovost-usgs.  The resulting subroutine is shorter and likely slightly faster and will produce the same results.

### DIFF
--- a/src/Model/ModelUtilities/SfrCrossSectionUtils.f90
+++ b/src/Model/ModelUtilities/SfrCrossSectionUtils.f90
@@ -529,10 +529,7 @@ contains
     real(DP) :: xlen
     real(DP) :: dlen
     real(DP) :: slope
-    real(DP) :: dx
     real(DP) :: xt
-    real(DP) :: xt0
-    real(DP) :: xt1
     !
     ! -- calculate the minimum and maximum depth
     dmin = min(d0, d1)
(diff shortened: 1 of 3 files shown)


src/Model/ModelUtilities/SfrCrossSectionUtils.f90
aprovost-usgs, line 1: The "if d is less than or equal to ..." comment is still correct.  The next comment should describe the case d >= dmax: "if d is greater than or equal to the maximum value the station length is the full wetted width".  The comment that's there now ("if d is between ...") can be moved down to the final else case.

src/Model/ModelUtilities/SwfCxsUtils.f90
aprovost-usgs, line 1: see my comment on the previous file

src/Model/ModelUtilities/SfrCrossSectionUtils.f90
langevin-usgs: thanks for picking up on that

src/Model/ModelUtilities/SwfCxsUtils.f90
langevin-usgs: thanks

PR #2182, opened 2025-01-31
fix(utl-hpc.dfn): fix flopy subpackage metadata

wpbonelli:
I think the final term here should be `hpc_data`, to follow the convention used for other subpackages and produce the generated files as they currently appear in flopy &mdash; at least assuming the same code is to handle all subpackages. Seems like there is inconsistent handling currently, depending whether they are attached to simulation or to a model or package. Which we will need to dig out unless https://github.com/modflowpy/flopy/pull/2333 merges first.


mjr-deltares:
@wpbonelli : let me know if there is any implications for the Fortran side of this. Or, if it helps you to change anything there.

wpbonelli:
I think, given the passing tests, that this won't break flopy..

PR #2185, opened 2025-01-31
fix(prt): fix particle reporting

wpbonelli:
Reporting was incorrect for permanently unreleased particles. We want to report them once whenever they are to be released. #2177 fixed an issue where they were reported every time step, but the solution was incorrect, only reporting them if the release occurred in the first time step.

Also fix reporting for timed-out particles. The new status code 10 introduced by #2177 covers not only the simulation end but the particle reaching its stop time as well.

Update some comments and notes too.


src/Model/ModelUtilities/TrackFile.f90
aprovost-usgs, line 1: Did you intend for the TODO notes to remain on the istatus 0 and 10 descriptions?  The one on istatus 0 is probably helpful, as we talked about circling back to figure that out at some point.  Don't think it's critical for this release.  istatus 10 appears to be a done deal, and is so, can delete that TODO

aprovost-usgs (commented):
nice job on this issue.  just one last comment (about TODOs)

src/Model/ModelUtilities/TrackFile.f90
wpbonelli: Thanks- deleted the todo on 10.

aprovost-usgs (commented):
looks good

PR #2188, opened 2025-02-01
refactor(gwt-mst): apply enumeration to mst package for readability

emorway-usgs:
Improve `MST` code readability with enumeration

- [x] Similar changes made to `EST` in pull request #2155 

### DIFF
--- a/src/Model/GroundWaterTransport/gwt-mst.f90
+++ b/src/Model/GroundWaterTransport/gwt-mst.f90
@@ -109,7 +109,6 @@ contains
   !> @ brief Create a new package object
   !!
   !!  Create a new MST object
-  !!
   !<
   subroutine mst_cr(mstobj, name_model, inunit, iout, fmi)
     ! -- dummy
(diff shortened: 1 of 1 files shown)


src/Model/GroundWaterTransport/gwt-mst.f90
```diff
@@ -855,11 +871,11 @@ subroutine mst_calc_csrb(this, cnew)
       csrb = DZERO
       if (this%ibound(n) > 0) then
         distcoef = this%distcoef(n)
-        if (this%isrb == 1) then
+        if (this%isrb == SORPTION_LINEAR) then
```
langevin-usgs, line 874: Not important, but just to mention,  you could use a case statement here (and elsewhere) instead to improve clarity.  

```
select case (this%isrb)
case (SORPTION_LINEAR)
...
case (SORPTION_FREUND)
...
case (SORPTION_LANG)

end select
```

langevin-usgs (approved):
Looks nice @emorway-usgs.  Hope you like the enums.  I think it's clearer.

src/Model/GroundWaterTransport/gwt-mst.f90
emorway-usgs: Thanks for reminding me of this use case for an `ENUMERATE`, implemented in 6f0d6bb

PR #2202, opened 2025-02-05
refactor(prt): deconflict naming, use enum for tracking level

wpbonelli:
`Particle[Store]Type%idomain` has nothing to do with discretization IDOMAIN, name it `itrdomain` for "tracking domain" to clarify what we mean. Also use an enum for method tracking level.

### DIFF
--- a/doc/mf6io/prt/prt.tex
+++ b/doc/mf6io/prt/prt.tex
@@ -115,6 +115,7 @@ The ISTATUS field indicates the status of the particle:
 \item \texttt{7}: particle terminated in an inactive cell
 \item \texttt{8}: particle terminated immediately upon release into a dry cell
 \item \texttt{9}: particle terminated in a subcell with no exit face
+\item \texttt{10}: particle terminated at stop time or end of simulation
 \end{description}
 
 The IREASON field indicates the reason the particle track record was saved:
--- a/src/Solution/ParticleTracker/Particle.f90
+++ b/src/Solution/ParticleTracker/Particle.f90
@@ -14,6 +14,12 @@ module ParticleModule
   !! more methods as appropriate for finer-grained levels.
   integer, parameter :: MAX_LEVEL = 4
 
+  !> @brief Particle tracking level.
+  !!
+  !! Identifies the domain over which a tracking method is responsible for
+  !! moving a particle. Tracking methods at lower levels (coarser-grained)
+  !! defer to higher-level (finer-grained) methods within their subdomains.
+  !<
   enum, bind(C)
     enumerator :: LVL_MODEL = 1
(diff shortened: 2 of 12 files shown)


doc/mf6io/prt/prt.tex
```diff
@@ -115,6 +115,7 @@ \subsection{Particle Track Output}
 \item \texttt{7}: particle terminated in an inactive cell
 \item \texttt{8}: particle terminated immediately upon release into a dry cell
 \item \texttt{9}: particle terminated in a subcell with no exit face
+\item \texttt{10}: particle terminated at stop time or end of simulation
```
wpbonelli, line 118: neglected to add this in #2185 and followup

src/Solution/ParticleTracker/Particle.f90
aprovost-usgs, line 1: Enumerators are a great idea.  My only question is where's the best place to define them.  Ideally, the particle shouldn't have to know what kind of a domain it's being tracked though.  (Hopefully that's true except for the new enumerators, but I'm not 100% sure to be honest.)  Could they be defined in the relevant tracking methods? In addition to a desirable separation of concerns, this would also allow them to be customized for different kinds of tracking domains in the future, e.g., a stream discretized into reaches.  Would be happy to chat if you think this is feasible and worth exploring

aprovost-usgs (commented):
great idea.  just one comment - happy to chat

src/Solution/ParticleTracker/Particle.f90
wpbonelli: Good point- agree that method is the better place to define these. Maybe instead of an enum, each level could then just be a constant defined in the corresponding base method type.

wpbonelli: I figure the particle module is a good place for the status and event enums though

aprovost-usgs: totally agree - the status and event enums are best kept in the particle module

aprovost-usgs: regarding domain levels - so each method type would have its level baked in?  seems like that would work

wpbonelli:
This one can wait until after the release, since it sounds like we want to rethink where to define tracking level constants. Maybe introduce a base method type for each level, like we just did for cells in #2203?

aprovost-usgs:
> This one can wait until after the release, since it sounds like we want to rethink where to define tracking level constants. Maybe introduce a base method type for each level, like we just did for cells in #2203?

maybe that's a good approach - agree it can wait, and let's think on it some more.  i'm generally wary of adding too many classes, but sometimes it just makes sense

wpbonelli:
superseded

PR #2203, opened 2025-02-05
refactor(prt): introduce cell method type/module

wpbonelli:
Introduce `CellMethodType` and move the `check()` routine there from the base `MethodType`. These checks occur immediately before tracking begins across a cell, so this is the proper scope &mdash; they didn't belong in the base method type.


src/Solution/ParticleTracker/MethodCell.f90
aprovost-usgs, line 1: This is a logical place for the check. The question with introducing a new class is of course, is it worth it.  MethodCell does only one thing right now, and I suppose an alternative might've been to make the check a utility.  But I think this operation is so fundamental to the cell methods that creating MethodCell for it is justifiable

PR #2219, opened 2025-02-11
fix(netcdf): add all models to model inputs list

mjreno:
Idm maintains a model inputs list so stress period data can be loaded at the right time.  To this point, input models have been added to the list only if it contains packages with dynamic data.  The NetCDF export capability uses this model list as well but in this case all input models are relevant so a NetCDF output file can be created whether or not the model has dynamic data.


mjr-deltares (approved):

mjr-deltares:
@mjreno Thanks for this! The models are generating netcdf output as expected now :-)

PR #2236, opened 2025-03-05
Update test_gwf_newton_under_relaxation.py

jlarsen-usgs:
* updates for model splitter node mapping changes
* corresponds to modflowpy/flopy#2465


jlarsen-usgs:
@wpbonelli 

What needs to be done to solve the pixi lockfile issue that I'm seeing? I just added a h5py dependency to `pixi.toml` for the new model splitter updates in FloPy. I'm not super familiar with pixi yet and don't want to break things

wpbonelli:
@jlarsen-usgs after adding new dependencies to `pixi.toml` the lock file `pixi.lock` needs updating. Assuming you have pixi you can `pixi install -a`, that should do it. The changeset might be huge, fair warning.

PR #2237, opened 2025-03-06
chore: bump min python version to 3.10

wpbonelli:
In the interest of eventually adopting the [scientific python convention](https://scientific-python.org/specs/spec-0000/) which is a narrower band of versions, but not making the transition too abrupt, only bumping the min to 3.10 for now


environment.yml
```diff
@@ -5,7 +5,7 @@ channels:
   - nodefaults
 
 dependencies:
-  - python>=3.9
+  - python>=3.10
```
jdhughes-dev, line 8: Should we fix this at the minimum or +1. I have recently run into issues where the latest version of python was installed and then ended up with datetime deprecation warnings. I ended up downgrading python to 3.11.

Just a thought.

jdhughes-dev: This would also be consistent with pixi.toml

jdhughes-dev (commented):
Just a suggestion. I am good otherwise.

environment.yml
wpbonelli: No strong sense for whether pixi.toml and environment.yml should be at min, max or in the middle. Looking elsewhere

- pandas is at [minimum](https://github.com/pandas-dev/pandas/blob/main/environment.yml#L6)
- numpy is [min + 1](https://github.com/numpy/numpy/blob/main/environment.yml#L10)

I don't know if this reflects different conventions for these projects, or if it is just haphazard.

I've hit the datetime warnings too. Maybe [this is the fix](https://til.simonwillison.net/python/utc-warning-fix)? I can give it a shot.

PR #2246, opened 2025-03-15
docs(mf6io): clarify/deduplicate pertim and totim

wpbonelli:
`PERTIM` and `TOTIM` were duplicated in the "Format of Budget File" section. Remove the duplicates and tweak the phrasing to reflect what we already say in the [comments](https://github.com/MODFLOW-ORG/modflow6/blob/b06308efc1106b1dfb1d8203673b1f27eb0ad183/src/Timing/tdis.f90#L30), which is a bit more verbose but IMO clearer.


christianlangevin (approved):
definitely clearer

PR #2248, opened 2025-03-17
docs: add release note for new GWE example

wpbonelli:
https://github.com/MODFLOW-ORG/modflow6-examples/pull/265

@cneyens let me know if the phrasing needs any tweaks. We add release notes for new example models, just FYI.

### DIFF
--- a/doc/ReleaseNotes/develop.toml
+++ b/doc/ReleaseNotes/develop.toml
@@ -16,3 +16,7 @@ parallel = "PARALLEL"
 section = "features"
 subsection = "basic"
 description = "The binary grid file's name may now be specified in all discretization packages with option GRB6 FILEOUT followed by a file path. If this option is not provided, the binary grid file will be named as before, identical to the discretization file name plus a ``.grb'' extension. Note that renaming the binary grid file may break downstream integrations which expect the default name."
+
+[[items]]
+section = "examples"
+description = "A new example was added demonstrating thermal loading of multiple borehole heat exchangers with the Groundwater Energy (GWE) model."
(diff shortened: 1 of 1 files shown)


doc/ReleaseNotes/develop.toml
```diff
@@ -16,3 +16,7 @@ parallel = "PARALLEL"
 section = "features"
 subsection = "basic"
 description = "The binary grid file's name may now be specified in all discretization packages with option GRB6 FILEOUT followed by a file path. If this option is not provided, the binary grid file will be named as before, identical to the discretization file name plus a ``.grb'' extension. Note that renaming the binary grid file may break downstream integrations which expect the default name."
+
+[[items]]
+section = "examples"
+description = "A new example was added demonstrating thermal loading of multiple borehole heat exchangers with the Groundwater Energy (GWE) model."
```
cneyens, line 22: Perhaps: "... demonstrating _transient_ thermal loading ..." to differentiate from the existing `ex-gwe-geothermal` example which also uses multiple borehole heat exchangers.

emorway-usgs: Thanks for tying-up this loose-end, @wpbonelli. In addition to what @cneyens is suggesting, you might also add a sentence highlighting that the new example compares well to an analytical solution: "The simulated GWE results match the temperature change predicted by the analytical solution for the multiple borehole heat exchangers used in this problem."

emorway-usgs (approved):
Just a couple of minor comments.  Looks great @wpbonelli.  Looks like there was a hiccup with one of the autotests?

wpbonelli:
@emorway-usgs I think CI will fail until #2236 merges

PR #2291, opened 2025-04-22
refactor(dfn2f90.py): use jinja to generate fortran

wpbonelli:
Use jinja in `dfn2f90.py` to generate Fortran source files from DFNs. Still work in progress. As a next step (separately from this PR) we can switch to consuming TOML definitions as [provided by devtools](https://modflow-devtools.readthedocs.io/en/latest/md/dfn.html#toml-migration), like we have done for flopy codegen.

___


utils/idmloader/scripts/dfn2f90.py
```diff
@@ -1,968 +1,386 @@
...
-        )
 
-        fh.write(s)
+def _get_template_env() -> Environment:
+    template_loader = FileSystemLoader(str(TEMPLATES_PATH))
+    template_env = Environment(
+        loader=template_loader,
+        trim_blocks=True,
+        lstrip_blocks=True,
+        line_statement_prefix="_",
```
mjreno: cleanup? or maybe there was some intent with this.

wpbonelli: it's not necessary you're right

PR #2292, opened 2025-04-23
refactor(dfn2f90.py): accept 1+ dfn files or directories

wpbonelli:
Instead of accepting a text file specifying DFN files &mdash; a temporary measure while IDM supports a subset of the full specification &mdash; just accept one or more path arguments, each of which can be a DFN file or a folder containing DFN files, and feed the `dfns.txt` file to this from the shell.

A prep step before code generation is switched to consume TOML DFNs, coming shortly.

**Note to model developers**: after this, the behavior of running `dfn2f90.py` directly with no arguments will change to include all DFNs by default. To get the old behavior (i.e. the subset of DFNs IDM supports), it will be necessary to use the pixi task or feed `dfns.txt` to the script with e.g. 

```shell
cat dfns.txt | xargs python ...
```

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -102,7 +102,7 @@ get-exes = { cmd = "pytest -v --durations 0 get_exes.py", cwd = "autotest" }
 autotest = { cmd = "pytest -v -n auto --durations 0 --keep-failed .failed", cwd = "autotest" }
 
 # common developer tasks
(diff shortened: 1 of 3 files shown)


wpbonelli:
@christianlangevin @emorway-usgs just wanted to run the above by you both since you are developing models which may be integrated with IDM.

wpbonelli:
discussed offline with @mjreno

PR #2300, opened 2025-05-02
test(framework): allow access to local external models

wpbonelli:
Support remote or local model access in the external model tests via an identical pattern. This uses some new stuff in the [models API](https://modflow-devtools.readthedocs.io/en/latest/md/models.html).  This is relevant for anyone needing to modify [test models](https://github.com/MODFLOW-ORG/modflow6-testmodels) during MF6 development.

By default, as has been the case since we started using the models API recently, the tests will use the GitHub-backed Pooch registry. To test against local models, introduce a `--models-path` pytest CLI option. If one or more is provided, the paths will be searched for models and tests will be run against those instead of models from the official registry.

By default, only MF6 models are discovered. There is a `--namefile-pattern` option to relax the discovery for MF2005 etc.

E.g. to run MF6 models in a local checkout of the `modflow6-testmodels` repo, from `<project root>/autotest/` of the MODFLOW 6 repo

```shell
pytest -m "external" --models-path ../../path/to/modflow6-testmodels/
# or pixi run autotest -m ...
```

Or to run mf5to6 converter tests against the mf2005 models

```shell
pytest -m "external" --models-path ../../path/to/modflow6-testmodels/mf5to6 --namefile-pattern "*.nam"
```

This PR updates the developer docs to reflect the above. Previously they described the old [fixtures pattern](https://modflow-devtools.readthedocs.io/en/latest/md/fixtures.html#loading-example-models).

Also update the CI to checkout the external model repos and use the new local registry mechanism rather than using the official registry. Since moving to the models API we have tested the models as frozen by the official registry but we want to test the latest. (Speaking of which, we will need to move to the `develop`/`master` branching strategy in test models and large test models repos so the official registry has something stable to point at &mdash; or consider moving those models into this repo.)

Also consolidate the external model tests from

- `test_testmodels_mf6.py`
- `test_testmodels_mf5to6.py`
- `test_largetestmodels.py`

into a single file `test_external_models.py`.


mjreno (approved):
Thanks for doing this @wpbonelli, looks good.  Seems like a decision on test model repository branching strategy is relevant to take full advantage- including also the concept of dev feature?

autotest/test_external_models.py
```diff
@@ -0,0 +1,134 @@
...
+    registry,
+    model_name,
+    tmp_path,
+    markers,
+    targets,
+    function_tmpdir,
+    original_regression,
+):
+    exclude = any(s in model_name for s in EXCLUDE)
+    devonly = "dev" in model_name and "not developmode" in markers
```
wpbonelli, line 81: > including also the concept of dev feature

@mjreno we talked offline, just to note for passersby, there is a mechanism here for test models which should run only in develop mode. I think the convention in the testmodels repo is to end the model name with "_dev".

wpbonelli:
> Seems like a decision on test model repository branching strategy is relevant to take full advantage

Agreed. I planned to break out develop/master on the testmodels repos after 6.6.2 is out the door. I think that is risk-free, and buys us coarse-grained (release-level) versioning for the new models registry, where there is no versioning right now.

If all the test models lived in the same repo I think we could version the registry down to the level of commits, like Pooch [recommends](https://www.fatiando.org/pooch/latest/sample-data.html#versioning). Their pattern is designed for Python packages where everything is in the same repo but it is adaptable. The registry and the data can live in separate projects, even several different projects (this is how it works right now) but if all the test models/examples/etc are here in mf6, we can version them together.

PR #2316, opened 2025-05-09
chore: update mf6ivar to ignore removed options

wpbonelli:
Previously removed options still made it into the MF6IO doc. Revise `mf6ivar.py` to filter them out. And simplify some conditionals.

### DIFF
--- a/doc/mf6io/mf6ivar/mf6ivar.py
+++ b/doc/mf6io/mf6ivar/mf6ivar.py
@@ -320,19 +320,16 @@ def write_block(vardict, block, blk_var_list, varexcludeprefix=None, indent=None
                 n = name.upper()
                 if n.startswith(varexcludeprefix.upper()):
                     addv = False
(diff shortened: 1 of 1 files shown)


doc/mf6io/mf6ivar/mf6ivar.py
```diff
@@ -320,19 +320,16 @@ def write_block(vardict, block, blk_var_list, varexcludeprefix=None, indent=None
...
-                    addv = False
+            if v.get("in_record", "") == "true":
+                # do not separately include this variable
+                # because it is part of a record
+                addv = False
+            if v.get("block_variable", "") == "true":
+                # do not separately include this variable
+                # because it is part of a record
+                addv = False
+            if v.get("deprecated", "") != "" or v.get("removed", "") != "":
```
wpbonelli, line 331: we were missing a removed check here. the other changes are just cleanup.

PR #2336, opened 2025-06-02
fix(build): add to library path for macos to fix extended

mjr-deltares:


wpbonelli:
#2342 fixes the macos issue

PR #2345, opened 2025-06-05
Add test-drive and pkg-config to pixi and conda envs

mwtoews:
This adds `test-drive` and `pkg-config` to pixi and conda environments. The former package was recently brought to conda-forge, and can only be discovered with the cross-platform conda-forge version of pkg-config.

With this change, it seems the developer docs to install test-drive are no longer relevant, but let me know if it is still relevant and it can be kept or rephrased.


wpbonelli (approved):
Thanks @mwtoews, didn't realize test-drive was on conda-forge.

wpbonelli:
We might have to revert this, as there seem to be differences between the pkg-config on conda-forge and the default one. The extended build doesn't like the conda-forge one.

mwtoews:
I can take a look at removing just `pkg-config` to see if that resolves the issue.

PR #2348, opened 2025-06-06
Remove pkg-config from pixi.toml/environment.yml

mwtoews:
This is a follow-up to #2345 to remove pkg-config from pixi.toml/environment.yml. Use pixi's env feature for the setup task to update PKG_CONFIG_PATH.


mwtoews:
I don't believe the one failing workflow "Test GNU fortran (windows-2022, false)" is related to pkg-config and/or test-drive, since this workflow has never relied on test-drive (tests before and after always have "Run-time dependency test-drive found: NO (tried pkgconfig and cmake)"). The error messages from "Test examples" don't show any useful trace information for their failure.

wpbonelli (approved):
@mwtoews you are right, it is unrelated, fixed in #2352. thanks for doing this

PR #2370, opened 2025-06-27
feat(SVD) Add the SVD-algorithm and the pseudoinverse method

Manangka:
This PR adds the SVD-algorithm and the pseudoinverse method.
This is in preparation of the new flux limiter that uses the pseudoinverse to invert rank deficient matrices.
The pinv isn't hooked up to anything yet.

The SVD-algorithm is quite complicated. I implemented it using the algorithms described in the book "Matrix Computations" of Golub and Van Loan. I've got some personal notes available which i can share if requested.

Checklist of items for pull request

- [ ] Replaced section above with description of pull request
- [ ] Closed issue #xxxx
- [ ] Referenced issue or pull request #xxxx
- [ ] Added new test or modified an existing test
- [ ] Ran `ruff` on new and modified python scripts in .doc, autotests, doc, distribution, pymake, and utils subdirectories.
- [ ] Formatted new and modified Fortran source files with `fprettify`
- [ ] Added doxygen comments to new and modified procedures
- [ ] Updated meson files, makefiles, and Visual Studio project files for new source files
- [ ] Updated [definition files](/MODFLOW-ORG/modflow6/tree/develop/doc/mf6io/mf6ivar)
- [ ] Updated [develop.toml](/MODFLOW-ORG/modflow6/doc/ReleaseNotes/develop.toml) with a plain-language description of the bug fix, change, feature; required for changes that may affect users
- [ ] Updated [input and output guide](/MODFLOW-ORG/modflow6/doc/mf6io)
- [ ] Removed checklist items not relevant to this pull request

For additional information see [instructions for contributing](/MODFLOW-ORG/modflow6/.github/CONTRIBUTING.md) and [instructions for developing](/MODFLOW-ORG/modflow6/.github/DEVELOPER.md).

PR #2367, opened 2025-06-23
refactor(tsp-adv) Move different interpolation schemes to separate files

Manangka:
This PR is in preparation of the new flux limiter.
It refactors the tsp-adv scheme, moving the different interpolation schemes to separate files. This makes the code easy extentably for other flux limiter. 

Furthermore an ADV_SCHEME enum has been introduced to improve readability and make it easier to find where which scheme is referenced

No functional changes have been made. Code has just been moved around

Checklist of items for pull request

- [ ] Replaced section above with description of pull request
- [ ] Closed issue #xxxx
- [ ] Referenced issue or pull request #xxxx
- [ ] Added new test or modified an existing test
- [ ] Ran `ruff` on new and modified python scripts in .doc, autotests, doc, distribution, pymake, and utils subdirectories.
- [ ] Formatted new and modified Fortran source files with `fprettify`
- [ ] Added doxygen comments to new and modified procedures
- [ ] Updated meson files, makefiles, and Visual Studio project files for new source files
- [ ] Updated [definition files](/MODFLOW-ORG/modflow6/tree/develop/doc/mf6io/mf6ivar)
- [ ] Updated [develop.toml](/MODFLOW-ORG/modflow6/doc/ReleaseNotes/develop.toml) with a plain-language description of the bug fix, change, feature; required for changes that may affect users
- [ ] Updated [input and output guide](/MODFLOW-ORG/modflow6/doc/mf6io)
- [ ] Removed checklist items not relevant to this pull request

For additional information see [instructions for contributing](/MODFLOW-ORG/modflow6/.github/CONTRIBUTING.md) and [instructions for developing](/MODFLOW-ORG/modflow6/.github/DEVELOPER.md).

PR #2374, opened 2025-06-30
(feat) UTVD fluxlimiter

Manangka:
This commit adds the new TVD flux limiter based on the paper of Darwish.

For now i called the new limiter **UTVD** where U stands for unstructured. But as @christianlangevin and i discussed last week this doesn't cover the because it is also suited for structured grids.

A small description of the code layout. 
The main starting point is the tsp-adv.f90 file. Here the cell based gradient is created and passed on to the UTVD class.
I would recommend reviewing tsp-adv.f90 -> IGradient.f90/LeastSquaredGradient.f90 -> UTVDScheme.f90 first to follow the logic of the implementation. 

A helper class `DisUtils `is added which provided common functions e.g. getting a cell centroid location

I've updated existing advection tests to also test the new limiter.
 
It depends on #2367 and #2370 

- [ ] Closed issue #xxxx


src/Model/TransportModel/tsp-adv.f90
aprovost-usgs, line 1: very minor thing - might consider replacing "LeastSquaredGradient" with "LeastSquaresGradient" in the names that contain it, as I think that's the more common term

src/Model/TransportModel/tsp-adv.f90
```diff
@@ -123,6 +128,9 @@ subroutine adv_ar(this, dis, ibound)
     this%dis => dis
     this%ibound => ibound
     !
+    ! -- Compute the gradient operator
+    this%gradient = LeastSquaredGradientType(this%dis, this%fmi)
+    !
```
aprovost-usgs: if the gradient calculation is cheap, maybe it doesn't matter, but if it's expensive, it would be good to do it only if it's needed

src/Model/TransportModel/Gradient/LeastSquaredGradient.f90
aprovost-usgs, line 1: great descriptions in the comments throughout

src/Model/TransportModel/Gradient/LeastSquaredGradient.f90
```diff
@@ -0,0 +1,170 @@
...
+
+  !> @brief Weighted least-squares gradient method for structured and unstructured grids.
+  !!
+  !! This class implements a least-squares gradient reconstruction for use on both structured and unstructured grids.
+  !! For each cell, it precomputes and caches a gradient operator using the Moore-Penrose pseudoinverse,
+  !! based on the geometry and connectivity of the mesh. The operator is created once during initialization
+  !! and can then be efficiently applied to any scalar field to compute the gradient in each cell.
+  !!
+  !! - The gradient operator is constructed using normalized direction vectors between cell centers,
+  !!   scaled by the inverse of the distance.
```
aprovost-usgs: What is called the "gradient operator" here and other places in the code is presumably what we've been calling the "G" matrix in our derivations?  If the relevant TVD or least-squares literature refers to this as the gradient operator, then it's fine.  But if not, consider a different name (such a "G matrix," "g_matrix," etc).  In common mathematical terminology, the gradient operator is the "derivative"-type operator represented by del, but that's apparently not what is meant here.  Rather, this a matrix that operates on the gradient vector (hence "gradient operator") if I understand correctly

src/Model/TransportModel/Gradient/LeastSquaredGradient.f90
```diff
@@ -0,0 +1,170 @@
...
+
+      d(local_pos, :) = dnm / length
+      grad_scale(local_pos, local_pos) = 1.0_dp / length
+
+      local_pos = local_pos + 1
+    end do
+
+    d_trans = transpose(d)
+
+    ! Compute the G and inverse G matrices
```
aprovost-usgs: pseudo-inverse, correct?

src/Model/Discretization/DisUtils.f90
```diff
@@ -0,0 +1,110 @@
...
+                               y_dir, z_dir, length)
+
+    ! -- Compute the distance vector
+    d(1) = x_dir * length
+    d(2) = y_dir * length
+    d(3) = z_dir * length
+
+  end function node_distance
+
+  !> @brief Returns the centroid coordinates of a given cell.
```
aprovost-usgs: Might we also already have such a function built into the discretization objects?

src/Model/Discretization/DisUtils.f90
```diff
@@ -0,0 +1,110 @@
...
+  !> @brief Returns the vector distance from cell n to cell m.
+  !!
+  !! This function computes the vector from the center of cell `n` to the center of cell `m`
+  !! in the discretization. If the cells are directly connected, the vector is computed along
+  !! the connection direction, taking into account cell geometry and, if available, cell saturation.
+  !! If the cells are not directly connected (e.g., when using an extended stencil such as neighbors-of-neighbors),
+  !! the vector is simply the difference between their centroids: `d = centroid(m) - centroid(n)`.
+  !! The returned vector always points from cell `n` to cell `m`.
+  !<
+  function node_distance(dis, fmi, n, m) result(d)
```
aprovost-usgs: I'm pretty sure @christianlangevin wrote similar routines that we use in XT3D -- I believe there's one in each discretization type, and it calculates the "connection vector" between two cell centers

aprovost-usgs (approved):
Elegant code.  Just minor things for your consideration

src/Model/TransportModel/Gradient/LeastSquaredGradient.f90
Manangka: The "gradient operator" ss a term i coined myself. 
When you have the equation:
image
Then the Gradient operator is:
image

I've looked around and found that this is sometime called the "gradient reconstruction" matrix =, denoted with "R"

I've updated the code and comments to use this new symbol

src/Model/TransportModel/Gradient/LeastSquaredGradient.f90
Manangka: Yes you're right. Changed it

src/Model/TransportModel/tsp-adv.f90
Manangka: Changed it

src/Model/TransportModel/tsp-adv.f90
Manangka: Fair point. At the moment it is only needed for the new limiter. I moved this line down, inside the ADV_SCHEME_UTVD. So it will only be calculated when that limiter is selected

src/Model/TransportModel/Gradient/LeastSquaredGradient.f90
Manangka: Thank you! I must admit, I've been using copilot to enhance my comments

src/Model/Discretization/DisUtils.f90
Manangka: I didn't find such a method. @langevin do you know is such a method exist?

src/Model/Discretization/DisUtils.f90
Manangka: This method wraps that method. The "connection vector" works great when cells are physically connected, however with the least squares method it is also possible to extend the stencil to add neighbours of neighbours. In that case the "connection vector" doesnt't work.

Manangka: I did notice that a mistake i made. When calling the connection_vector method i should not provide the saturation values and the noze=true flag because we are alwas interested at the distance from the centers of the cells

src/Model/TransportModel/Gradient/LeastSquaredGradient.f90
aprovost-usgs: it seems to be working well for you!

src/Model/TransportModel/Gradient/LeastSquaredGradient.f90
aprovost-usgs: gradient reconstruction matrix sounds great -- descriptive of what it does and unlikely to be confused with another operator

src/Model/Discretization/DisUtils.f90
aprovost-usgs: I'll have a look in the code and see if I can find such a routine.  (I could be mistaken)

src/Model/Discretization/DisUtils.f90
aprovost-usgs: Got it - sounds good

src/Model/Discretization/DisUtils.f90
aprovost-usgs: @Manangka i didn't find a function that returns the cell center (including the vertical, which is really the missing piece you're computing).  your function is straightforward - was just wondering if it was duplicating something we already had.

one detail is that the function name and comments refer to these cell centers as 'centroids.'  it's possible that we call them centroids elsewhere in the code, but i think in general we try to avoid that since, strictly speaking, they're not necessarily centroids.  we typically call them cell centers or nodes

src/Model/Discretization/DisUtils.f90
aprovost-usgs: @Manangka thinking again about whether to provide saturation.  if a cell is partially saturated (water table between the user-specified top and bottom of the cell), the cell center is halfway between the water table and the cell bottom.  do you want to account for this, or is that irrelevant for your purpose?

src/Model/Discretization/DisUtils.f90
Manangka: Alright, i leave the code as is but will change the name to `cell_center`

src/Model/Discretization/DisUtils.f90
Manangka: Interesting question.  I think it is irrelevant, at least for the Darwish method

In the Darwish method we use the distance between cell centers, but by follow up papers it is clear that this could be wrong. The connection vector should be perpendicular to connecting face. This is why the Darwish method works great on structured and voronoi grids. This would also mean that regardless of the saturation level the connection vector should be constant and perpendicular.

For the LSG-method I'm not so sure. We use it to create the Reconstruction matrix. We can do this once and store it under the assumption that the grid is static. If the cell centers are moving we technically  don't have a static grid anymore.  However the cell center's x and y coordinate stay the same so it would only affect the gradient computation in the z-direction. And since for the Darwish method we are only interested in the change normal to the connecting face i would suspect that this would have little impact.

aprovost-usgs: Thanks for explaining those issues.  Sounds like there's potential for refinement in the future, but using cell centers and assuming they're static is an approximation that you've demonstrated still typically gives improvement over the original MF6 TVD scheme

src/Model/TransportModel/Gradient/LeastSquaresGradient.f90
```diff
@@ -0,0 +1,167 @@
...
+    end do
+
+    d_trans = transpose(d)
+
+    ! Compute the G and pseudo-inverse G matrices
+    g = matmul(d_trans, d)
+    g_inv = pinv(g)
+
+    ! Compute the gradient reconstructions matrix
+    R = matmul(matmul(g_inv, d_trans), grad_scale)
```
aprovost-usgs: @Manangka While reviewing PR #2385 it occurred to me that (I think) the sequence for calculating R should simply be

    d_inv = pinv(d)
    R = matmul(d_inv, grad_scale)

This is based on the fact that (given the way d is already scaled) I believe the normal equations for this least-squares problem are 

[d_trans] [d] [grad_c] = [d_trans] [grad_scale] [dc]

for which the solution is

[grad_c] = [pinv(d)] [grad_scale] [dc]

It would be good to talk this through with you -- hopefully I'm mistaken.

Also, note that d is normalized by connection length (earlier in the routine):

      d(local_pos, :) = dnm / length
      grad_scale(local_pos, local_pos) = 1.0_dp / length

This is correct given that the choice of weighting is inverse length (which corresponds to weighting the terms in the least-squares objective function by inverse length squared).  Note that for a different choice of weighting, (again, I think) you'd want to scale d by that weighting:

      grad_scale(local_pos, local_pos) = weight
      d(local_pos, :) = dnm * grad_scale(local_pos, local_pos)

This is simply a more general calculation -- for the inverse distance weighting you're using it would obviously give the same result.  No problem here, just making a note.

Manangka: @aprovost-usgs I've tested your suggestion and things simplify neatly I haven't encountered any stability problems or performance issues so I've switched to your formulation

src/Model/Connection/GweGweConnection.f90
```diff
@@ -264,9 +264,11 @@ subroutine setGridExtent(this)
     hasCnd = this%gweModel%incnd > 0
 
     if (hasAdv) then
-      if (this%iIfaceAdvScheme == 2) then
+      if (this%iIfaceAdvScheme == ADV_SCHEME_TVD .or. &
+          this%iIfaceAdvScheme == ADV_SCHEME_UTVD) then
         this%exg_stencil_depth = 2
```
mjr-deltares, line 269: @Manangka let's double check the stencil requirement on the new method offline

src/Model/TransportModel/tsp-adv.f90
```diff
@@ -133,6 +138,10 @@ subroutine adv_ar(this, dis, ibound)
     case (ADV_SCHEME_TVD)
       this%face_interpolation = &
         TVDSchemeType(this%dis, this%fmi, this%ibound)
+    case (ADV_SCHEME_UTVD)
+      this%gradient = LeastSquaresGradientType(this%dis)
+      this%face_interpolation = &
+        UTVDSchemeType(this%dis, this%fmi, this%gradient)
```
mjr-deltares, line 144: very nice to have such minimal changes to extend adv :-)

src/Model/TransportModel/InterpolationScheme/UTVDScheme.f90
```diff
@@ -0,0 +1,205 @@
...
+  function limiter(this, r) result(theta)
+    ! -- return
+    real(DP) :: theta ! limited slope
+    ! -- dummy
+    class(UTVDSchemeType) :: this
+    real(DP) :: r ! ratio of successive gradients
+
+    select case (this%limiter_id)
+    case (2) ! van Leer
+      theta = max(0.0_dp, min((r + dabs(r)) / (1.0_dp + dabs(r)), 2.0_dp))
```
mjr-deltares, line 189: I know we have DZERO, DONE, etc. but I like it better how you have done it here, it's readable

src/Model/TransportModel/InterpolationScheme/UTVDScheme.f90
```diff
@@ -0,0 +1,205 @@
...
+    ! high_order = cl1 / (cl1 + cl2) * alimiter * qnm * (phi(idn) - phi(iup))
+    ! This is split into two parts:
+    ! coef_up = coef_up - relative_distance * alimiter
+    ! coef_dn = coef_dn + relative_distance * alimiter
+
+  end function compute
+
+  subroutine find_local_extrema(this, n, phi, min_phi, max_phi)
+    ! -- dummy
+    class(UTVDSchemeType), target :: this
```
mjr-deltares: why is/was the target needed here?

src/Model/Discretization/DisUtils.f90
mjr-deltares, line 1: Nice to isolate generic functionality. There are similar units in the ModelUtilities folder. Should this go there as well?

autotest/test_par_gwt_henry.py
mjr-deltares, line 1: So it works in parallel too. Booyah!!

mjr-deltares (approved):
I really like how you did this. I do think we need to present the core design to the team so if they visit tsp-adv, they know what is going on there. You have increased the level of software engineering to come up with a better structured code, and succeeded. Let's make sure people can follow along.

src/Model/TransportModel/Gradient/LeastSquaresGradient.f90
```diff
@@ -0,0 +1,156 @@
...
+    local_pos = 1
+    do ipos = this%dis%con%ia(n) + 1, this%dis%con%ia(n + 1) - 1
+      m = this%dis%con%ja(ipos)
+      dc(local_pos) = phi_new(m) - phi_new(n)
+      local_pos = local_pos + 1
+    end do
+
+    ! Compute the cells gradient
+    R => this%R(n)%data
+    grad_c = matmul(R, dc)
```
aprovost-usgs, line 152: This implements a change we discussed in light of an earlier review, and you've reported encouraging results.  There may be a further improvement we can make to the rigor of the least-squares inversion, but investigating this can wait.  Agree with @mjr-delatres that there will be substantial benefit to merging this PR sooner rather than later, so codevelopers can begin to see how the code is engineered

aprovost-usgs (approved):
Nice job, @Manangka

src/Model/Connection/GweGweConnection.f90
Manangka: There is another PR incoming with some performance improvements. Let's include any changes that come out of our discussion in that PR.

src/Model/TransportModel/InterpolationScheme/UTVDScheme.f90
Manangka: Good catch. It shouldn't be needed. I removed it.

src/Model/TransportModel/Gradient/LeastSquaresGradient.f90
Manangka: I agree. There is another PR incoming with some performance improvements. It can be part of that PR or it can be a separate PR

PR #2376, opened 2025-07-01
refactor(prt): introduce particle event system

wpbonelli:
Building on #2373. Simple synchronous pub/sub.

Main motivation here is that prt needs to report things in many places throughout the solution instead of just one place like other models. This is more readable imo and cleans up some particle state transition code that was duplicated before. Tracking methods also no longer need to know about output files, just call the dispatcher to publish events.

We are paying for an allocation and a little extra overhead every time something happens to a particle, as opposed to just a procedure call before. But this seems worth it?

The dispatcher is responsible for the event lifecycle i.e. allocating before publishing, deallocating after consumption. If we want to batch events later on we will need to think on how to deallocate them on flush.

Only one consumer per dispatcher right now. This can be generalized to multiple consumers if needed.

I added a developer option `DEV_DUMP_EVENT_TRACE` to the output control package which prints verbose event summaries to stdout (below is from an extended tracking example with a single period/step, so tracking time goes past the end of the tdis)

```
Particle (Model: 2  PRP Package: 1  Release Point: 9  Release Time: 0.0000000000000000  Name: 9) released in (Layer: 1  Cell: 1  Zone: 0) at (Time 0.0000000000000000  Period 1  Timestep 1) with (Status: 1  Context: )
Particle (Model: 2  PRP Package: 1  Release Point: 9  Release Time: 0.0000000000000000  Name: 9) exited cell in (Layer: 1  Cell: 1  Zone: 0) at (Time 0.63459865171749452E-1  Period 1  Timestep 1) with (Status: 1  Context: )
...
Particle (Model: 2  PRP Package: 1  Release Point: 9  Release Time: 0.0000000000000000  Name: 9) exited cell in (Layer: 1  Cell: 90  Zone: 0) at (Time 42.728754848837930  Period 1  Timestep 1) with (Status: 1  Context: )
Particle (Model: 2  PRP Package: 1  Release Point: 9  Release Time: 0.0000000000000000  Name: 9) terminated in (Layer: 1  Cell: 100  Zone: 0) at (Time 42.728754848837930  Period 1  Timestep 1) with (Status: 5  Context: )
```

This is a convenience for debugging which will show all events, not just the ones selected for output to track files. The "context" is a string I stuck on the event type as another debugging aid, currently unused, maybe overkill.

Some more cleanup and reorganization too.


aprovost-usgs (approved):
Based on our discussion on 7/2/2025, it's good to merge

wpbonelli:
We discussed defining event routines in the relevant methods, rather than in the dispatcher. This seems more appropriate because events might be scoped to a particular method type, while the plumbing should be able to handle any event type. Will come back to this.

Issue #2393, opened 2025-07-11
mf5to6 does not write SFR CONNECTIONDATA block entry for SFR reaches with no connections


kzeiler:
**Describe the bug**
MF6 SFR input requires an entry in the CONNECTIONDATA block for all reaches, even for those reaches with no connections. However, mf5to6 skips writing a CONNECTIONDATA entry for reaches with no connections.

**To reproduce**
1. Download the model.zip file for the USGS SIR2017-5158 [data release](https://www.sciencebase.gov/catalog/item/631405bad34e36012efa3032).
2. Unzip the model.zip file.
3. Run mf5to6 to convert this MODFLOW-NWT model to MODFLOW 6
4. Open the MODFLOW 6 SFR Package input file, scroll to the beginning of the CONNECTIONDATA block and see that reaches 5 and 6 are missing entries; note that there are a total of 50 reaches with no connections that are missing from the CONNECTIONDATA block.
5. Alternatively, execute mf6 for the converted model input files and see error message from mf6 regarding missing CONNECTIONDATA information.

**Expected behavior**
Entries in the CONNECTIONDATA block should appear for all reaches, including reaches with no connections.

**Suggested solution**
Moving write() statements in the write_connectivity() subroutine of SfrPackageWriter.f90 outside the if statement checking whether each reach has >0 connections will then write the appropriate CONNECTIONDATA block information for reaches without connections.

```
    do i=1,nrch
      reach => this%GetReach(i)
      if (.not. reach%KeepReachActive) cycle
      ncon = size(reach%iconn)
      write(line,70)reach%NewReachNum
      if (ncon > 0) then
        do j=1,ncon
          write(ic,80)reach%iconn(j)
          line = trim(line) // ' ' // ic
        enddo
      endif
      write(iu,60)trim(line)
    enddo
```
instead of 
```
    do i=1,nrch
      reach => this%GetReach(i)
      if (.not. reach%KeepReachActive) cycle
      ncon = size(reach%iconn)
      if (ncon > 0) then
        write(line,70)reach%NewReachNum
        do j=1,ncon
          write(ic,80)reach%iconn(j)
          line = trim(line) // ' ' // ic
        enddo
        write(iu,60)trim(line)
      endif
    enddo
```
This change solved the issue for the Williston Basin NWT model of USGS SIR2017-5158.

**Environment**
This issue should be environment independent.

PR #2394, opened 2025-07-11
fix(SfrPackageWriter.f90) mf5to6 write CONNECTIONDATA for reaches with 0 connections

kzeiler:
Moved write() statements in the write_connectivity() subroutine of SfrPackageWriter.f90 outside the if statement checking whether each reach has >0 connections to write the appropriate CONNECTIONDATA block information for reaches without connections. Added item explaining fix in converter_mf5to6.tex. Resolves issue #2393.

### DIFF
--- a/doc/ConverterGuide/converter_mf5to6.tex
+++ b/doc/ConverterGuide/converter_mf5to6.tex
@@ -287,7 +287,7 @@ This section describes changes introduced into the MODFLOW 6 converter with each
 \item
 \currentmodflowversion
 \begin{itemize}
-  \item None.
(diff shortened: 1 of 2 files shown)


christianlangevin (approved):
Thanks for this, @kzeiler.  Presumably you have an example where this bug causes problems and your fix solves it.  I think this is fine to go in without a dedicated test.  Good to keep the converter limping along.

kzeiler:
My example was the Williston Basin NWT model, which is relatively large, and this change fixed the issue for the 50 SFR reaches without connection to other reaches. I didn't see that there were any tests in the autotest directory for mf5to6, so I didn't create a dedicated small test.

wpbonelli:
Yeah, the converter is just tested on the models here https://github.com/MODFLOW-ORG/modflow6-testmodels/tree/master/mf5to6, I don't think there are any handwritten tests for it.

PR #2412, opened 2025-07-23
chore(dfns): use consistent array syntax, update dfn spec

wpbonelli:
Use parentheses for array shape in `utl-ts.dfn`. Pointed out by @martclanor in https://github.com/MODFLOW-ORG/modflow-devtools/pull/229#issuecomment-3105295416.

Clarify also how the `shape` attribute can be used for array or string variables in the DFN specification. E.g., in `utl-tas.dfn`, the `sfacval` variable's `shape time_series_name` is validly interpreted as is, to indicate that `sfacval` has the same shape as `time_series_name`.


martclanor:
I've noticed a few additional similar patterns that you might want to consider including, in case they're valid:

- `shape any1d` -> `shape (any1d)`
- `shape lenbigline` -> `shape (lenbigline)`
- `shape (:)` -> `shape`
- `shape (unknown)` -> `shape`

Also, just like for other fields in the DFN, would it be a good idea to remove the line for `shape` when it has no value?

wpbonelli:
Thanks @martclanor. I like the idea of removing shape for non-array variables, and leaving `shape` with no value for unknown-shape variables. But right now I think empty `shape` is interpreted widely by our tooling to mean "not an array". So, taking these one by one

- `shape any1d` yes, should be fixed to `shape (any1d)`
- `shape lenbigline` is correct as is since it specifies string length not array shape
- `shape (:)` should probably become `shape (any1d)`?
- `shape (unknown)` should probably stay for now, until we've mapped out a potential move to just `shape` (empty) to indicate unknown shape, and no `shape` meaning not an array

mjreno (approved):

PR #2421, opened 2025-07-28
fix(par): memory exception when writing outer convergence info to csv

mjr-deltares:
The exception is caused by mixing up the model index in the solution versus the index in the global model list (= model id). In serial they are the same and the problem cannot occur.


emorway-usgs (approved):
Seems like a nice simplification to the code and if it works with Kyle Davis's Great Salt Lake model, this seems ready to go in.

PR #1901, opened 2024-06-26
feat(prp): support coordinate transformations

wpbonelli:
PRT requires particle release points (except for z via `LOCAL_Z`) specified in the model coordinate system, and also reports pathlines in the model coordinate system. This PR enables input coordinate transformations with keyword options for the particle release point package.

With this change, points may be specified as

- model coordinates (default, as before)
- local cell coordinates scaled to the unit interval with `LOCAL_XY` and/or `LOCAL_Z` &mdash; only structured grids
- offsets from the cell center, no rescaling, via `LOCAL_XY_OFFSET` &mdash; structured or unstructured grids
- global coordinates via dev option `DEV_GLOBAL_XY` &mdash; transform release points to model coordinates using grid georeference information

___

Checklist of items for pull request

- [x] Added new test or modified an existing test
- [x] Formatted new and modified Fortran source files with `fprettify`
- [x] Added doxygen comments to new and modified procedures
- [x] Updated [definition files](/MODFLOW-USGS/modflow6/tree/develop/doc/mf6io/mf6ivar)
- [x] Updated [develop.tex](/MODFLOW-USGS/modflow6/doc/ReleaseNotes/develop.tex) with a plain-language description of the bug fix, change, feature; required for changes that may affect users
- [x] Updated [input and output guide](/MODFLOW-USGS/modflow6/doc/mf6io)
- [x] Removed checklist items not relevant to this pull request

PR #2424, opened 2025-07-30
feat(prt): make release coordinate checks optional

wpbonelli:
While it's on the user to provide both release coordinates and cell IDs, for some time we have checked at release time that release coordinates fall within the specified cell. Alternative ways to specify release points are in the works in #1901 e.g. local coordinates, which would remove the redundancy, but so long as release points are in global coordinates and we don't do cell identification ourselves, the coordinate checks are a performance bottleneck when particle count is high.

Add a `COORDINATE_CHECK_METHOD` option to allow opting out. Checks on (by default) with `eager`, trust me I know what I'm doing mode with `none`.

Later, consider adding a `lazy` mode where validation happens at tracking time, this might save some time if we can reuse terms of other calculations for the checks, but contradicts "fail early" philosophy so should probably not become default. Need to test and get some performance numbers.

If in future we did start computing cell IDs (e.g. with something fast like the [celltree](https://github.com/NOAA-ORR-ERD/cell_tree2d) algorithm) this option could just be ignored when cell IDs are not provided by the user. So one could decide whether to do it in preprocessing or let PRT handle it.

--- 

### DIFF
--- a/autotest/test_prt_release_coord_check.py
+++ b/autotest/test_prt_release_coord_check.py
@@ -1,0 +1,217 @@
+"""
+Test coordinate checking functionality for PRT release points.
+
+Tests the COORDINATE_CHECK_METHOD option which can be set to:
+- 'eager' (default): check at release time
+- 'none': skip release coordinate validation
+
+Two test cases:
+1. checks on ('eager'): Should catch mismatching points/cellids with informative error
+2. no checks ('none'): Should allow mismatches and let tracking proceed. This produces
+incorrect pathlines where particles jump to the specified cells after reporting initial
+positions at the release coordinates.
+"""
+
+import flopy
+import matplotlib.cm as cm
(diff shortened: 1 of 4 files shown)


aprovost-usgs (commented):
Looks good.  Excellent summary of the related considerations in the Conversation, btw

martclanor:
I just noticed that the `default` attribute in the DFN is probably meant to be `default_value`. Unless maybe this is intended?

wpbonelli:
@martclanor thanks, this was a mistake. It's `default` in the new schema https://github.com/MODFLOW-ORG/modflow-devtools/blob/3ef75eb78a3496cdca3b29df7ec79331c4743a62/modflow_devtools/dfn/parse.py#L148 but the DFN file should still use `default_value`, I guess I was living in the future 😅

PR #2435, opened 2025-08-03
feat(prt): add developer option for cycle detection

wpbonelli:
Cyclic particle pathlines may be legitimate over multiple time steps, but within a given application of the tracking method i.e. in a single time step they so far reliably indicate a programming error.

Add a developer option `DEV_CYCLE_DETECTION_WINDOW` to PRT PRP enabling cycle detection. Disabled (set to 0) by default. When enabled, look at the last `DEV_CYCLE_DETECTION_WINDOW` events and terminate the simulation with error if a cycle is found. A cycle is defined as a duplicate cell exit (i.e. through the same face) within a single application of the tracking method.

For a speed improvement we could keep a `HashTable` alongside the list of recent events, where the key is the event string, then just a single lookup instead of iterating over the full window. But since this is meant for infrequent use as a developer debugging tool not a user option, and the window will be small, I will punt performance optimization for now.

Consequentially, this PR also makes it the responsibility of the event firer to deallocate the event when it's no longer needed. The event firing routines generally now deallocate the event after dispatch/handling. When cycle detection is off, the cell exit event firing routine does the same. When cycle detection is on, events cannot be deallocated until they are popped from the history, so they are cleaned up at that time by the list node's value destruction mechanism.

___


doc/ReleaseNotes/develop.toml
```diff
@@ -51,7 +51,7 @@ description = "Fixed a memory access exception that can occur when writing conve
 [[items]]
 section = "features"
 subsection = "solution"
-description = "Added an optional new COORDINATE\\_CHECK\\_METHOD option to the PRT model's Particle Release Package, accepting values 'EAGER' (default) or 'NONE'. 'NONE' disables release-time verification that release point coordinates and release point cell IDs match, which can significantly reduce runtime when releasing many particles."
```
wpbonelli, line 54: retrospective fix, just noticed that we don't use quotes around option values in release notes

PR #2438, opened 2025-08-04
fix(prt): distinguish cell exit from subcell exit events

wpbonelli:
We currently define the PRT OC package's `TRACK_EXIT` keyword option as enabling reporting when particles exit any grid feature (model, cell or subcell), and report all such events with event code (i.e. reason code) 1. Thus subcell exits in the ternary method appear identical in output to cell transitions, despite the fact that subcells are an implementation detail of the ternary method and occur in the interior of user-defined grid cells.

Distinguish subcell exit events from cell exits. Introduce a new PRT OC option `TRACK_SUBFEATURE_EXIT` which applies to all grid features below the level of the user-defined discretization (currently only subcell exits in DISV grids, but this could apply later to e.g. subreaches in surface water models, etc). The old `TRACK_EXIT` option now only applies to grid-scale features and above, i.e. cells (and models once we implement particle exchanges).

While this changes the results reported by PRT under the default settings (see changed test snapshots for instance &mdash; fewer path points, now only on cell faces, which nicely matches mp7 when ternary is forced on a structured grid) we think that since subgrid features are not likely to be of interest to most users, this is reasonable to consider a fix, not a breaking change.

This PR also moves exit event firing routines to their respective method types, following on #2433.

#### Rejected alternatives

We aborted/punted some alternative ideas to resolve this:

<details>
<summary><h5>Feature-specific event codes and OC options</h5></summary>
Introduce a new particle event code 6 for subcell exits, and new PRT OC options `TRACK_CELLEXIT` and `TRACK_SUBCELLEXIT` toggling reporting for the respective events. `TRACK_EXIT` remains and continues to work as defined, enabling exit reporting for all features, but becomes disabled default. Counterargument: proliferating reason codes and OC options like this (i.e., one for each specific sub-grid-scale feature type, for all possible grid types) seems not ideal.
</details>

<details>
<summary><h5>New output column for domain level</h5></summary>
Consider introducing a new output column indicating the domain level at which a feature exit event occurs: model = 1, cell (or equivalent e.g. reach) = 2, subgrid features = 3. This would allow keeping the same reason code for all exit events while giving the ability to distinguish them. Counterargument: pretty big change, think about it some more.
</details>

___

- [x] Referenced issue or pull request #2435


aprovost-usgs:
Haven't looked at the PR closely yet - just some initial thoughts
- Good idea to not output subcell exits by default and to provide a separate control for that
- "Cell" and "subcell" can be seen as specific to tracking through a grid. For now that's all PRT can do, but someday when there's tracking through, say, streams, those terms might not apply naturally. Some alternatives are (1) take a liberal view of what "cell" and "subcell" mean in other contexts and explain that in the documentation, (2) structure things so that other terms (like "reach" and "subreach" - just making that up) are used in other contexts, or (3) use generic terms that cover all contexts (though can't think of good ones right now)

aprovost-usgs:
... In advanced packages, the analogs of cells (e.g., stream reaches) are called 'features,' and the current description of TRACK_EXIT refers to features, though they include tracking domains at all levels:

TRACK_EXIT—keyword to indicate that particle tracking output is to be written when a particle exits a feature (a
model, cell, or subcell)

wpbonelli:
> ... In advanced packages, the analogs of cells (e.g., stream reaches) are called 'features,' and the current description of TRACK_EXIT refers to features, though they include tracking domains at all levels:

thanks good point, I changed the wording in the description to refer to "features"

> "Cell" and "subcell" can be seen as specific to tracking through a grid. For now that's all PRT can do, but someday when there's tracking through, say, streams, those terms might not apply naturally. Some alternatives are (1) take a liberal view of what "cell" and "subcell" mean in other contexts and explain that in the documentation, (2) structure things so that other terms (like "reach" and "subreach" - just making that up) are used in other contexts, or (3) use generic terms that cover all contexts (though can't think of good ones right now)

Maybe the existing `TRACK_EXIT` sets us up nicely as a generic term, as it can apply to all of them, then `TRACK_CELLEXIT` and `TRACK_SUBCELLEXIT` could stay specific and we could introduce `TRACK_REACH` and so on as needed later? Avoiding the need to redefine "cell" "subcell" etc. Any option which doesn't apply to a simulation could be ignored (maybe with a printed warning)

wpbonelli:
Updated after review/discussion. Removed tentative new `TRACK_CELLEXIT` and `TRACK_SUBCELL_EXIT` options and added a more generic `TRACK_SUBFEATURE_EXIT` option allowing opt-into any sub-grid-scale particle transition events. `TRACK_EXIT` applies only to grid-scale features and above after this. No new reason codes here either, so pending a new output column identifying domain level, users opting into subfeature transition reporting will still not be able to distinguish cell exits from subcell exits. But this was not possible before now either.

doc/ReleaseNotes/develop.toml
```diff
@@ -61,4 +61,9 @@ description = "Add DEPENDENT\\_VARIABLE\\_SCALING option to scale the dependent
...
\ No newline at end of file
+description = "The Particle Tracking (PRT) Model previously reported cell exit events when a dry particle stranded in a partially saturated cell was dropped to the water table, as occurs with DRY\\_TRACKING\\_METHOD DROP (the default). An event is no longer reported at this time."
+
+[[items]]
+section = "fixes"
+subsection = "solution"
+description = "Previously the Particle Tracking (PRT) Model's generalized DISV tracking method did not distinguish subcell exit events from cell exits, reporting both with the same reason code (1). Since subcell exits occur at a scale below that defined by the model's grid discretization, they are an implementation detail of the generalized tracking method, and will therefore no longer be written by default to particle tracking output. The PRT Output Control (OC) package's TRACK\\_EXIT option will now apply exclusively to features at grid-scale or above. Particle transitions between sub-grid-scale features are now opt-in with a new keyword option TRACK\\_SUBFEATURE\\_EXIT."
```
aprovost-usgs: suggest "Writing particle transitions between sub-grid-scale features to output is now opt-in ..."

src/Solution/ParticleTracker/Particle/ParticleEvent.f90
```diff
@@ -17,10 +18,10 @@ module ParticleEventModule
   !<
   enum, bind(C)
     enumerator :: RELEASE = 0 !< particle was released
-    enumerator :: CELLEXIT = 1 !< particle exited a cell
+    enumerator :: FEATEXIT = 1 !< particle exited a grid feature
     enumerator :: TIMESTEP = 2 !< time step ended
     enumerator :: TERMINATE = 3 !< particle terminated
-    enumerator :: WEAKSINK = 4 !< particle exited a weak sink
+    enumerator :: WEAKSINK = 4 !< particle entered a weak sink
```
aprovost-usgs, line 24: I seem to recall we had a conversation about whether to report entering or exiting a weak sink.  Think one of our colleagues was interested in knowing when a particle enters a weak sink, even if it's not captured.  Is that what code 4 is intended to do, by noting entrance to a weak sink?  Is there a way to distinguish if it's captured?  Sorry if I'm rehashing something that's been previously settled

aprovost-usgs (approved):

src/Solution/ParticleTracker/Particle/ParticleEvent.f90
wpbonelli: Yes, we report on entry to a weak sink. (During the assessment step after cell load but before tracking starts.) For a while it was erroneously described as upon exit from a weak sink, I fixed that in the docs, but missed this comment.

wpbonelli: If it's captured, there should be a termination event following the weak sink event.

wpbonelli: Sorry, first answer was wrong.. If `stop_at_weak_sink` is on, the particle will terminate with status 4. If off, there will be a weak sink event with status 1, reason 4. In the latter case it might still terminate in that cell later, I think, if a cycle is detected between top/bottom faces i.e. the bottom of a well.

aprovost-usgs: Sounds good - thanks for checking

PR #2446, opened 2025-08-10
fix(prt): account for assigned boundary faces

wpbonelli:
The particle tracking method "sees" boundary cell flows but not boundary faces. When boundary faces are assigned internal to the model, i.e. an `IFLOWFACE` is adjacent to another active cell, this can lead to cycles, as the two adjacent cells do not agree on the flow through their shared face. 

MODPATH 7 avoids this by ignoring internal boundary faces, treating the cell as if they were not assigned.

Add FMI option `INTERNAL_IFLOWFACE_METHOD` determining what to do with particles in cells with an `IFLOWFACE` boundary assigned to an "internal" face (facing another active cell).

- `STOP` at internal assigned boundary face(s) after tracking through the cell, the default
- `IGNORE` internal assigned boundary face(s) like MP7 and issue a warning.

Boundary faces are stored in an integer bitmask which with 32 slots can fit 30 lateral faces, way above the current limit of 10, but the implementation may need to change if face count restriction is ever relaxed.

___

### DIFF
--- a/autotest/test_prt_iflowface_stopzone.py
+++ b/autotest/test_prt_iflowface_stopzone.py
@@ -2,7 +2,7 @@
 1D unconfined model with a constant head boundary on the left,
 5 RIV boundary cells on the right with the same stage, and recharge.
 
-Particles are placed in every cell center. There are several cases
+Particles are placed at every cell center. There are several cases
 with extended tracking off, and several identical cases with it on.
 
 The cases exercise different combinations of IFLOWFACE, either none
--- a/src/Model/ParticleTracking/prt-fmi.f90
+++ b/src/Model/ParticleTracking/prt-fmi.f90
@@ -1,13 +1,15 @@
 module PrtFmiModule
 
-  use KindModule, only: DP, I4B
-  use ConstantsModule, only: DZERO, LENAUXNAME, LENPACKAGENAME
-  use SimModule, only: store_error
-  use SimVariablesModule, only: errmsg
+  use KindModule, only: DP, I4B, LGP
+  use ErrorUtilModule, only: pstop
+  use ConstantsModule, only: DZERO, LENAUXNAME, LENPACKAGENAME, LENVARNAME
+  use SimModule, only: store_error, store_warning
+  use SimVariablesModule, only: errmsg, warnmsg
   use FlowModelInterfaceModule, only: FlowModelInterfaceType
   use BaseDisModule, only: DisBaseType
   use BudgetObjectModule, only: BudgetObjectType
   use CellModule, only: MAX_POLY_CELLS
+  use CellDefnModule, only: CellDefnType
 
   implicit none
   private
@@ -18,16 +20,23 @@ module PrtFmiModule
 
   type, extends(FlowModelInterfaceType) :: PrtFmiType
 
-    double precision, allocatable, public :: SourceFlows(:) ! cell source flows array
-    double precision, allocatable, public :: SinkFlows(:) ! cell sink flows array
-    double precision, allocatable, public :: StorageFlows(:) ! cell storage flows array
-    double precision, allocatable, public :: BoundaryFlows(:) ! cell boundary flows array
+    integer(I4B), pointer :: iffmeth => null() !< internal iflowface method: 0 = stop, 1 = ignore
+    real(DP), allocatable, public :: SourceFlows(:) ! cell source flows
+    real(DP), allocatable, public :: SinkFlows(:) ! cell sink flows
+    real(DP), allocatable, public :: StorageFlows(:) ! cell storage flows
+    real(DP), allocatable, public :: BoundaryFlows(:) ! cell boundary flows
+    integer(I4B), allocatable, public :: BoundaryFaces(:) ! bitmask of cell iflowface, 0-30: lateral, 31 (-2): bottom, 32 (-1): top
 
   contains
 
     procedure :: fmi_ad
     procedure :: fmi_df => prtfmi_df
+    procedure :: allocate_scalars => prtfmi_allocate_scalars
+    procedure :: source_options => prtfmi_options
     procedure, private :: accumulate_flows
+    procedure :: mark_boundary_face
+    procedure :: is_boundary_face
+    procedure :: validate_boundary_faces
 
   end type PrtFmiType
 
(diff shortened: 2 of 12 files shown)


autotest/test_prt_iflowface_stopzone.py
```diff
@@ -1,3 +1,18 @@
+"""
+1D unconfined model with a constant head boundary on the left,
+5 RIV boundary cells on the right with the same stage, and recharge.
+
+Particles are placed in every cell center. There are several cases
```
aprovost-usgs: "... placed at every cell center ..."?

src/Model/ParticleTracking/prt-fmi.f90
```diff
@@ -140,18 +148,72 @@ subroutine prtfmi_df(this, dis, idryinactive)
...
+
+    ! If GWF-Model exchange is active, setup flow terms
+    if (.not. this%flows_from_file) then
+      call this%initialize_gwfterms_from_gwfbndlist()
+    end if
+
+    ! Set flag that stops dry flows from being deactivated in a GWE
+    ! transport model since conduction will still be simulated.
+    ! 0: GWE (skip deactivation step); 1: GWT (default: use existing code)
+    this%idryinactive = idryinactive
```
aprovost-usgs, line 195: holdover from GWE - wonder if it's relevant to PRT. suspect it is but haven't thought it through. side issue

src/Model/ParticleTracking/prt-fmi.f90
```diff
@@ -196,10 +258,12 @@ subroutine accumulate_flows(this)
         iflowface = 0
         if (iauxiflowface > 0) then
           iflowface = NINT(this%gwfpackages(ip)%auxvar(iauxiflowface, ib))
+          if (this%iffmeth /= 1 .and. iflowface > 0) &
```
aprovost-usgs, line 260: with the introduction of the bitmask, there are three related face numberings -- let's call them IN, END, and BIT:
*  IN -- numbering used for IFLOWFACE input, with bottom and top being -2 and -1, respectively
*   END -- like IN, but with bottom and top moved to the end, i.e., MAX_POLY_CELLS-1 and MAX_POLY_CELLS, respectively
*   BIT -- like END, but with top and bottom at bits 30 and 31, respectively

coming into this conditional, IFLOWFACE is following the IN convention.  when mark_boundary face is called, it converts from the IN convention to BIT when storing the face in the bitmask.  later, when, say, MethodDis uses the bitmask in its pass_dis routine, it converts from END to IN (using iface_to_iflowface) and then to BIT in the call to is_boundary face.  seems like it would be possible to eliminate nearly all the conversions between conventions by converting IN to END right off the bat and then essentially using END throughout.  this would mean storing bottom and top in bits MAX_POLY_CELLS-1 and MAX_POLY_CELLS, respectively.

a few side issues:
*  it looks like bottom and top, which have negative IFLOWFACE at this point, will never be marked as boundary faces, which doesn't seem right -- what's proposed above should solve this problem, since bottom and top would then have positive IFLOWFACE here
*  also looks like this logic doesn't distinguish between internal and external boundary faces

src/Model/ParticleTracking/prt-fmi.f90
```diff
@@ -196,10 +258,12 @@ subroutine accumulate_flows(this)
...
+            call this%mark_boundary_face(i, iflowface)
           ! this maps bot -2 -> 9, top -1 -> 10; see note re: max faces below
```
aprovost-usgs: update this comment to reference MAX_POLY_CELLS

also, i think the load_boundary_flows_to_defn* procedures in MethodDis and MethodDisv should be updated -- they hardwire bottom and top at indices 9 and 10, respectively.  note that besides using indices of ioffset+9 and ioffset+10, ioffset itself is set assuming a stride of 10.  does this go bad if MAX_POLY_CELLS is set to other than 10?

src/Model/ParticleTracking/prt-fmi.f90
```diff
@@ -196,10 +258,12 @@ subroutine accumulate_flows(this)
...
           if (iflowface < 0) iflowface = iflowface + MAX_POLY_CELLS + 1
```
aprovost-usgs, line 263: this is the conversion from IN to END that could be done right away (see comment above)

src/Model/ParticleTracking/prt-fmi.f90
```diff
@@ -196,10 +258,12 @@ subroutine accumulate_flows(this)
...
         end if
-        if (iflowface .gt. 0) then
+        if (this%iffmeth /= 1 .and. iflowface > 0) then
```
aprovost-usgs, line 265: looks like this logic doesn't distinguish between internal and external boundary faces -- will IFACEFLOW be ignored on external boundary faces, as well, if iffmeth==1 ?

src/Model/ParticleTracking/prt-fmi.f90
```diff
@@ -140,18 +148,72 @@ subroutine prtfmi_df(this, dis, idryinactive)
...
+    ! Set flag that stops dry flows from being deactivated in a GWE
+    ! transport model since conduction will still be simulated.
+    ! 0: GWE (skip deactivation step); 1: GWT (default: use existing code)
+    this%idryinactive = idryinactive
+
     ! Allocate arrays
     allocate (this%StorageFlows(this%dis%nodes))
     allocate (this%SourceFlows(this%dis%nodes))
     allocate (this%SinkFlows(this%dis%nodes))
     allocate (this%BoundaryFlows(this%dis%nodes * MAX_POLY_CELLS))
```
aprovost-usgs, line 201: side issue -- instead of using the MAX_POLY_CELLS parameter from CellModule, which needs to be available early, when the method pool is being created, it seems like we could have the fmi determine the actual max number of cell faces from the grid info and use that for dimensioning BoundaryFlows and indexing the bottom and top faces, saving storage.  could add a get_npolyverts procedure (which returns the just the number of vertices, rather than the vertices themselves) to the discretization packages to support this

aprovost-usgs (commented):
happy to chat if you think it would be helpful

src/Model/ParticleTracking/prt-fmi.f90
wpbonelli: oops, missed when copy-pasting.. removed GWE mentions and left note to come back and consider relevance to PRT

src/Model/ParticleTracking/prt-fmi.f90
wpbonelli: updated comment here and routines in `MethodDis/v` to use `MAX_POLY_CELLS`.

> note that besides using indices of ioffset+9 and ioffset+10, ioffset itself is set assuming a stride of 10. does this go bad if MAX_POLY_CELLS is set to other than 10?

I think we're good now that the stride is itself `MAX_POLY_CELLS`? or maybe I'm missing something, and there are other places we assume stride 10

aprovost-usgs: the question "does this go bad ...?" was meant to refer to all the instances of using hardwired "9" and "10" mentioned in that paragraph.  sorry for the confusion -- probably seems like it refers only to the ioffset issue.  in any case, it looks like you fixed all the instances, at least all the ones i could find

src/Model/ParticleTracking/prt-fmi.f90
wpbonelli: working on this separately.

aprovost-usgs: good idea -- first get MAX_POLY_CELLS everywhere it needs to be, then replace it with npolyvertsmax where appropriate

PR #2462, opened 2025-08-24
refactor(prt): rename dry_tracking_method stay to strand

wpbonelli:
Deprecate `DRY_TRACKING_METHOD STAY` in favor of `STRAND`. Plan to support `STAY` until 6.8. Discussed with @aprovost-usgs. Supports forthcoming patch for edge case induced cyclic trajectories in partially saturated cells, which will introduce a new stranding event.

Since this is just a renaming of a valid option for a string variable, not a renaming of the variable itself, it won't show up in the deprecations docs.

Also some misc cleanup/tidying in PRP


wpbonelli:
decided against this renaming

PR #2472, opened 2025-08-28
feat(dis): add get_npolyverts and get_max_npolyverts functions

wpbonelli:
Add two new functions to the discretization types, for counting cell vertices and getting the maximum cell vertex count in the grid, respectively. Use the latter in PRT to avoid hard-coding an upper limit and save space on array allocations. Followup to #2470, originally suggested in #2446.

### DIFF
--- a/src/Model/Discretization/Dis2d.f90
+++ b/src/Model/Discretization/Dis2d.f90
@@ -50,6 +50,8 @@ module Dis2dModule
     procedure :: supports_layers
     procedure :: get_ncpl
     procedure :: get_polyverts
+    procedure :: get_npolyverts
+    procedure :: get_max_npolyverts
     procedure :: connection_vector
     procedure :: connection_normal
     ! -- private
@@ -1138,6 +1140,21 @@ contains
       polyverts(:, nverts + 1) = polyverts(:, 1)
     !
   end subroutine
+
+  !> @brief Get the number of polygon vertices.
+  function get_npolyverts(this, ic) result(npolyverts)
+    class(Dis2dType), intent(inout) :: this
+    integer(I4B), intent(in) :: ic
+    integer(I4B) :: npolyverts
+    npolyverts = 4
+  end function get_npolyverts
+
+  !> @brief Get the maximum number of polygon vertices.
+  function get_max_npolyverts(this) result(max_npolyverts)
+    class(Dis2dType), intent(inout) :: this
+    integer(I4B) :: max_npolyverts
+    max_npolyverts = 4
+  end function get_max_npolyverts
 
   !> @brief Read an integer array
   !< TODO: REMOVE?
--- a/src/Model/Discretization/Dis2d.f90
+++ b/src/Model/Discretization/Dis2d.f90
@@ -1142,18 +1142,26 @@ contains
   end subroutine
 
   !> @brief Get the number of polygon vertices.
-  function get_npolyverts(this, ic) result(npolyverts)
+  function get_npolyverts(this, ic, closed) result(npolyverts)
     class(Dis2dType), intent(inout) :: this
     integer(I4B), intent(in) :: ic
+    logical(LGP), intent(in), optional :: closed !< whether to close the polygon, duplicating a vertex
     integer(I4B) :: npolyverts
     npolyverts = 4
+    if (present(closed)) then
+      if (closed) npolyverts = 5
+    end if
   end function get_npolyverts
 
   !> @brief Get the maximum number of polygon vertices.
-  function get_max_npolyverts(this) result(max_npolyverts)
+  function get_max_npolyverts(this, closed) result(max_npolyverts)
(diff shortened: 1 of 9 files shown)


src/Solution/ParticleTracker/Method/MethodDis.f90
```diff
@@ -512,9 +512,11 @@ subroutine load_boundary_flows_to_defn(this, defn)
     class(MethodDisType), intent(inout) :: this
     type(CellDefnType), intent(inout) :: defn
     ! local
+    integer(I4B) :: max_faces
     integer(I4B) :: ioffset
 
-    ioffset = (defn%icell - 1) * MAX_POLY_VERTS
+    max_faces = this%fmi%dis%get_max_npolyverts() + 2
```
aprovost-usgs: since get_max_npolyverts involves looping over all cells, could fmi calculate max_faces once and store it?  then this would become
max_faces = this%fmi%max_faces

src/Solution/ParticleTracker/Method/MethodDisv.f90
```diff
@@ -522,16 +520,14 @@ subroutine load_boundary_flows_to_defn_poly(this, defn)
...
-    integer(I4B) :: npolyverts
-    integer(I4B) :: ioffset
-    integer(I4B) :: iv
+    integer(I4B) :: ic, iv, ioffset, npolyverts, max_faces
 
     ic = defn%icell
     npolyverts = defn%npolyverts
-
-    ioffset = (ic - 1) * MAX_POLY_VERTS
+    max_faces = this%fmi%dis%get_max_npolyverts() + 2
```
aprovost-usgs: see earlier comment about storing fmi%max_faces

src/Model/ModelUtilities/DiscretizationBase.f90
```diff
@@ -685,7 +687,26 @@ subroutine get_polyverts(this, ic, polyverts, closed)
...
+  function get_npolyverts(this, ic) result(npolyverts)
+    class(DisBaseType), intent(inout) :: this
+    integer(I4B), intent(in) :: ic !< cell number (reduced)
+    integer(I4B) :: npolyverts
+    npolyverts = 0 ! suppress compiler warning
+    errmsg = 'Programmer error: get_npolyverts must be overridden'
+    call store_error(errmsg, terminate=.true.)
+  end function get_npolyverts
+
+  !> @brief Get the maximum number of cell polygon vertices.
```
aprovost-usgs, line 703: like that this function is in the discretization

src/Model/Discretization/Dis2d.f90
```diff
@@ -1139,6 +1141,21 @@ subroutine get_polyverts(this, ic, polyverts, closed)
     !
   end subroutine
 
+  !> @brief Get the number of polygon vertices.
+  function get_npolyverts(this, ic) result(npolyverts)
```
aprovost-usgs: should we add an optional "closed" argument for consistency with get_polyverts?  we won't use it in PRT, but perhaps somebody might.  if so, we'd add 1 to npolyverts in the case that the "closed" argument is present

src/Model/Discretization/Disv.f90
```diff
@@ -1466,6 +1468,33 @@ subroutine get_polyverts(this, ic, polyverts, closed)
...
+    integer(I4B), intent(in) :: ic !< cell number (reduced)
+    integer(I4B) :: npolyverts
+    ! local
+    integer(I4B) :: ncpl, icu, icu2d
+
+    ncpl = this%get_ncpl()
+    icu = this%get_nodeuser(ic)
+    icu2d = icu - ((icu - 1) / ncpl) * ncpl
+    npolyverts = this%iavert(icu2d + 1) - this%iavert(icu2d) - 1
+    if (npolyverts .le. 0) npolyverts = npolyverts + size(this%javert)
```
aprovost-usgs: i know this is from get_polyverts, but i'm not sure what actual scenario this is intended to address.  would be good to understand it so we can be sure it would never be a problem for PRT.  i'll try to figure it out

src/Solution/ParticleTracker/Method/MethodDis.f90
wpbonelli: cached `max_faces` in fmi

src/Model/Discretization/Dis2d.f90
wpbonelli: done

src/Model/Discretization/Disv.f90
wpbonelli: seems like this would handle the case where the last element of `iavert` is not `size(javert) + 1` but 1.. like it wraps around. but looking at the sparse matrix module that is not how it seems to work. so idk why I ever thought this was necessary before. dropped it.

PR #2478, opened 2025-08-30
fix(prt): prevent top exits from partially saturated cells

wpbonelli:
Cyclic particle pathlines were possible in Newton models with `DRY_TRACKING_METHOD DROP` (the default) due to numerically insignificant upward flows in partially saturated cells. This occurred because PRT did not distinguish the water table from the cell top, so a particle at the water table in a partially saturated cell could obtain an exit solution through the cell's top face, jump to the dry cell above, then drop back down, and so on.

Prevent particles from exiting through the top face of partially saturated cells. If a particle obtains an exit solution upwards through the water table, its behavior now depends on whether the cell top is an assigned boundary face, and the `DRY_TRACKING_METHOD` setting. If the cell top is an assigned boundary, or if `DRY_TRACKING_METHOD` is `STOP`, the particle will terminate. Otherwise it will stay in that location. (Future work may expand on this behavior; this is a minimal bug fix.)

The issue is reproducible on the Keating example, so I added no new tests here. I added a snapshot comparison to Keating to check the new behavior.

This fix allows removing the old ad hoc trap for cycles between vertical faces. Also some other misc cleanup.
___

- [x] Referenced #2468


aprovost-usgs (approved):
nice job, looks good

PR #2479, opened 2025-09-01
fix(par): specific discharge for parallel simulations

mjr-deltares:
When specific discharge is not calculated by NPF, it cannot be synchronized across parallel processes. This PR adds a flag to indicate if the data is available or not.

### DIFF
--- a/autotest/test_gwt_adv03_gwtgwt.py
+++ b/autotest/test_gwt_adv03_gwtgwt.py
@@ -1,0 +1,322 @@
+"""
+Test gwt-gwt coupling for advection without dispersion.
+"""
+
+import os
+
+import flopy
+import numpy as np
+import pytest
+from framework import TestFramework
+
+cases = ["adv03_gwtgwt"]
+scheme = "tvd"
+
+# solver settings
+nouter, ninner = 100, 300
+hclose, rclose, relax = 1e-6, 1e-6, 1.0
+
+g_delr = 1.0
+
+
+def get_gwf_model(sim, gwfname, gwfpath, modelshape, chdspd=None, welspd=None):
+    nlay, nrow, ncol, xshift, yshift = modelshape
+    delr = g_delr
+    delc = 1.0
+    top = 1.0
+    botm = [0.0]
+    strt = 1.0
+    hk = 1.0
+    laytyp = 0
+
+    gwf = flopy.mf6.ModflowGwf(
+        sim,
+        modelname=gwfname,
+        save_flows=True,
+    )
+
+    dis = flopy.mf6.ModflowGwfdis(
+        gwf,
+        nlay=nlay,
+        nrow=nrow,
+        ncol=ncol,
+        delr=delr,
+        delc=delc,
+        top=top,
+        botm=botm,
+        xorigin=xshift,
+        yorigin=yshift,
+    )
+
+    # initial conditions
+    ic = flopy.mf6.ModflowGwfic(gwf, strt=strt)
+
+    # node property flow
+    npf = flopy.mf6.ModflowGwfnpf(
+        gwf,
+        icelltype=laytyp,
+        k=hk,
+        save_specific_discharge=False,
+    )
+
+    # chd files
+    if chdspd is not None:
+        chd = flopy.mf6.modflow.mfgwfchd.ModflowGwfchd(
+            gwf,
+            # maxbound=len(c),
+            stress_period_data=chdspd,
+            save_flows=False,
+            pname="CHD-1",
+        )
+
+    # wel files
+    if welspd is not None:
+        wel = flopy.mf6.ModflowGwfwel(
+            gwf,
+            print_input=True,
+            print_flows=True,
+            # maxbound=len(w),
+            stress_period_data=welspd,
+            save_flows=False,
+            auxiliary="CONCENTRATION",
+            pname="WEL-1",
+        )
+
+    # output control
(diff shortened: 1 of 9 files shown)


src/Distributed/VirtualGwtModel.f90
```diff
@@ -136,15 +139,24 @@ subroutine vgwt_prepare_stage(this, stage)
...
+
+      nr_nodes = this%element_maps(MAP_NODE_TYPE)%nr_virt_elems
+      nr_conns = this%element_maps(MAP_CONN_TYPE)%nr_virt_elems
+
+      call this%map(this%fmi_gwfhead%base(), nr_nodes, (/STG_BFR_EXG_AD/))
+      call this%map(this%fmi_gwfsat%base(), nr_nodes, (/STG_BFR_EXG_AD/))
+      if (this%fmi_igwfspdis%get() > 0) then
+        call this%map(this%fmi_gwfspdis%base(), 3, nr_nodes, (/STG_BFR_EXG_AD/))
+      else
+        call this%map(this%fmi_gwfspdis%base(), 3, 0, (/STG_NEVER/))
```
Manangka, line 156: Just wondering, when fmi_igwfspdis indicates the variable is not active you set the size of the mapping to zero.  The logic of checkin if a variable is active happens more often in this file, for instance on line 130 with the cnd_idisp variable. However at the other places the mapped size isn't set to zero. No size is set at all.  What is the reason we do have to do it here?

src/Distributed/VirtualGwtModel.f90
```diff
@@ -136,15 +139,24 @@ subroutine vgwt_prepare_stage(this, stage)
...
       if (this%indsp%get() > 0 .and. this%inmst%get() > 0) then
         call this%map(this%mst_thetam%base(), nr_nodes, (/STG_AFT_CON_AR/))
       end if
 
+    else if (stage == STG_AFT_CON_AR) then
+
+      nr_nodes = this%element_maps(MAP_NODE_TYPE)%nr_virt_elems
+      nr_conns = this%element_maps(MAP_CONN_TYPE)%nr_virt_elems
+
+      call this%map(this%fmi_gwfhead%base(), nr_nodes, (/STG_BFR_EXG_AD/))
```
Manangka, line 151: Why are these calls moved from the `STG_BFR_CON_AR` to the `STG_AFT_CON_AR` stage?

src/Distributed/MpiMessageBuilder.f90
```diff
@@ -815,4 +826,68 @@ subroutine get_mpitype_for_dbl2d(mem, el_displ, el_type, el_map)
...
+  subroutine check_map_dbl2d(mem, map)
+    type(MemoryType), pointer :: mem
+    integer, dimension(:), pointer :: map
+    ! local
+    logical(LGP) :: is_valid
+    integer(I4B) :: min_val, max_val
+
+    if (.not. associated(map)) return ! nothing to check
+
+    min_val = minloc(map, dim=1)
```
Manangka: is `dim=1` necessary? The way I interpret the documentation it will return a single index if the array is 1D

src/Distributed/MpiMessageBuilder.f90
```diff
@@ -815,4 +826,68 @@ subroutine get_mpitype_for_dbl2d(mem, el_displ, el_type, el_map)
...
+    max_val = maxloc(map, dim=1)
+    is_valid = max_val <= size(mem%adbl2d, dim=2) .and. min_val > 0
```
Manangka: I'm not familiar with the structure of the `map`.  `maxloc` find the index of the maximum value in the map variable.  Because you use `maxloc` to find it and not take the last value of  `mem` I assume that   the `map` has some sort of padding at the end of the array? Or it isn't sorted?

autotest/test_gwt_adv03_gwtgwt.py
```diff
@@ -0,0 +1,319 @@
...
+            pname="CHD-1",
+        )
+
+    # wel files
+    if welspd is not None:
+        wel = flopy.mf6.ModflowGwfwel(
+            gwf,
+            print_input=True,
+            print_flows=True,
+            # maxbound=len(w),
```
Manangka: You can remove this

mjr-deltares: Done

src/Distributed/MpiMessageBuilder.f90
mjr-deltares: yes, from the GNU pages on the intrinsic: "If DIM is absent, the result is a rank-one array with a length equal to the rank of ARRAY"

src/Distributed/VirtualGwtModel.f90
mjr-deltares: To keep the fmi arrays together, I think it is more readable

src/Distributed/VirtualGwtModel.f90
mjr-deltares: good point, I think there is two different ideas that I followed over time so it is not consistent. I think I can refactor all and get rid of the 'else' part everywhere and initialize data items with STG_NEVER to be overwritten later, or not. I will do that in a different PR.

src/Distributed/MpiMessageBuilder.f90
mjr-deltares: (Alternatively, you can accept that the return value is an array and access the first element of it)

Manangka (approved):

PR #2503, opened 2025-09-12
fix(prt): ignore non-boundary upwards flow thru water table

wpbonelli:
#2478 introduced and #2496 revised special handling at the water table. This revises it again. Rather than immobilizing the particle and/or applying special termination conditions if it reaches a water table from below, "cap" cells containing a water table at zero upwards flow in the absence of an assigned top boundary face. This guarantees that a particle exiting through the top of a cell is carried not by negligible Newton flows but legitimate flow through an assigned boundary face.


aprovost-usgs (approved):
Looks good

PR #2499, opened 2025-09-11
feat(idm): support dfn prerelease attribute

mjreno:
Input support to mark specific parameters as prerelease and not available for use in releases:
  - define parameter `prerelease true` in dfn
  - only supports parameters but updating a required parameter in a package renders the package inaccessible (e.g. `readarraygrid` in the the grid based packages)
  - supported for all blocks in packages that are IDM integrated except `STO` and `PRT-PRP` `PERIOD` blocks (soon `OC`as well) 

Checklist of items for pull request

- [x] Replaced section above with description of pull request
- [X] Ran `ruff` on new and modified python scripts in .doc, autotests, doc, distribution, pymake, and utils subdirectories.
- [X] Formatted new and modified Fortran source files with `fprettify`
- [X] Updated meson files, makefiles, and Visual Studio project files for new source files
- [X] Updated [definition files](/MODFLOW-ORG/modflow6/tree/develop/doc/mf6io/mf6ivar)
- [X] Removed checklist items not relevant to this pull request

For additional information see [instructions for contributing](/MODFLOW-ORG/modflow6/.github/CONTRIBUTING.md) and [instructions for developing](/MODFLOW-ORG/modflow6/.github/DEVELOPER.md).

PR #2504, opened 2025-09-13
doc(mf6ivar): omit prerelease variables

wpbonelli:
Following on #2499, skip `prerelease` variables when building docs in the `mf6ivar.py` script


doc/mf6io/mf6ivar/mf6ivar.py
```diff
@@ -116,21 +116,6 @@
 #   output control rewritten entirely, and implemented in the code
 #
 
-# DEFINITION FILE KEYWORDS
```
wpbonelli, line 119: the [DFN spec](https://modflow6.readthedocs.io/en/latest/_dev/dfn.html) is the source of truth here

PR #2513, opened 2025-09-19
test(framework): fix original regression capability, clean the attic

wpbonelli:
fix `--original-regression`, it got broken somewhere along the way. remove some unneeded workspace setup functions and just use `copytree` instead. tidy up relevant bits and pieces of the test framework and dev docs


DEVELOPER.md
```diff
@@ -701,25 +701,21 @@ Tests using models from external repositories can be selected with the `external
...
-Large test models are excluded from commit-triggered CI and only run on GitHub Actions nightly. To run the models from a local clone of the repository:
+The test framework can also exercise the mf5to6 converter program. To do this, relax the namefile search pattern during model discovery:
 
 ```shell
-pixi run autotest -m "external" --models-path /path/to/modflow6-largetestmodels
+pixi run autotest -m "external" --models-path /path/to/modflow6-testmodels/mf5to6 --namefile-pattern "*.nam"
 ```
 
-Tests load external models from fixtures provided by `modflow-devtools`. External model tests can be selected by model or simulation name, or by packages used. See the [`modflow-devtools` documentation](https://modflow-devtools.readthedocs.io/en/latest/md/fixtures.html#filtering) for usage examples. Note that filtering options only apply to tests using external models, and will not filter tests defining models in code &mdash; for that, the `pytest` built-in `-k` option may be used.
+The framework will then convert any MF2005 models found to MF6 models and run them. This can be conveniently combined with `--original-regression` to compare MF2005 and MF6 results.
```
jdhughes-dev: Does it only convert mf2005 models or will it also convert mfnwt models? How about mfusg model?

autotest/test_external_models.py
```diff
@@ -109,20 +121,13 @@ def test_model(
         )
         assert success
 
-    setup_simulation(src=src_workspace, dst=mf6_workspace)
+    copytree(src_workspace, mf6_workspace)
     if (
         comparison := detect_comparison(tmp_path)
-        if original_regression
+        if original_regression or any(s in model_name for s in OG_REG)
```
jdhughes-dev, line 127: I tested with this change locally and worked perfectly. Great!!!

jdhughes-dev (approved):
Looks good to me. Just a question about the documentation.

DEVELOPER.md
wpbonelli: it also converts mfnwt and mflgr. updated the text here. I didn't realize the converter worked for nwt and lgr.

PR #2540, opened 2025-09-23
doc(mf6io; dfns): fix table break across pages; clarify table entries; fix dfn typo

aprovost-usgs:


wpbonelli (approved):

aprovost-usgs:
cancelled the re-checks because i reverted that commit

PR #2546, opened 2025-09-26
refactor(prt): introduce functions for face numbering schemes

wpbonelli:
It's tricky to get these conversions right, so pull them into functions.

There are three face numbering schemes to keep straight. Given `NLATFACES` is each cell's number of lateral faces and `NMAXFACES` is the maximum number of 3d faces (i.e. including top and bottom) of any cell in the grid,

- iflowface: described in mf6io and specified by the user, lateral `1..NLATFACES`, bottom `-2`, top `-1`
- iboundary: internal compressed numbering. lateral `1..NLATFACES`, bottom `NLATFACES + 2`, top `NLATFACES + 3`. the extra offset is because the first lateral face "wraps around". this scheme is used by the tracking method
- icellface: internal grid-aligned numbering. lateral `1..NLATFACES`, bottom `NMAXFACES - 1`, top `NMAXFACES`. this is the numbering used by PRT FMI and expected as input to assigned boundary face check functions

This is as described in https://github.com/MODFLOW-ORG/modflow6/pull/2446#discussion_r2298384910 except for icellface numbering being aligned to the grid's max face count instead of the bit mask size. The iboundary scheme is used in the tracking method since it aligns with the arrays holding cell geometry. The "grid-aligned" icellface scheme is used by FMI so callers can reliably specify top/bottom face to the boundary face check routines without knowing a particular cell's face count, just the grid maximum.

The new conversion functions are private for now but could be made public later if conversion is necessary elsewhere

### DIFF
--- a/src/Model/ParticleTracking/prt-fmi.f90
+++ b/src/Model/ParticleTracking/prt-fmi.f90
@@ -33,7 +33,7 @@ module PrtFmiModule
     procedure :: mark_boundary_face
     procedure :: is_boundary_face
     procedure :: is_net_out_boundary_face
-    procedure, private :: iflowface_to_iface
+    procedure, private :: iflowface_to_icellface
 
   end type PrtFmiType
 
@@ -155,7 +155,7 @@ contains
     class(PrtFmiType) :: this
     ! local
     integer(I4B) :: j, i, ip, ib
-    integer(I4B) :: ioffset, iflowface, iauxiflowface, iface
+    integer(I4B) :: ioffset, iflowface, iauxiflowface, icellface
     real(DP) :: qbnd
     character(len=LENAUXNAME) :: auxname
     integer(I4B) :: naux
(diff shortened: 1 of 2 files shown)


aprovost-usgs (changes requested):
good idea, and nicely done.  my only suggestion would be to consider renaming "iface", since that could be confused with MODPATH's iface.  maybe call it something like igridface or ifmiface, to reflect either that it's "grid aligned" or that it's used by the fmi.  those are just two possibilities - please feel free to choose something else.  all feel free to merge once you're happy with the name

wpbonelli:
Tentatively went with `icellface`

In the long run maybe we can use negative array indices as came up in discussion. and iflowface numbering everywhere but the bitmask. get rid of 1 transformation at least.

aprovost-usgs:
perfect

aprovost-usgs (approved):

PR #2563, opened 2025-10-10
refactor(prt-fmi): use negative indices for boundary flow array 2nd dim

wpbonelli:
Followup on #2546, convert PRT FMI boundary flows from flat array with fixed offset to 2D array with negative indices in the 2nd dimension (face numbers), indexed by iflowface numbering rather than by icellface. Simplifies the structured and vertex methods, no more need to compute the offset. Also define constants for top/bottom face numbers in iflowface numbering


wpbonelli:
closing in favor of #2586

PR #2565, opened 2025-10-11
docs(dfns): deprecate reader, infer from other attributes

wpbonelli:
The `reader` attribute for DFN variables seems unnecessary. Deprecate it and infer it in `mf6ivar.py` from `shape`/`type`. It is only used for generating documentation. Leave `reader` in the DFNs for now in case other people are using it. Also, the DFN spec listed old, outdated values like "u1ddbl" as still in use, clean that up

### DIFF
--- a/doc/mf6io/mf6ivar/mf6ivar.py
+++ b/doc/mf6io/mf6ivar/mf6ivar.py
@@ -200,6 +200,18 @@ MD_DIR_PATH.mkdir(exist_ok=True)
 TEX_DIR_PATH.mkdir(exist_ok=True)
 
 
+def infer_reader(v):
+    """
+    Infer the reader from the variable's attributes.
+    Arrays use 'readarray', everything else `urword`.
+    """
+    type_ = v.get("type", "")
+    has_shape = "shape" in v and v["shape"].strip() != ""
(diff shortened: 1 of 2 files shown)


wpbonelli:
@martclanor do you use `reader` for your plugin?

wpbonelli:
@mjreno am I right that only the docs use `reader`? Not IDM or the block parser or anything?

martclanor:
> @martclanor do you use `reader` for your plugin?

[Yes](https://github.com/martclanor/vscode-mf6-syntax/blob/90a88ecdb80223a7797f3d8d40ff9143a986a259/utils/gen_from_dfn.py#L163C5-L165C42), but it should be possible to adapt it in the same way to infer from shape and type instead. Thanks for considering!

mjreno:
> @mjreno am I right that only the docs use `reader`? Not IDM or the block parser or anything?

IDM doesn't use the reader, I believe it is docs only but I'll circle back on this one tomorrow.

mjreno (approved):
The terminology is in various places like InputOutput.f90 but I don't think there is any dependence on the dfn fields other than what you've noted.  Looks good to me.

PR #2586, opened 2025-10-29
refactor(prt): simplify face numbering scheme

wpbonelli:
Followup on #2546, convert PRT FMI boundary flows from flat array with fixed offset to 2D array with 2nd dimension face numbers. The 2nd dimension uses `icellface` numbering, the internal "grid-aligned" scheme which shifts `iflowface` negative top and bottom to the end of an interval from 0 to (grid max lateral cell faces + 2).

This simplifies the structured and vertex methods, no more need to compute an offset. Also define constants for top/bottom face numbers in the iflowface numbering scheme.

### DIFF
--- a/src/Model/ParticleTracking/prt-fmi.f90
+++ b/src/Model/ParticleTracking/prt-fmi.f90
@@ -41,7 +41,6 @@ module PrtFmiModule
     procedure :: is_boundary_face
     procedure :: is_net_out_boundary_face
     procedure, private :: iflowface_to_icellface
-    procedure, private :: icellface_to_iflowface
 
   end type PrtFmiType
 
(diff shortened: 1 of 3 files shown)


src/Model/ParticleTracking/prt-fmi.f90
aprovost-usgs, line 1: where is function icellface_to_iflowface being used?

aprovost-usgs (commented):
Looks good, just that question about icellface_to_iflowface

src/Model/ParticleTracking/prt-fmi.f90
wpbonelli: it is not. just figured to add it for symmetry but doing so does violate the YAGNI principle.

aprovost-usgs: guess i'd be inclined to leave it out for that reason but don't feel strongly about it and defer to your judgment

aprovost-usgs (approved):

src/Model/ParticleTracking/prt-fmi.f90
wpbonelli: I'm persuaded, I will take it out.

Issue #2577, opened 2025-10-23
ESL file AUXMULT not working


kam2:
I'm working on a GWE model (mf6.6.3) that includes the ESL package. Although the AUXILIARY and AUXMULTNAME options are listed in the ESL package and the same string is provided with each, the GWE LST file stress period budget summary shows the SENERRATE variable is getting applied as-is without the <aux>multiplier. Adding the SAVE_FLOWS option in the ESL package corroborated that individual cells were receiving the raw SENERRATE values.

christianlangevin:
Hello @kam2, sorry to hear this is not working for you.  Is there any chance you have a small example that demonstrates the problem?  We will take a look.  Thanks.

emorway-usgs:
Fix submitted with #2597.  Apologies for the delay in getting to this.

emorway-usgs:
Closed with #2597

PR #2597, opened 2025-11-14
fix(gwe-esl): aux multiplier has hooks but no functionality

emorway-usgs:
Even though AUXILIARY and AUXMULTNAME keywords could be specified in the OPTIONS block of the ESL input file, the source code didn't do anything with them.  

- [x] Addresses issue #2577 

### DIFF
--- a/src/Model/GroundWaterEnergy/gwe-esl.f90
+++ b/src/Model/GroundWaterEnergy/gwe-esl.f90
@@ -32,7 +32,6 @@ module GweEslModule
     procedure :: bnd_da => esl_da
     procedure :: define_listlabel
     procedure :: bound_value => esl_bound_value
-    procedure :: ener_mult
     ! -- methods for observations
     procedure, public :: bnd_obs_supported => esl_obs_supported
     procedure, public :: bnd_df_obs => esl_df_obs
--- a/src/Model/GroundWaterEnergy/gwe-esl.f90
+++ b/src/Model/GroundWaterEnergy/gwe-esl.f90
@@ -150,7 +150,7 @@ contains
     integer(I4B) :: node
     ! -- formats
     character(len=*), parameter :: fmtenermulterr = &
-      "('DRN BOUNDARY (',i0,') ESL MULTIPLIER (',g10.3,') IS &
+      "('ESL BOUNDARY (',i0,') ESL MULTIPLIER (',g10.3,') IS &
       &LESS THAN ZERO')"
     !
     ! -- check stress period data
--- a/src/Model/GroundWaterEnergy/gwe-esl.f90
+++ b/src/Model/GroundWaterEnergy/gwe-esl.f90
@@ -2,6 +2,8 @@ module GweEslModule
   !
   use KindModule, only: DP, I4B
   use ConstantsModule, only: DZERO, DEM1, DONE, LENFTYPE
+  use SimVariablesModule, only: errmsg
+  use SimModule, only: count_errors, store_error, store_error_filename
   use BndExtModule, only: BndExtType
   use ObsModule, only: DefaultObsIdProcessor
   use GweInputDataModule, only: GweInputDataType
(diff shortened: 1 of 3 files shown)


src/Model/GroundWaterEnergy/gwe-esl.f90
```diff
@@ -286,4 +326,25 @@ function esl_bound_value(this, col, row) result(bndval)
...
+    ! -- modules
+    use ConstantsModule, only: DZERO
+    ! -- dummy variables
+    class(GweEslType), intent(inout) :: this !< BndExtType object
+    integer(I4B), intent(in) :: row
+    ! -- result
+    real(DP) :: ener
+    !
+    if (this%iauxmultcol > 0) then
+      ener = this%senerrate(row) * this%auxvar(this%iauxmultcol, row)
```
aprovost-usgs, line 341: Since you're using this%senerrate directly here, it looks like function esl_bound_value is no longer needed?

src/Model/GroundWaterEnergy/gwe-esl.f90
```diff
@@ -136,6 +140,39 @@ subroutine esl_allocate_arrays(this, nodelist, auxvar)
...
+  !<
+  subroutine esl_ck(this)
+    ! -- dummy
+    class(GweEslType), intent(inout) :: this
+    ! -- local
+    integer(I4B) :: i
+    integer(I4B) :: node
+    ! -- formats
+    character(len=*), parameter :: fmtenermulterr = &
+      "('ESL BOUNDARY (',i0,') ESL MULTIPLIER (',g10.3,') IS &
```
aprovost-usgs, line 153: Since this isn't a conductance (as in DRN and other places where such a multiplier is used), might someone legitimately want a negative multiplier?

aprovost-usgs (commented):
Just a couple of questions

src/Model/GroundWaterEnergy/gwe-esl.f90
emorway-usgs: It looks like `esl_bound_value()` was introduced with #2426 ([this link](https://github.com/MODFLOW-ORG/modflow6/pull/2426/files#diff-b17d5c905a26a7130f2d20c2fc43b6a13ff92d67216880a514dbdaed7e99ba49) shows the addition...scroll down just a bit).  Thus, I don't think we want to get rid of it because it will break IDM?  However, I do wonder how IDM handles `AUXMULTNAME`?  Perhaps we need to check with @mjreno to see if `esl_bound_value()` needs updating to account for the existance of a potential multiplier?

src/Model/GroundWaterEnergy/gwe-esl.f90
emorway-usgs: Hmm, great point.  I suppose someone might want it.  I can't speak to the legitimacy of it, but there's no reason to prevent, I suppose.  Adjusted the PR in response to this comment with 9fed24f

src/Model/GroundWaterEnergy/gwe-esl.f90
mjreno: The `bound_value()` routines are intended to be a generic way to access list style input values for the package especially from elsewhere in the code- using column and row indexes.  The base class uses these routines when logging package input values (the list table) so they should be defined, but internal to the package using pointers to the input context package arrays is more explicit.

aprovost-usgs: Thanks @mjreno, I see now that each boundary package has a bound_value => xxx_bound_value function defined that's needed by the base class BndExtType.

@emorway-usgs, what got me on this track was that ESL itself previously used bound_value to get the rate, i.e., q = this%bound_value(1, i).  This was replaced by q = this%ener_mult(i), and within function ener_mult I guess I expected to see the call to bound_value, but instead this%senerate is used directly.  Nothing wrong with this, just not what I expected (and I originally didn't see that esl_bound_value was in fact being used by the parent class)

mjreno: I noticed `ESL` isn't calling `write_list()` is that intentional?  Apologies if these gaps were created during the IDM integration- my intention generally is to do a straight translation based on what the package was doing pre-IDM. If it is intentional not to log the input then `bound_value()` isn't strictly needed (although probably best to in keeping with the pattern).  The package does support the `PRINT_INPUT` option.

mjreno: I'm thinking I missed the logging with the integration and it might make most sense to go into the base class.  I can look at that this week if that sounds right.

aprovost-usgs: @mjreno That would be great (as we discussed at the weekly meeting) - thanks!

emorway-usgs: @mjreno and I discussed keeping this PR scoped as is and dealing with the logging issue in a separate PR.

emorway-usgs: Per discussion, the `write_list()` issue will be dealt with in a dedicated PR to help keep this fix scoped as is.

PR #2598, opened 2025-11-14
fix(gwt-src): aux multiplier has hooks but no functionality

emorway-usgs:
Even though AUXILIARY and AUXMULTNAME keywords could be specified in the OPTIONS block of the SRC input file, the source code didn't do anything with them.

- [x] Similar to #2597 
- [x] Motivated by issue #2577 

### DIFF
--- a/src/Model/GroundWaterTransport/gwt-src.f90
+++ b/src/Model/GroundWaterTransport/gwt-src.f90
@@ -2,6 +2,8 @@ module GwtSrcModule
   !
   use KindModule, only: DP, I4B, LGP
   use ConstantsModule, only: DZERO, DEM1, DONE, LENFTYPE, LENVARNAME
+  use SimVariablesModule, only: errmsg
+  use SimModule, only: count_errors, store_error, store_error_filename
   use TspFmiModule, only: TspFmiType
   use BndExtModule, only: BndExtType
   use ObsModule, only: DefaultObsIdProcessor
@@ -30,11 +32,13 @@ module GwtSrcModule
     procedure :: source_options => src_options
     procedure :: bnd_rp => src_rp
     procedure :: bnd_cf => src_cf
+    procedure :: bnd_ck => src_ck
     procedure :: bnd_fc => src_fc
     procedure :: bnd_da => src_da
     procedure :: define_listlabel
     procedure :: set_nodesontop
     procedure :: bound_value => src_bound_value
+    procedure :: mass_mult
     ! -- methods for observations
     procedure, public :: bnd_obs_supported => src_obs_supported
     procedure, public :: bnd_df_obs => src_df_obs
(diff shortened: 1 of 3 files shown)


src/Model/GroundWaterTransport/gwt-src.f90
```diff
@@ -197,9 +201,41 @@ subroutine src_allocate_arrays(this, nodelist, auxvar)
...
+  !<
+  subroutine src_ck(this)
+    ! -- dummy
+    class(GwtSrcType), intent(inout) :: this
+    ! -- local
+    integer(I4B) :: i
+    integer(I4B) :: node
+    ! -- formats
+    character(len=*), parameter :: fmtenermulterr = &
+      "('SRC BOUNDARY (',i0,') SRC MULTIPLIER (',g10.3,') IS &
```
aprovost-usgs: Assuming the mass source loading can be negative, might someone legitimately want to use a negative multiplier?

src/Model/GroundWaterTransport/gwt-src.f90
```diff
@@ -394,4 +433,24 @@ function src_bound_value(this, col, row) result(bndval)
...
+    ! -- modules
+    use ConstantsModule, only: DZERO
+    ! -- dummy variables
+    class(GwtSrcType), intent(inout) :: this !< BndExtType object
+    integer(I4B), intent(in) :: row
+    ! -- result
+    real(DP) :: ener
+    !
+    if (this%iauxmultcol > 0) then
+      ener = this%smassrate(row) * this%auxvar(this%iauxmultcol, row)
```
aprovost-usgs, line 417: Is function src_bound_value needed anymore?

aprovost-usgs (commented):
Similar comments to PR #2597 (ESL)

src/Model/GroundWaterTransport/gwt-src.f90
emorway-usgs: In the ESL package, the removal of more and more energy from a grid cell could lead to increasingly "negative" temperatures; however, that isn't necessarily a big deal in GWE since the temperature scale is relative to whatever the user wants (i.e., <0 deg temperatures may not mean a frozen condition).  However, in `SRC`, we certainly cannot have a negative concentration...thus I think it would be safer to preclude a switch in the specified sign...though perhaps there are other checks in place to prevent negative concentrations from occurring?

src/Model/GroundWaterTransport/gwt-src.f90
emorway-usgs: Similar to the response over on #2597 (see [this](https://github.com/MODFLOW-ORG/modflow6/pull/2597#discussion_r2539903699)), this is something for @mjreno to weigh in on.

mjreno: responded to similar question in #2597

src/Model/GroundWaterTransport/gwt-src.f90
aprovost-usgs: Don't recall if there are checks to prevent negative concentrations.  But in any case I'm pretty sure that removal of mass (or energy) via SRC (or ESL) is allowed, i.e., you can set the rate to be negative.  If so, then I think a negative multiplier should be allowed, as well.  The user could set a positive rate, then hit it with a negative multiplier to get a negative rate.  That strikes me as no worse than setting a negative rate to begin with.

src/Model/GroundWaterTransport/gwt-src.f90
aprovost-usgs: @emorway-usgs and @mjreno, roger that, thanks

src/Model/GroundWaterTransport/gwt-src.f90
emorway-usgs: All sound reasoning.  Have removed enforcement of a strictly positive multiplier with 809068c

aprovost-usgs (approved):

PR #2612, opened 2025-12-11
fix(prt): add warning for high exit_solve_tolerance

wpbonelli:
Setting PRP option `EXIT_SOLVE_TOLERANCE` too high can produce invalid/meaningless results with the generalized DISV tracking method, as the algorithm then considers every lateral exit solution from every subcell equally valid.

Add a warning for tolerances above 0.01. The variable being solved varies from 0 to 1, so in most cases the tolerance should be much less than 1, though appropriate values will vary by model and desired performance (higher tolerance = faster).

Also update the DFN parameter description. 

___

### DIFF
--- a/src/Model/ParticleTracking/prt-prp.f90
+++ b/src/Model/ParticleTracking/prt-prp.f90
@@ -2,7 +2,7 @@ module PrtPrpModule
   use KindModule, only: DP, I4B, LGP
   use ConstantsModule, only: DZERO, DEM1, DEM5, DONE, LENFTYPE, LINELENGTH, &
                              LENBOUNDNAME, LENPAKLOC, TABLEFT, TABCENTER, &
-                             MNORMAL, DSAME, DEP3, DEP9
+                             MNORMAL, DSAME, DEP3, DEP9, DEM2
   use BndModule, only: BndType
   use BndExtModule, only: BndExtType
   use ObsModule, only: DefaultObsIdProcessor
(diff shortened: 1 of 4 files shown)


src/Model/ParticleTracking/prt-prp.f90
```diff
@@ -683,6 +683,13 @@ subroutine prp_options(this)
       &[character(len=LENVARNAME) :: 'NONE', 'EAGER']
     character(len=LINELENGTH) :: trackfile, trackcsvfile, fname
     type(PrtPrpParamFoundType) :: found
+    character(len=*), parameter :: fmtextolwrn = &
+      "('WARNING: EXIT_SOLVE_TOLERANCE is set to ',g10.3,' &
+      &which is much greater than the default value of 0.00001. &
```
aprovost-usgs: Consider making the default value output dynamic (not hardwired in the format) in case we want to change the value in the future?

src/Model/ParticleTracking/prt-prp.f90
```diff
@@ -683,6 +683,13 @@ subroutine prp_options(this)
...
+      &The tolerance that strikes the best balance between accuracy &
+      &and runtime is problem-dependent. Since the variable being solved &
+      &solved varies from 0 to 1, tolerance values much less than 1 &
```
aprovost-usgs, line 692: "solved" appears twice

aprovost-usgs (commented):
a couple of minor things

src/Model/ParticleTracking/prt-prp.f90
wpbonelli: done. and pulled out a constant for the default.

src/Model/ParticleTracking/prt-prp.f90
wpbonelli: thanks

PR #2624, opened 2025-12-20
fix(prt-prp): fix release local z coord calculation

wpbonelli:
Fix conversion of local z release point coordinates to model z coordinates: take into account whether the cell is convertible or confined, and avoid incorrectly computing model z coordinates to be below the cell bottom. If convertible, define the cell's vertical extent from the bottom to the saturated top, clamping sat top to the bottom if head is below the cell bottom elevation; if confined, use the cell's full vertical extent from the bottom to the geometric top.

In Newton models, the failure to clamp to the bottom could produce pathlines with invalid (negative) times. A particle released by local z coord into a cell with head below the cell bottom would effectively be placed below the cell bottom, which poisoned subsequent tracking calculations.

The user must now provide a binary grid file to PRT FMI if release coordinates are configured with `LOCAL_Z`, to distinguish convertible from confined cells at release time.

Internally, this required giving FMI a pointer to cell type as configured in GWF NPF. If GWF and PRT are in the same simulation, cell type is set from the EXG. If in separate sims, it comes from the GRB file.

This in turn required tweaking the grid file reader, as the preexisting array read functions did not support reading into preallocated, memory-managed arrays. 

---


wpbonelli:
Some plots demonstrating the new behavior

#### Head within convertible cell

head_within_convertible

#### Head within confined cell

head_within_confined

#### Head below convertible cell

head_below_convertible

#### Head below confined cell

head_below_confined

wpbonelli:
CI failure is a chicken and egg situation with the Keating example snapshot

autotest/test_prt_localz_celltype.py
```diff
@@ -0,0 +1,522 @@
...
+Fix #2: Use correct top based on cell type
+    - Confined cells (icelltype=0): Use geometric top for local z conversion
+    - Convertible cells (icelltype!=0): Use head as effective top for local z conversion
+
+The test includes 4 cases, each with a simple single-layer model:
+1. Convertible cell with head within bounds → uses head as top
+2. Confined cell with head within bounds → uses geometric top
+3. Convertible cell with head < bottom (dry cell) → uses clamped head
+4. Confined cell with head < bottom → uses geometric top (ignores head)
+
```
aprovost-usgs, line 20: For completeness, would it be good to also test the cases where the head is above the geometric top, or would that be overkill?  (Both confined and unconfined should use the geometric top as the top in that case)

doc/ReleaseNotes/develop.toml
```diff
@@ -77,3 +77,8 @@ description = "The OPTIONS block in the Mass Source Loading (SRC) Package accept
...
+
+[[items]]
+section = "fixes"
+subsection = "model"
+description = "In Newton models with local z release coordinates configured via LOCAL_\\Z, invalid negative particle tracking times could be produced. This was caused by an incorrect conversion of local release point vertical coordinates to model coordinates, erroneously placing particles below the bottom of the release cell. To correct this, take into account whether the cell is convertible or confined. If convertible, define the cell's vertical extent from the bottom to the saturated top, clamping saturated top to the cell bottom if head is below the cell bottom; if confined, use the cell's full vertical extent from the bottom to the geometric top. The user must now specify a binary grid file entry (GWFGRID) to the PRT FMI package if release coordinates are configured with LOCAL_\\Z, so that convertible and confined cells can be distinguished at particle release time."
```
aprovost-usgs, line 84: "... could be produced when the head in the release-point cell is below the cell bottom."  Is this correct?  Would be good to give readers a sense of when this could occur

aprovost-usgs, line 84: is the binary grid file still required if the GWF and PRT are in the same sim?

src/Model/ParticleTracking/prt-prp.f90
```diff
@@ -465,6 +465,12 @@ subroutine initialize_particle(this, particle, ip, trelease)
     integer(I4B) :: ic, icu, ic_old
     real(DP) :: x, y, z
     real(DP) :: top, bot, hds
+    ! formats
+    character(len=*), parameter :: fmticterr = &
+      "('Error in ',a,': Flow model interface does not contain ICELLTYPE. &
```
aprovost-usgs, line 470: Is this relevant only in the case of PRT in a separate sim?

src/Model/ParticleTracking/prt-prp.f90
```diff
@@ -510,14 +516,29 @@ subroutine initialize_particle(this, particle, ip, trelease)
...
+      ! calculate model z coord from local z coord.
+      ! if cell is confined (icelltype == 0) use the
+      ! actual cell height (geometric top - bottom).
+      ! otherwise use head as cell top, clamping to
+      ! the cell bottom if head is below the bottom
       top = this%fmi%dis%top(ic)
       bot = this%fmi%dis%bot(ic)
       hds = this%fmi%gwfhead(ic)
-      z = bot + this%rptz(ip) * (hds - bot)
+      if (this%fmi%gwfceltyp(icu) /= 0) top = hds
```
aprovost-usgs, line 539: think we want to limit this to min(top, hds), i.e., clamp to top if head is above top

src/Utilities/GridFileReader.f90
```diff
@@ -30,15 +30,23 @@ module GridFileReaderModule
...
+    procedure, public :: read_grid_shape
+    ! array read functions (allocate and return)
     procedure, public :: read_int_1d
     procedure, public :: read_dbl_1d
     procedure, public :: read_charstr
-    procedure, public :: read_grid_shape
-    procedure, private :: read_header
-    procedure, private :: read_header_meta
-    procedure, private :: read_header_body
+    ! array read subroutines (populate preallocated)
```
aprovost-usgs, line 46: is the issue of reading into preallocated, memory-managed arrays related to the idm?  it's just my limited familiarity with that, but i'm wondering why we haven't needed these routines before (or for other data, as well)

aprovost-usgs (commented):
happy to chat if it would be helpful

src/Model/ParticleTracking/prt-prp.f90
wpbonelli: it's always relevant if local z coordinates are used. if separate sims, the cell type array comes from the binary grid file via FMI. if same sim, it comes from the EXG. but PRT needs it in either case if local z. as implemented, it will tolerate not knowing cell type if local z coords are not used.

doc/ReleaseNotes/develop.toml
wpbonelli: no, it's only required if GWF and PRT are separate. will update the language here

autotest/test_prt_localz_celltype.py
wpbonelli: agreed. will add that case.

src/Model/ParticleTracking/prt-prp.f90
wpbonelli: thanks, yes- this will be exercised by the new test case you suggest above

src/Utilities/GridFileReader.f90
wpbonelli: this just reflects the fact that I wrote the grid file reader in an unconventional way in #2153, with array read functions returning arrays. that was sufficient for loading idomain into local arrays and checking it. I didn't anticipate that the grid file reader would need to load data into IDM-managed arrays.

doc/ReleaseNotes/develop.toml
wpbonelli: yes- thanks, will edit that in.

src/Utilities/GridFileReader.f90
aprovost-usgs: got it, thanks for explaining

PR #2626, opened 2025-12-23
chore(mst/ist) Refactor isotherm

Manangka:
This PR refactors the Isotherms in the MST and IST packages
It doesn't contain any functional changes, 

The 3 isotherms are moved to their own files and they all implement the IsothermType interface.
A Isotherm factory and Isotherm enum module is added as well

By making these changes a lot of if-else and select case blocks of logic can be simplified.
The main reason for this PR is to prepare the code base for the next PR in which a new formulation for the sorption term in the mst package will be added

I cleaned up most of the code related to the isotherm. However there are 2 areas that need more attention but which i consider out of scope:
- the sorpteddecay formulation
- the ist formulation


src/Model/TransportModel/Isotherm/IsothermEnum.f90
```diff
@@ -0,0 +1,11 @@
+module IsothermEnumModule
```
mjr-deltares, line 1: We have been using iso c enums in the Constants.f90 file to define enumerators, though not consistently. It would avoid using a file only for this single enumeration. In any case, we should probably decide on best practice here, so we can move forward and streamline the code base.

mjr-deltares (approved):
Nice job. I like it. But, we should inform the team on the details and purpose of this pattern. I think that can be done after merging though.

Issue #2610, opened 2025-12-11
Profiling is turned on by default and can't be turned off


p-ortega:
Hi dev team,

I apologise for the basic description of this and if is the intended behaviour, but I've encountered this when running a mf6 sim with several transport models. 

**Describe the bug**
The profiling option is always on by default and the following does not turn it off:

```
BEGIN options
  PROFILE_OPTION NONE
END options
``` 

**To reproduce**
I have attached a multi-species model that hits the MAX_NR_TIMED_SECTIONS when profiling is turned on.
[issue.tar.gz](https://github.com/user-attachments/files/24093223/issue.tar.gz)

**Expected behavior**
Profiling off by default and be able to turn it off using the block option

**Environment**
 - Same issue across Linux, MacOS, and Windows.
 - MODFLOW 6 version 6.6.2, 6.6.3 and 6.7.0-dev from nightly build.

**Context**
I got this when I was testing a mf6api application across platforms.
[example](https://github.com/p-ortega/mf6rtm/actions/runs/20119547353/job/57736546766?pr=45)

mjr-deltares:
hi @p-ortega , thank you for reporting this! There is some improvement required for the profiling tool. I think that your example shows that using MAX_NR_TIMED_SECTIONS  was a bad idea because there is no sensible maximum to be defined. Increasing the number of species will eventually always lead to the problem your are experiencing. I will look into making the container with profiling data dynamically sized which should solve that. Setting profiling to NONE currently only cancels the printing of the numbers but not the actual measurements, and therefore doesn't help you. We will change that as well.

mjr-deltares:
Hi @p-ortega , a fix has been merged into develop. Let us know if you are still experiencing problems.

p-ortega:
Thank you @mjr-deltares ! Appreciate the quick turn around.

PR #2630, opened 2026-01-06
fix(profiling): profiling crashes for large number of solution and can't be turned off

mjr-deltares:
This fixes the issue reported in #2610. Solution is to delay initialization of profiler so nr. of solutions is available to allocate space.

- [x] Closed issue #xxxx
- [x] Referenced issue or pull request #xxxx


mjreno (approved):
Looks good to me @mjr-deltares , just a comment to consider (outside the current scope of the fix) for later:
 - when running in `summary` mode only the summary table is printed but this doesn't reflect all category timings that were captured (e.g. alloc/dealloc, prepare, netcdf export).  There might be a reason for this but it would be nice if all summary categories could be written, perhaps combining certain categories in the summary but ideally with all time accounted for.

mjr-deltares:
thanks @mjreno for reviewing. I appreciate your comment about printing the summary in a more complete manner. I will create an enhancement issue to cover this and maybe we can look into it after the upcoming release.

PR #2, opened 2018-02-07
Initial commit of VKD process modifications

WilliamWitterick:
First pass of VKD process. No VKD specific tests at this stage although should at least pass existing tests. 

I had to make some changes to the develop makefile in order to compile (on top of those required for new code).

Issue #1, opened 2018-01-10
Clarify MAW conductance descriptions in dfn


jdhughes-usgs:
From: Winston, Richard <rbwinst@usgs.gov>
Date: Tue, Jan 9, 2018 at 7:32 PM
Subject: more issues with MODFLOW-6 input descriptions
To: Christian D Langevin <langevin@usgs.gov>

In the description of condeqn in MAW, the description of "MEAN" has "using using" instead of just "using".

In the description of condeqn in MAW, the description of "MEAN" says that the "aquifer and screen hydraulic conductivity" are used in calculating the conductance. After looking at T&M-55 and the source code, I think only the screen hydraulic conductivity is used.

In the description of condeqn in MAW, the description of "SKIN" says that the "screen hydraulic conductivity" is used. After looking at T&M-55 and the source code, I think both the aquifer and screen hydraulic conductivity are used in calculating the conductance.

jdhughes-usgs:
closed by commit [0d6b99ae551f839a1735d5f5e1d46e1ae624ab52](https://github.com/MODFLOW-USGS/modflow6/commit/0d6b99ae551f839a1735d5f5e1d46e1ae624ab52)

PR #2631, opened 2026-01-06
fix(prt-prp): revise local z coord calculation

wpbonelli:
As suggested in https://github.com/MODFLOW-ORG/modflow6/pull/2624#discussion_r2662461128, clamp effective cell top not only to the bottom (if head is below the cell) but to the top (if head is above), and revise release note.


autotest/test_prt_localz_celltype.py
```diff
@@ -4,19 +4,21 @@
...
     When using head as the effective cell top (in convertible cells), ensure the
-    head doesn't fall below the cell bottom. The calculation uses max(head, bottom).
+    head doesn't fall above the cell top or below the cell bottom.
 
 Fix #2: Use correct top based on cell type
-    - Confined cells (icelltype=0): Use geometric top for local z conversion
+    - Confined cells (icelltype==0): Use geometric top for local z conversion
     - Convertible cells (icelltype!=0): Use head as effective top for local z conversion
 
 The test includes 4 cases, each with a simple single-layer model:
```
aprovost-usgs: "includes 6 cases"

autotest/test_prt_localz_celltype.py
```diff
@@ -37,7 +39,7 @@
 
 simname = "prtz"
 
-# Four test cases, each with FMI and exchange variants
+# Six test cases, each with FMI and exchange variants
 cases = [
     f"{simname}cvt",  # convertible, normal head
```
aprovost-usgs: consider saying "bottom < head < top" in place of "normal head"?  would be more specific and consistent with other inline comments

autotest/test_prt_localz_celltype.py
```diff
@@ -4,19 +4,21 @@
 This test verifies two fixes in the particle local-to-model z coordinate conversion
 when using the LOCALZ option in the PRP package:
 
-Fix #1: Clamp head to cell bottom
+Fix #1: Clamp head to cell vertical extent
```
aprovost-usgs: i'd say it's the effective cell top (not the head) that gets clamped.  from PRT's point of view, the geometric cell top, the cell bottom, and the head in the cell are given, and it sets the effective top to one of those, depending on the context

wording like this would be consistent with that view:

Fix #1: Constrain effective top to cell vertical extent
    - For convertible cells, whose saturated thickness depends on head, set the effective cell top used for local z conversion no higher than the geometric cell top and no lower than the cell bottom.
Fix #2: Use correct top based on cell type
    - Confined cells (icelltype==0): Use geometric top for local z conversion
    - Convertible cells (icelltype!=0): Use head (subject to constraint in fix #1) as effective top for local z conversion

autotest/test_prt_localz_celltype.py
```diff
@@ -131,6 +137,42 @@
...
+        "botm": [5.0],
+        "strt": [[12.0]],  # head above top
+        "icelltype": [1],
+        "releasepts": [
+            [0, 0, 0, 0, 2.0, 5.0, 0.0],
+            [1, 0, 0, 0, 5.0, 5.0, 0.5],
+            [2, 0, 0, 0, 8.0, 5.0, 1.0],
+        ],
+        "expected_z": {
+            # Head clamped to top: top = min(10.0, 12.0) = 10.0
```
aprovost-usgs: suggest "effective top clamped to top"

autotest/test_prt_localz_celltype.py
```diff
@@ -131,6 +137,42 @@
...
+            0: 5.0,  # bot + 0.0 * (min(top,head) - bot) = 5.0 + 0.0 * 5.0 = 5.0
+            1: 7.5,  # bot + 0.5 * (min(top,head) - bot) = 5.0 + 0.5 * 5.0 = 7.5
+            2: 10.0,  # bot + 1.0 * (min(top,head) - bot) = 5.0 + 1.0 * 5.0 = 10.0
+        },
+        "description": "Convertible cell, head > top → clamped to top",
+    },
+    "cnfhi": {  # Case 6: Confined, head > top
+        "top": 10.0,
+        "botm": [5.0],
+        "strt": [[12.0]],  # head above top (artesian conditions)
```
aprovost-usgs: head > top might represent artesian conditions, but it would require head > land surface elevation.  might be best to just leave it at "head above top", as i think any further description is unlikely to add to the reader's understanding of the main issue (converting local z for particle tracking) anyway

similarly, consider leaving out "negative pressure head" further up.  having the head (which is "measured" at the cell center) below the bottom of the confined cell would be sufficient to result in negative pressure head.  but it might be better to describe it simply as "head below bottom" and not get into hydrology for the present purpose

doc/ReleaseNotes/develop.toml
```diff
@@ -81,7 +81,7 @@ description = "Setting the PRT model's Particle Release Point (PRP) package opti
...
...ewton models with local z release coordinates configured via LOCAL\\_Z, invalid negative particle tracking times could be produced when the head in the release-point cell is above the cell top or below the cell bottom. This was caused by an incorrect conversion of local release point vertical coordinates to model coordinates, erroneously placing particles above or below the vertical extent of the release cell. To correct this, take into account whether the cell is convertible or confined. If convertible, define the cell's vertical extent from the bottom to the saturated top, clamping saturated top to the cell bottom if head is below the bottom and to the cell top if head is above the top; if confined, use the cell's full vertical extent from the bottom to the geometric top. So that convertible and confined cells can be distinguished at particle release time, the user must now specify a binary grid file entry (GWFGRID) to the PRT FMI package if release coordinates are configured with LOCAL\\_Z and the PRT model is running in a separate simulation from the GWF model. A binary grid file entry is not required in the PRT FMI package if the GWF and PRT models run in the same simulation."
```
aprovost-usgs: "saturated top" could be taken to mean the top when the cell is fully saturated, i.e., the geometric top.  suggest the following wording in place of the fourth sentence:

The release point is interpolated vertically between the effective top of the cell and the cell bottom. If the cell is convertible, the effective top is determined by the head in the cell and is constrained to be no higher than the geometric top of the cell and no lower than the cell bottom.  If the cell is confined, the effective top is the geometric top.

src/Model/ParticleTracking/prt-prp.f90
```diff
@@ -532,12 +532,13 @@ subroutine initialize_particle(this, particle, ip, trelease)
...
-      ! the cell bottom if head is below the bottom
+      ! the cell bottom if head is below the bottom.
       top = this%fmi%dis%top(ic)
       bot = this%fmi%dis%bot(ic)
       hds = this%fmi%gwfhead(ic)
-      if (this%fmi%gwfceltyp(icu) /= 0) top = hds
-      if (top < bot) top = bot
+      if (this%fmi%gwfceltyp(icu) /= 0) &
+        top = min(top, hds)
+      top = max(top, bot)
```
aprovost-usgs: "top = max(top, bot)" isn't necessary if the cell is confined, so this max operation can be put inside the "if" and executed only if the cell is convertible.  similarly, "hds = this%fmi%gwfhead(ic)" is needed only if convertible, so it can also be put inside the "if".  the gain in efficiency would probably be insignificant, but i think it would look cleaner

src/Model/ParticleTracking/prt-prp.f90
```diff
@@ -532,12 +532,13 @@ subroutine initialize_particle(this, particle, ip, trelease)
       ! if cell is confined (icelltype == 0) use the
       ! actual cell height (geometric top - bottom).
       ! otherwise use head as cell top, clamping to
-      ! the cell bottom if head is below the bottom
+      ! the cell bottom if head is below the bottom.
```
aprovost-usgs, line 535: "and to the geometric cell top if head is above the top"

aprovost-usgs (approved):
just a few wording tweaks and an ultra-minor refactor for your consideration

src/Model/ParticleTracking/prt-prp.f90
aprovost-usgs: i think this is correct (for a convertible cell) and should be added to the end of the above description

autotest/test_prt_localz_celltype.py
```diff
@@ -4,15 +4,17 @@
 This test verifies two fixes in the particle local-to-model z coordinate conversion
 when using the LOCALZ option in the PRP package:
 
-Fix #1: Clamp head to cell vertical extent
-    When using head as the effective cell top (in convertible cells), ensure the
-    head doesn't fall above the cell top or below the cell bottom.
+Fix #1: Constrain effective top to cell vertical extent
+    - For convertible cells whose saturated thickness depends on head set the effective
```
aprovost-usgs, line 8: i'd put commas around "whose saturated thickness depends on head" so readers interpret it as a general statement about convertible cells

aprovost-usgs (commented):
sorry, just a couple of fiddly things

src/Model/ParticleTracking/prt-prp.f90
wpbonelli: sorry for missing this

PR #2636, opened 2026-01-08
fix(prt): minor revisions

wpbonelli:
Address comments from #2631

### DIFF
--- a/src/Model/ParticleTracking/prt-prp.f90
+++ b/src/Model/ParticleTracking/prt-prp.f90
@@ -532,7 +532,8 @@ contains
       ! if cell is confined (icelltype == 0) use the
       ! actual cell height (geometric top - bottom).
       ! otherwise use head as cell top, clamping to
-      ! the cell bottom if head is below the bottom.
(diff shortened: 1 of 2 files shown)


src/Model/ParticleTracking/prt-prp.f90
```diff
@@ -532,7 +532,8 @@ subroutine initialize_particle(this, particle, ip, trelease)
       ! if cell is confined (icelltype == 0) use the
       ! actual cell height (geometric top - bottom).
       ! otherwise use head as cell top, clamping to
-      ! the cell bottom if head is below the bottom.
+      ! the cell bottom if head is below the bottom and
+      ! and to geometric cell top if head is above top
```
aprovost-usgs, line 536: there's an extra "and"

aprovost-usgs (approved):
thanks for making these changes.  please feel free to merge without further review after removing the extra "and"

src/Model/ParticleTracking/prt-prp.f90
wpbonelli: thanks!

PR #2644, opened 2026-01-10
fix(prt): fix particle event kper/kstp assignment

wpbonelli:
PRT previously reported events occurring exactly on a boundary between time steps as occurring during the preceding step. This properly 'credited' the event to the flow conditions responsible for the particle's trajectory up to that time. However, this reporting logic was inappropriately applied to events such as particle releases, as well, which caused the stress period and time step to be reported incorrectly for some events. Instead of altering stress period and time step post hoc, PRT now reports each event as occurring in the stress period and time step in which the event is triggered. This means simultaneous releases can be credited to different time steps. For example, a particle release scheduled using an explicit time specification that falls on the boundary between two time steps occurs at the first opportunity, i.e., at the end of the earlier time step. On the other hand, if a period block is used to schedule a particle release at the beginning of the later time step, the release occurs in that time step, not the previous one.

The changes mainly consist of tightening up the semantics and implementation of `TimeSelectModule`, which will now use slices `(t0, t1]`, i.e exclusive lower bound and inclusive upper bound, and the `advance` routine was incorrect in several ways. Conditionals in the various tracking methods are also updated to implement the new approach, which can be summarized as "report events as soon as possible, without duplication".

`test_prt_release_timing.py` is updated here with 3 new cases. `test_prt_watertable.py` is also updated to reflect new, corrected results. PRT/MP7 examples 1-3 also needed snapshot updates.


doc/ReleaseNotes/develop.toml
```diff
@@ -117,3 +117,8 @@ description = "Numerical error in the iterative subcell exit solution can leave
...
+
+[[items]]
+section = "fixes"
+subsection = "model"
+description = "PRT previously attempted to report events occurring exactly on a boundary between time steps as occurring during the preceding step, in order to properly 'credit' the flow conditions responsible for the particle's trajectory up to that time. This logic was implemented incorrectly and could cause stress period and time step to be assigned incorrectly for some events. Instead of altering stress period and time step post hoc, PRT will now report each event as occurring in the stress period and time step in which the reporting condition responsible for triggering it is recognized. By consequence, particle release events occurring exactly at a time step boundary may now be assigned to either side of the boundary, depending how the release is configured: if configured via explicit release time, it will fall on the preceding side, while if configured via stress period settings, it will be reported on the following side."
```
aprovost-usgs: Consider:
PRT previously reported events occurring exactly on a boundary between time steps as occurring during the preceding step. This properly 'credited' the event to the flow conditions responsible for the particle's trajectory up to that time. However, this reporting logic was inappropriately applied to events such as particle releases, as well, which caused the stress period and time step to be reported incorrectly for some events. Instead of altering stress period and time step post hoc, PRT now reports each event as occurring in the stress period and time step in which the event was triggered. For example, a particle release scheduled using an explicit time specification that falls on the boundary between two time steps is triggered at the first opportunity, i.e., it occurs, and is reported to have occurred, at the end of the earlier time step.  On the other hand, if a period block is used to schedule a particle release at the beginning of particular time step, the release occurs, and is reported to have occurred, on that time step, not the previous one.

aprovost-usgs (approved):

Issue #2656, opened 2026-01-19
Settings unknown value for cell averaging in GWF-GWF exchange crashes MODFLOW


mjr-deltares:
When setting the GWF-GWF Exchange option CELL_AVERAGING to some unknown value, the simulation crashes:

 ```
Writing simulation list file: mfsim.lst
 Using Simulation name file: mfsim.nam

At line 1267 of file ../src/Exchange/exg-gwfgwf.f90
Fortran runtime error: Index '0' of dimension 1 of array 'cellavg_method' below lower bound of 1

Error termination. Backtrace:
#0  0x7f44d7f6be59 in ???
#1  0x7f44d7f6ca71 in ???
#2  0x7f44d7f6d082 in ???
...
```

mjr-deltares:
fixed and closed

PR #2660, opened 2026-01-20
fix(input): set errors for non-valid option strings

mjreno:
Set consistent errors for non-valid string input options:
- #2656
- verify that when an input string with a valid set is read that it matches an expected string and set an error if it does not
- some packages are already validating strings and setting errors: `prt-prp`, `gwt-mst`, `gwt-ist`, `tsp-adv`
- the `DIS` packages are outliers as the documentation states that `FEET`, `METERS` and `CENTIMETERS` are supported as `LENGTH_UNITS` but some tests set the input string as `UNDEFINED`.  Currently any unrecognized `LENGTH_UNITS` input values are assigned as `UNDETERMINED`/`UNDEFINED`/`UNKNOWN`.

- [X] Referenced issue or pull request #xxxx
- [X] Added new test or modified an existing test
- [X] Ran `ruff` on new and modified python scripts in .doc, autotests, doc, distribution, pymake, and utils subdirectories.

### DIFF
--- a/src/Model/GroundWaterFlow/gwf-npf.f90
+++ b/src/Model/GroundWaterFlow/gwf-npf.f90
@@ -1418,6 +1418,15 @@ contains
       call tvk_cr(this%tvk, this%name_model, this%intvk, this%iout)
     end if
     !
+    ! -- verify ALTERNATIVE_CELL_AVERAGING input value is supported
+    if (found%cellavg) then
+      if (this%icellavg == 0) then
+        errmsg = 'Unrecognized input value for ALTERNATIVE_CELL_AVERAGING option.'
+        call store_error(errmsg)
+        call store_error_filename(this%input_fname)
+      end if
+    end if
+    !
     ! -- log options
     if (this%iout > 0) then
       call this%log_options(found)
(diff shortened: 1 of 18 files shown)


src/Model/GroundWaterFlow/gwf-npf.f90
```diff
@@ -1418,6 +1418,15 @@ subroutine source_options(this)
       call tvk_cr(this%tvk, this%name_model, this%intvk, this%iout)
     end if
     !
+    ! -- verify ALTERNATIVE_CELL_AVERAGING input value is supported
+    if (found%cellavg) then
+      if (this%icellavg == 0) then
```
mjr-deltares, line 1423: I had to think a little bit about this one, but agree that it's correct. If found but the value equals the default (harmonic), apparently the string was not supported. This then all works because here because we specify ALTERNATIVE_cell_averaging where the default=harmonic is not among the valid strings...

mjr-deltares (approved):
I like how you have solved this. My only remark was about the confusing use of ALTERNATIVE_CELL_AVERAGING in NPF where we set CELL_AVERAGING on the exchange. But, even though we might straighten that out at some point, I don't think this should be part of this PR.

PR #2664, opened 2026-01-21
doc(prt): rephrase PRT budget item in known issues section

wpbonelli:
update the description of the PRT budget, which has been balanced in the meantime and is minimal but functional

### DIFF
--- a/doc/ReleaseNotes/ReleaseNotes.tex
+++ b/doc/ReleaseNotes/ReleaseNotes.tex
@@ -234,7 +234,8 @@ If a GWF-GWF Exchange is active, then neither of the connected GWF Models can ha
 \item 
 If a GWT-GWT or GWE-GWE Exchange is active, then constant concentration (CNC) or constant temperature (CTP) conditions should not be assigned on the interface between the two connected models because the budget reported in the listing file will contain errors.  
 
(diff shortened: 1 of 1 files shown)


doc/ReleaseNotes/ReleaseNotes.tex
```diff
@@ -234,8 +234,7 @@ \section{Known Issues and Incompatibilities}
 \item 
 If a GWT-GWT or GWE-GWE Exchange is active, then constant concentration (CNC) or constant temperature (CTP) conditions should not be assigned on the interface between the two connected models because the budget reported in the listing file will contain errors.  
 
-\item
-The PRT Model reports an incomplete and unbalanced particle mass budget to the listing file and the binary cell budget file.
+The PRT Model reports a high-level particle mass budget to the listing file and the binary cell budget file, with terms for release, storage, and termination only. It does not yet account for flows to distinct boundary conditions.
```
aprovost-usgs: how about something like:
The PRT Model reports balanced particle mass budgets to the listing file and the binary cell budget file, with budget entries for release, storage, and termination. However, the termination term is aggregated and does not distinguish particle outflows to different types of boundary conditions.

emphasizes that the budget is balanced, i.e., nothing's "missing."  also, saying "not yet" felt like a sort of promise that it will someday offer a more granular budget.  saying "does not currently" might be a distinction without a difference, but it seems to at least allow for possibility that it might not ever, i.e., it doesn't commit us to it

aprovost-usgs (approved):

PR #2675, opened 2026-01-28
fix(GeomUtilModule): fix inverse compose

wpbonelli:
As PRT tracks a particle it maintains a transformation between model coordinates and coordinates local to the current tracking domain. Since the domain is hierarchical, multiple transformations are applied as tracking is delegated from higher to lower-level tracking methods.

Coordinate transformations are supported by two geometry routines, `transform` which transforms points, and `compose` accumulates transformations. Both routines have a parameter `invert`. When a submethod hands control of a particle back to a delegating method after subdomain tracking is done, its corresponding transform is inverted to restore the delegating method's coordinate system.

Up to now, `compose(..., invert=.true.)` tried to apply a rotation regardless whether the transform added with `compose` had a rotation or not. So a pure translation on top of a transform with rotation would be incorrectly inverted but a pure translation on another pure translation would be inverted correctly.

This could produce invalid particle coordinates on rotated* disv grid cells, with the error magnitude increasing with the magnitude of coordinates, without substantially disturbing other parts of the pathline.

*I think we failed to notice this til now because grids are typically defined axis-aligned then rotated via metadata e.g. `angrot`. The tracking method only sees non-axis-aligned rectilinear cells on a DISV if the grid vertices/cells are not themselves axis-aligned.

Without the fix, on a disv grid where the grid is not aligned to x/y axes

prt_bug

With the fix

prt_bug_fix

___


aprovost-usgs (approved):
reviewed and discussed

PR #2698, opened 2026-02-12
fix(api): ensure IDM advance is called when advancing model through the API

jlarsen-usgs:
Fixes issue in MODFLOW-ORG/modflowapi#72

PR adds a conditional to `idm_ad()` in `Mf6DoTimestep()` and adds `idm_ad()` to `Mf6PrepareTimestep()`. This resolves an issue with the API where timeseries inputs are not updated when using the `xmi_prepare_solve`, `xmi_solve`, `xmi_finalize_solve` convention.

- [x] Referenced issue or pull request #xxxx


christianlangevin (commented):
Looks like the call to `idm_ad()` was not being called for certain API cases, which explains why the API gave different results for models with time series.  It would be good to get @mjreno to give his opinion.  My understanding is that `idm_ad()` will advance all dynamic packages in the entire simulation, so this refactoring will introduce extra work.  

@mjreno what are your thoughts on this?

mjreno:
> Looks like the call to `idm_ad()` was not being called for certain API cases, which explains why the API gave different results for models with time series. It would be good to get @mjreno to give his opinion. My understanding is that `idm_ad()` will advance all dynamic packages in the entire simulation, so this refactoring will introduce extra work.
> 
> @mjreno what are your thoughts on this?

This looks good to me and makes sense that the disconnect occurs when API based simulations circumvent `mf6core`.  You are right @christianlangevin ,  `idm_ad()` will advance all (IDM) model dynamic packages with configured timeseries.

christianlangevin:
> > Looks like the call to `idm_ad()` was not being called for certain API cases, which explains why the API gave different results for models with time series. It would be good to get @mjreno to give his opinion. My understanding is that `idm_ad()` will advance all dynamic packages in the entire simulation, so this refactoring will introduce extra work.
> > @mjreno what are your thoughts on this?
> 
> This looks good to me and makes sense that the disconnect occurs when API based simulations circumvent `mf6core`. You are right @christianlangevin , `idm_ad()` will advance all (IDM) model dynamic packages with configured timeseries.

Hey @mjreno, since `idm_ad()` will be called for each numerical solution, will this cause problems or introduce inefficiency that we need to be concerned about?  While this fixes the immediate problem with the API, I worry about the broader consequences.

mjreno:
@christianlangevin @jlarsen-usgs  that's a good point.  Calling `idm_ad()` multiple times per time step will result in repeated calls to timeseries manager advance.  If the timeseries manager recalculates everything at each call this would introduce inefficiencies.  I can look at multi-solution simulation behavior after this is merged or happy to talk about it more.

jlarsen-usgs:
@mjreno @christianlangevin I think another option could be adding a `idm_ad()` call to the xmi prepare solve call. This would reduce/avoid multiple recalculations and likely accomplish the same/similar behavior.

christianlangevin (approved):

PR #2700, opened 2026-02-13
fix(central/utvd) Mass balance issues

Manangka:
A user reported mass‑balance inconsistencies when using the new UTVD scheme.
After investigation, the issue was traced to how the cl1 and cl2 distances were computed. These values depend on cell/node numbering, but the previous implementation implicitly assumed that the ordering was consistent.

Consider two neighboring cells, n1 and n2, where n1 has a smaller cell index than n2.
When distances are retrieved with respect to n1, the behavior is correct:
- cl1: distance from n1 to the shared face
- cl2: distance from n2 to the shared face

However, retrieving distances with respect to n2 produces inconsistent results:
- cl1 is still computed as the distance from n1 to the shared face
- The code incorrectly assumed it represented the distance from n2 to the shared face

This mismatch leads to incorrect flux evaluations and ultimately mass‑balance errors.

```
 ------------------------------
 |              |              |
 |     n1       |      n2      |
 |              |              |
 ------------------------------
```

In this PR
- A fix is added that ensures cl1 and cl2 are interpreted consistently, regardless of cell indexing.
- A unit test is added that verifies that the mass balance for the UTVD scheme is now correct.
- The unit test also revealed the same problem in the Central scheme, so a corresponding fix is included there as well.

- [ ] Closed issue #xxxx


christianlangevin (commented):
Nice find @Manangka.  It looks like that issue was also in the central scheme, which goes way back.  We were bitten by this in the past (in NPF and BUY).  Great to see this fix.

PR #2705, opened 2026-02-20
refactor(prt): switch to ExplicitModel, improve error checking

wpbonelli:
Make PRT an `ExplicitModel` instead of `NumericalModel` and refactor the model types, moving shared things into `BaseModel`
* `create_lstfile` routine
* `nja`
* `flowja`
* `filename`
* `bndlist`
* `dis`

@christianlangevin and I discussed and were not sure if everything belongs in the base type, e.g. do all models share the concept of a single set of face flows, or might that change someday, but this is all easily changed if/when needed.

Making PRT an explicit model is more correct and unlocks error-checking the solution/model match i.e. IMS is used with GWF/GWT/GWE and EMS with PRT, which is a common stumbling block. Add a test for this.
___

### DIFF
--- a/autotest/test_sim_config_errors.py
+++ b/autotest/test_sim_config_errors.py
@@ -1,0 +1,151 @@
+"""
+Test to verify simulation configuration error checking.
+
+Tests the following error conditions:
+1. IMS6 solver paired with PRT model (should require EMS6)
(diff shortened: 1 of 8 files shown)


aprovost-usgs (approved):
makes sense, good call

PR #2723, opened 2026-03-11
fix(GwfExchangeMvr): ensure model name listed in exchng MVR is part of the exchng

emorway-usgs:
When a MVR connection was included as part of an exchange between two models, it was possible to list a model name in the PACKAGES block of the MVR input file that's not one of the two models connected by the exchange.  A check was added that results in an error message when a model name listed in the exchange MVR input file is not one of the two models connected by the exchange.

- [x] Modified an existing test (test_gwt_mvt04.py)

### DIFF
--- a/autotest/test_gwt_mvt04.py
+++ b/autotest/test_gwt_mvt04.py
@@ -2,9 +2,17 @@
 Test problem for GWT with MVT active.  Another autotest will check this autotest
 in parallel.
 
-Test that the MVT successfully transfers solute from two models is successfully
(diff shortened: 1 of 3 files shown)


emorway-usgs:
@mjr-deltares first suspected the bug this PR addresses.  Using one of the new autotests developed for testing the parallel MVR, namely `test_gwt_mvt04.py` (#2718), I confirmed that a user could inadvertently include a model name in an exchange's MVR input.  MF6 will now error out when this situation occurs.

mjr-deltares (approved):
Looks great!

PR #2748, opened 2026-03-31
feat(idm): integrate MAW package

mjreno:
Integrate `GWF-MAW` with `IDM`.

- [X] Added new test or modified an existing test
- [X] Ran `ruff` on new and modified python scripts in .doc, autotests, doc, distribution, pymake, and utils subdirectories.
- [X] Added doxygen comments to new and modified procedures
- [X] Updated meson files, makefiles, and Visual Studio project files for new source files
- [X] Updated [definition files](/MODFLOW-ORG/modflow6/tree/develop/doc/mf6io/mf6ivar)


jdhughes-dev:
@mjreno if possible I would wait before merging this PR until #2858 is merged. Hopefully @christianlangevin can review #2858 before you are ready.

jdhughes-dev:
`read_from_parser_keystring` assigns a sentinel only for real and character members, so an integer member of a keystring setting keeps whatever the column already held on rows that select a different setting — a value from a previous stress period, or uninitialized pre-allocated memory (`src/Utilities/Idm/mf6blockfile/StructArray.f90:1133-1138`).

```fortran
select case (this%struct_vectors(icol)%memtype)
case (MTYPE_DBL) ! DOUBLE: use DNODATA sentinel
  this%struct_vectors(icol)%dbl1d(irow) = DNODATA
case (MTYPE_STR) ! STRING or KEYWORD: use empty string sentinel
  this%struct_vectors(icol)%charstr1d(irow) = ''
end select
```

Nothing hits this today because no converted definition file has an integer member in a keystring; the MAW period members are all keyword records, strings, and reals. I ran into it prototyping a new MAW period setting on top of this branch that takes a connection number and a status, which is the first setting with an integer member. Adding a case for integers is enough:

```fortran
case (MTYPE_INT) ! INTEGER: use zero sentinel
  this%struct_vectors(icol)%int1d(irow) = 0
```

Reading the member only when the dispatch keyword selects it avoids the stale value, so this is not blocking, but a setting that reads its integer unconditionally would pick up the wrong connection or index. Worth folding into this pull request while the file is open.

Everything else worked as expected: the compound dispatch reads an integer followed by a string without any change, and `numeric_index` round-trips correctly through FloPy.

mjreno:
Thanks @jdhughes-dev will definitely take a look at this as part of the current PR.

jdhughes-dev:
The SETTING column stores the parameter tagname, which is silently truncated for any tag longer than 16 characters, so a package that dispatches on such a tag never matches and the setting is ignored without an error.

`StructArray.f90` writes the tagname:

```fortran
! write dispatch keyword as tagname to SETTING column when present
if (setting_icol > 0) then
  this%struct_vectors(setting_icol)%charstr1d(irow) = &
    trim(this%struct_vectors(found_col)%idt%tagname)
end if
```

`tagname` is `character(len=LINELENGTH)`, but the column is allocated at `LENVARNAME` in `Mf6FileKeystring.f90`:

```fortran
call this%structarray%mem_create_vector(sa_icol, this%ctx%setting_idt, &
                                        charlen=LENVARNAME)
```

MAW `CONNECTION_STATUS` is 17 characters, so the column holds `CONNECTION_STATU`. Four MAW PERIOD tags are already over the limit: `CONNECTION_STATUSRECORD` (23), `CONNECTION_STATUS` (17), `FLOWING_WELLRECORD` (18) and `RATE_SCALINGRECORD` (18).

Storing `mf6varname` instead would fit by construction, since it is declared `character(len=LENVARNAME)`:

```fortran
! write dispatch keyword as mf6varname to SETTING column when present
if (setting_icol > 0) then
  this%struct_vectors(setting_icol)%charstr1d(irow) = &
    trim(this%struct_vectors(found_col)%idt%mf6varname)
end if
```

That is not a drop-in change, because three of the comparisons in `maw_rp` dispatch on names that differ from the internal name and would have to change with it:

| tagname | mf6varname | |
| --- | --- | --- |
| `STATUS` | `STATUS` | unchanged |
| `HEAD_LIMIT` | `HEAD_LIMIT` | unchanged |
| `FLOWING_WELL` | `FLOWING_WELL` | unchanged |
| `SHUT_OFF` | `SHUT_OFF` | unchanged |
| `RATE_SCALING` | `RATE_SCALING` | unchanged |
| `RATE` | `RATE_IN` | comparison changes |
| `WELL_HEAD` | `WELL_HEAD_IN` | comparison changes |
| `AUXILIARY` | `PERIOD_AUXILIARY` | comparison changes |

The alternative, if the tagname is the semantics you want, is to widen the column instead and leave the comparisons alone:

```fortran
call this%structarray%mem_create_vector(sa_icol, this%ctx%setting_idt, &
                                        charlen=LINELENGTH)
```

Either resolves the truncation. I would prefer the `mf6varname` form, because the internal name is unique, is bounded by the column width by definition, and is the name the generated definitions and the memory manager already use, but the choice is yours.

Found while rebasing #2914 onto this branch: its `CONNECTION_STATUS` dispatch reads `CONNECTION_STATU` and the connection status is quietly not applied.

mjreno:
Thanks @jdhughes-dev for the report.  This is definitely an issue- if possible can you delay on the rebase until after I push again?  I'll update this branch later today and fix but there are other changes coming as well (mostly in the Idm code).

PR #2753, opened 2026-04-03
fix(prt): support exchange coupling to ATS flow models

wpbonelli:
Support exchange-coupling PRT models to GWF models that use adaptive time stepping (ATS). This would previously produce spurious/incorrect results, as output would be emitted during discarded/retried solve attempts. Spurious output would also be produced on discarded Picard iterations if `MXITER > 1` and PRT is in the same solution group as a flow model. Flopy puts models in the same solution group, but we didn't hit this before now because `MIXITER = 1` by default.

The root cause here is that PRT advance/solve routines are stateful where the other models' equivalents are idempotent. Make the PRT routines idempotent by "staging" state during each solve and only "committing" it on successful solves.

This required fixes in `mem_reallocate()` routines to correctly handle array shrinking (when particles are released in a time step that then fails, they need to be "unreleased").
___


src/Solution/ParticleTracker/Particle/ParticleTracks.f90
```diff
@@ -223,86 +311,55 @@ logical function should_save(this, particle, file) result(save)
             (file%iprp == -1 .or. file%iprp == particle%iprp))
   end function should_save
 
-  !> @brief Save an event to a binary or CSV file.
-  subroutine save_event(iun, particle, event, csv)
+  !> @brief Save an event record to a binary or CSV file.
+  !! Mirrors save_event but expects an existing record
```
wpbonelli: oops, this is outdated, I dropped `save_event` completely

doc/ReleaseNotes/develop.toml
```diff
@@ -43,3 +43,8 @@ description = "When a horizontal flow barrier is placed between two cells, MODFL
 section = "fixes"
 subsection = "parallel"
 description = "The adaptive time stepping algorithm (ATS) was not working correctly in parallel mode. This applies to GWT and GWE models, when the model time step derived from the ATS\\_PERCEL parameter is not uniform across processes. In such a case, a deadlock occurred and the simulation was stalled. This fix ensures that the actual calculation time steps are synchronized across all processes such that they remain in lockstep."
+
+[[items]]
+section = "fixes"
+subsection = "model"
+description = "A PRT model could not previously be coupled via an exchange to a GWF model that uses adaptive time stepping (ATS). This is now supported."
```
aprovost-usgs: is the issue that it could not be coupled, or that when coupled to a GWF-ATS model it did not work properly (in that it did not reset the particles)?

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
```diff
@@ -213,6 +234,73 @@ logical function is_selected(this, event) result(selected)
...
+    integer(I4B) :: n, i
+    type(TrackRecordType) :: rec
+    type(ParticleTrackFileType) :: file
+
+    do n = 1, this%npending
+      rec = this%pending(n)
+      do i = 1, this%ntrackfiles
+        file = this%files(i)
+        if (file%iun > 0 .and. &
+            (file%iprp == -1 .or. file%iprp == rec%iprp)) &
```
aprovost-usgs: It looks like the ParticleTracks object is in charge of determining where the buffered records are written.  To do this, it needs to be aware of the kinds of things that can have an associated output file -- currently, the model (iprp == -1 ?) and the prp's.  just wondering whether there's a (reasonable) way to arrange this so that ParticleTracks doesn't need to be aware of that, and whether that would be desirable.  I.e., the model and the prp's (and whatever other objects creating track output there might hypothetically be in the future -- advanced packages?) decide for themselves which records they want in their output.  Maybe each even has its own buffer?  Could be something to chat about at some point.

src/Model/ParticleTracking/prt-prp.f90
```diff
@@ -61,6 +61,7 @@ module PrtPrpModule
     ! members
     type(PrtFmiType), pointer :: fmi => null() !< flow model interface
     type(ParticleStoreType), pointer :: particles => null() !< particle store
+    type(ParticleStoreType), pointer :: particles_old => null() !< old particle state (for ATS)
```
aprovost-usgs: From the point of view of anything within PRT, does it matter why the tracking is being retried (e.g., because of ATS, or because of Picard)?  If not, then maybe most comments (like this one) don't need to refer specifically to ATS, except perhaps in one place just to cite examples of why you might need this retry capability.  In other words, maybe the comments can be less ATS-specific overall.

src/Model/ParticleTracking/prt.f90
```diff
@@ -360,17 +363,25 @@ end subroutine prt_rp
...
     class(PrtModelType) :: this
     class(BndType), pointer :: packobj
     ! local
-    integer(I4B) :: irestore
     integer(I4B) :: ip, n, i
 
-    ! Reset state variable
-    irestore = 0
-    if (iFailedStepRetry > 0) irestore = 1
+    ! The model should advance just once per (step, ATS retry) pair.
```
aprovost-usgs: Mentioning ATS and Picard as specific examples of retries is good here.  However, instead of calling it a "(step, ATS retry) pair" up front, you might consider just "(step, retry) pair", and maybe even rename iFailedStepRetry to simply iRetry to keep it generic?

src/Model/ParticleTracking/prt-prp.f90
```diff
@@ -430,6 +435,24 @@ subroutine prp_ad(this)
     ! Coincident release times are merged to
     ! a single time by the release scheduler.
 
+    ! Save or restore particle state for ATS
```
aprovost-usgs: "for retry"?

src/Model/ParticleTracking/prt.f90
```diff
@@ -573,7 +588,10 @@ subroutine prt_ot(this)
     integer(I4B) :: ibudfl
     integer(I4B) :: ipflag
 
-    ! Note: particle tracking output is handled elsewhere
+    ! Flush buffered track events to disk. prt_ot is called once per
+    ! converged time step from Mf6FinalizeTimestep, outside the ATS
```
aprovost-usgs: "outside the retry loop"?

src/Solution/ParticleTracker/Particle/Particle.f90
```diff
@@ -339,6 +341,45 @@ subroutine put(this, particle, ip)
     this%icycwin(ip) = particle%icycwin
   end subroutine put
 
+  !> @brief Copy mutable particle state to this store from another one.
+  !!
+  !! Copies all mutable state arrays from source to this store. Copying
+  !! only mutable state (positions, status, tracking time, etc) without
+  !! immutable state (identity, options) saves space and time, but note
+  !! that it assumes this store already has all the immutable state; if
+  !! this routine is ever used beyond the context of ATS compatibility,
```
aprovost-usgs: "retry (e.g., ATS) compatibility"?

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
```diff
@@ -213,6 +234,73 @@ logical function is_selected(this, event) result(selected)
 
   end function is_selected
 
+  !> @brief Buffer a particle event for deferred write.
+  !! The buffer can be flushed to disk by flush_buffer,
+  !! or discarded by discard_buffer (e.g., on a failed
+  !! ATS retry). The buffer starts at 64 events, then
```
aprovost-usgs: this reference to ATS as an example is good i think

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
```diff
@@ -213,6 +234,73 @@ logical function is_selected(this, event) result(selected)
...
+        if (file%iun > 0 .and. &
+            (file%iprp == -1 .or. file%iprp == rec%iprp)) &
+          call save_record(file%iun, rec, file%csv)
+      end do
+    end do
+    this%npending = 0
+  end subroutine flush_buffer
+
+  !> @brief Discard buffered events.
+  !! Called from prt_ad at the start of each new ATS retry attempt,
```
aprovost-usgs: "each new tracking retry"?

aprovost-usgs (approved):
I have some comments/questions, but don't let them hold up the merge unless you feel there's more discussion needed

doc/ReleaseNotes/develop.toml
wpbonelli: Good point, it could be coupled but the results would be incorrect. Will improve the phrasing.

src/Model/ParticleTracking/prt-prp.f90
wpbonelli: Will remove mention of ATS here. It may not be the only use case for the old state snapshot in the future, in any case.

src/Model/ParticleTracking/prt.f90
wpbonelli: `iFailedStepRetry` comes from `SimVariablesModule`, so it'd need a local variable if we want to clean this up (which I'm not against doing)

Re: ATS vs Picard, I tried to clarify the comment. The PR description didn't really nail this either at first, I'll edit that too. I figure it may help to reserve "retry" to mean "ATS retry" and "rerun" or "repeated invocation" or similar for when advance is called more than once. Retry reruns need to run normally, but not Picard-triggered reruns.

src/Model/ParticleTracking/prt-prp.f90
wpbonelli: it does matter why the advance hook is being called though (detail in comment below)

src/Model/ParticleTracking/prt.f90
wpbonelli: maybe just not say anything here, and let it be explained elsewhere?

src/Solution/ParticleTracker/Particle/Particle.f90
wpbonelli: updated phrasing

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
wpbonelli: updated

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
wpbonelli: Nice idea.. Currently the model owns a single `ParticleTracksType` which knows about model- and PRP-scoped output files, which is awkward. Would make more sense for each to manage its own output. It would be really nice if model and package could select events independently. Need to consider how to configure that, since event selection is in OC.

wpbonelli:
#2789 should go before this

autotest/test_gwf_ats01.py
```diff
@@ -114,11 +109,18 @@ def build_models(idx, test):
     ic = flopy.mf6.ModflowGwfic(gwf, strt=strt)
 
     # node property flow
-    npf = flopy.mf6.ModflowGwfnpf(gwf, save_flows=False, icelltype=laytyp, k=hk)
+    npf = flopy.mf6.ModflowGwfnpf(
+        gwf,
+        save_flows=True,
```
wpbonelli, line 114: this test is modified with `save_flows=True` so PRT tests can reuse the GWF setup

autotest/test_gwf_ats01.py
```diff
@@ -156,7 +158,7 @@ def build_models(idx, test):
         budget_filerecord=f"{gwfname}.cbc",
         head_filerecord=f"{gwfname}.hds",
         headprintrecord=[("COLUMNS", 10, "WIDTH", 15, "DIGITS", 6, "GENERAL")],
-        saverecord=[("HEAD", "ALL")],
+        saverecord=[("HEAD", "ALL"), ("BUDGET", "ALL")],
```
wpbonelli, line 161: same reason as above

autotest/test_prt_exg.py
```diff
@@ -34,25 +34,21 @@
     FlopyReadmeCase,
     check_budget_data,
     check_track_data,
+    get_model_name,
```
wpbonelli, line 37: just some unrelated/opportunistic cleanup in this test

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
wpbonelli: prototyped here https://github.com/MODFLOW-ORG/modflow6/compare/develop...wpbonelli:modflow6:prt-event-selection, will open a PR at some point

PR #2754, opened 2026-04-04
fix(fmi): trap invalid tdis in final stress period

wpbonelli:
FMI conducts TDIS compatibility checks during each stress period to ensure that the coupled transport or tracking model's time discretization is compatible with that of the flow model. In particular, if a flow model stress period has just one time step, the transport or tracking model period may have any number of time steps. However, if a flow model stress period has more than one time step, the transport or tracking model must have the same number of time steps.

Simulations would previously stop with an error if the transport or tracking model violated this rule for any stress period up to the simulation's last, but would not raise an error for the last stress period. Update the compatibility check to cover the last stress period. Consolidate the error message and make the language generic (it referred specifically to transport models before). And update `test_gwt_fmi02.py` to use `TestFramework` and exercise the error traps.

___

### DIFF
--- a/autotest/test_gwt_fmi02.py
+++ b/autotest/test_gwt_fmi02.py
@@ -1,20 +1,40 @@
-"""Tests to ability to run flow model first followed by transport model"""
+"""
+Tests ability to run a transport model in an independent
+simulation, consuming flows with a flow model interface.
 
-import os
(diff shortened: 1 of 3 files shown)


christianlangevin (approved):
Nice find, Wes.  Nice clean fix.

PR #2757, opened 2026-04-08
fix(exg): trap multiple connections per cell pair

wpbonelli:
Warn if any pair of connected cells in a like-model exchange has multiple connections. This will be an error for PRT-PRT exchanges, but multiple connections are tolerated for other models for flexibility.

___

### DIFF
--- a/src/Exchange/exg-gwegwe.f90
+++ b/src/Exchange/exg-gwegwe.f90
@@ -249,7 +249,9 @@ contains
   subroutine validate_exchange(this)
     ! -- dummy
     class(GweExchangeType) :: this !<  GweExchangeType
-    !
+    ! -- local
+    integer(I4B) :: n, m, ncount
+    logical(LGP), dimension(:), allocatable :: checked_conns
 
     ! Ensure gwfmodel names were entered
     if (this%gwfmodelname1 == '') then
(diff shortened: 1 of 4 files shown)


wpbonelli:
Re: deciding whether this should be a general restriction or only apply in some cases, @mjr-deltares figured that people may be using DISU in nonstandard ways involving multiple connections between the same two cells. E.g.

> 90% of the shared face has a low conductivity and the 10% has a high conductivity to model some kind of geological feature or artifact from mining

aprovost-usgs:
> Re: deciding whether this should be a general restriction or only apply in some cases, @mjr-deltares figured that people may be using DISU in nonstandard ways involving multiple connections between the same two cells. E.g.
> 
> > 90% of the shared face has a low conductivity and the 10% has a high conductivity to model some kind of geological feature or artifact from mining

Guess it comes down to whether the risk of a problem due to accidentally duplicative connections going unnoticed outweighs the benefit for creative modeling like @mjr-deltares described.  Maybe consider a warning?

wpbonelli:
> Maybe consider a warning?

Switched to a warning

src/Exchange/exg-gwegwe.f90
```diff
@@ -297,6 +302,31 @@ subroutine validate_exchange(this)
       call store_error(errmsg)
     end if
     !
+    ! At most one connection is allowed between any two cells.
```
aprovost-usgs, line 305: is this comment still true given that the condition now triggers a warning, not an error?  (there are several instances of this comment)

src/Exchange/exg-gwegwe.f90
```diff
@@ -297,6 +302,31 @@ subroutine validate_exchange(this)
...
+    allocate (checked_conns(this%nexg))
+    checked_conns = .false.
+    do n = 1, this%nexg
+      if (checked_conns(n)) cycle
+      ! Skip connections with unresolved nodes (model on remote process)
+      if (this%nodem1(n) == -1 .or. this%nodem2(n) == -1) cycle
+      ncount = 0
+      do m = 1, this%nexg
```
aprovost-usgs, line 313: it's basically a search for duplicates in a list of (nodem1, nodem2).  the number of comparisons here is on the order of nexg^2.  if this proves to be slow when there are lots of exchange connections, maybe consider a more efficient approach?

short of that, the number could be (almost) cut in half by looping over
do n = 1, this%nexg - 1
and
do m = n + 1, this%nexg
(though this would still scale as nexg^2)

aprovost-usgs (commented):
some comments in case we decide to reopen this in the future

PR #2759, opened 2026-04-11
feat(memory): support param input context deallocation

mjreno:
Support memory manager per parameter input context deallocation before the numerical model is run.  Significant (>10%) reduction in allocated memory possible; most input context memory not read over the course of the simulation is released before it begins. There are 2 mechanisms:
  - `mem_set_value()` interfaces now take an optional release argument that deallocates memory after copy
  - added `memorystore_release()` which takes arguments for any parameter memory location

Memory counters are updated as part of the deallocation so that memory usage reflects the adjustment.  For comparison, here are local usage (`mfsim.lst`) comparisons with the develop branch for 2 autotests:
  - `npf01a_75x75`: Total => develop: 3.72361 MB;  branch:  3.33675 MB  | difference: 0.38686 MB (10.4%)
  - `csub_subwt02a`: Total => develop: 1.08720 MB; branch:  0.91232 MB  │ difference: 0.17488 MB (16.1%)
 
- [X] Added doxygen comments to new and modified procedures

### DIFF
--- a/src/Exchange/DisConnExchange.f90
+++ b/src/Exchange/DisConnExchange.f90
@@ -251,6 +251,7 @@ contains
   subroutine source_data(this, iout)
     ! -- modules
     use MemoryManagerModule, only: mem_setptr
+    use MemoryManagerExtModule, only: memorystore_release
     ! -- dummy
     class(DisConnExchangeType) :: this !< instance of exchange object
     integer(I4B), intent(in) :: iout !< the output file unit
@@ -389,6 +390,15 @@ contains
       call store_error('Errors encountered in exchange input file.')
       call store_error_filename(this%filename)
     end if
+    !
+    call memorystore_release('CELLIDM1', this%input_mempath)
+    call memorystore_release('CELLIDM2', this%input_mempath)
+    call memorystore_release('IHC', this%input_mempath)
+    call memorystore_release('CL1', this%input_mempath)
+    call memorystore_release('CL2', this%input_mempath)
+    call memorystore_release('HWVA', this%input_mempath)
+    call memorystore_release('AUXVAR', this%input_mempath)
+    call memorystore_release('BOUNDNAME', this%input_mempath)
   end subroutine source_data
 
   !> @brief Allocate scalars and initialize to defaults
--- a/src/Exchange/DisConnExchange.f90
+++ b/src/Exchange/DisConnExchange.f90
@@ -104,12 +104,13 @@ contains
     integer(I4B) :: ival, n
     !
     ! -- update defaults with idm sourced values
-    call mem_set_value(this%naux, 'NAUX', this%input_mempath, found%naux)
+    call mem_set_value(this%naux, 'NAUX', this%input_mempath, found%naux, &
+                       release=.false.)
     call mem_set_value(this%ipakcb, 'IPAKCB', this%input_mempath, found%ipakcb)
     call mem_set_value(this%iprpak, 'IPRPAK', this%input_mempath, found%iprpak)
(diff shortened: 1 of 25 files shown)


src/Exchange/DisConnExchange.f90
```diff
@@ -104,12 +104,13 @@ subroutine source_options(this, iout)
     integer(I4B) :: ival, n
     !
     ! -- update defaults with idm sourced values
-    call mem_set_value(this%naux, 'NAUX', this%input_mempath, found%naux)
+    call mem_set_value(this%naux, 'NAUX', this%input_mempath, found%naux, &
```
mjr-deltares, line 107: at this point I am not sure I understand why these are no released, maybe discuss

src/Model/Discretization/Disu.f90
```diff
@@ -901,6 +904,11 @@ subroutine source_cell2d(this)
       write (this%iout, '(1x,a)') 'Discretization Cell2d data loaded'
     end if
     !
+    call memorystore_release('ICELL2D', this%input_mempath)
```
mjr-deltares, line 907: see comment with DISV

src/Model/Discretization/Disv.f90
```diff
@@ -543,6 +544,8 @@ subroutine source_vertices(this)
       write (this%iout, '(1x,a)') 'Discretization Vertex data loaded'
     end if
     !
+    call memorystore_release('XV', this%input_mempath)
```
mjr-deltares, line 547: > I think this will cause an issue with netcdf? We want to use these input data in add_mesh_data in disvncmesh.f90 which is called not until export_cr in Mf6Initialize. The DISV source loading happens from simulation_cr => source_simulation_nam => models_create => ... => gwf_cr for example, which is before.

> Let's discuss the best strategy here. Maybe first one to touch the input secures it in permanent storage and releases, so after that, other components would need to get it from the permanent storage? Or, a release of memory is not done until after all initialization is complete, though I am not sure that's practical...

Argghhh. I should have looked through the entire PR more carefully before starting this... Of course, you had already thought about that and changed it. I guess you went with the first strategy which has my preference as well. Sorry about that.

src/Model/ModelUtilities/BoundaryPackageExt.f90
```diff
@@ -306,12 +306,13 @@ subroutine source_options(this)
     integer(I4B) :: n
     !
     ! -- update defaults with idm sourced values
-    call mem_set_value(this%naux, 'NAUX', this%input_mempath, found%naux)
+    call mem_set_value(this%naux, 'NAUX', this%input_mempath, found%naux, &
```
mjr-deltares, line 309: I guess same question here as above

src/Utilities/Memory/MemoryManager.f90
```diff
@@ -796,7 +797,7 @@ subroutine allocate_dbl(sclr, name, mem_path)
     end if
     !
     ! -- update counter
-    nvalues_aint = nvalues_aint + 1
+    nvalues_adbl = nvalues_adbl + 1
```
mjr-deltares, line 800: great catch! :-)

mjr-deltares (changes requested):
I think this is a great improvement! I'd like to discuss a few minor things. But, what's maybe most important is if we can define (and document) a consistent way of dealing with this new mechanism to avoid doubling up or maybe prematurely releasing memory.

src/Exchange/DisConnExchange.f90
mjr-deltares: OK, so it is for IDM internals you explained me, for the stress period loads

mjr-deltares (approved):

PR #2764, opened 2026-04-14
feat(fmi) Expose the cell saturated fraction to the other models

Manangka:
In the Transport and Energy models the saturation, $S_w$, is required in the mst and est packages.
However these values aren't exposed through the FMI. 
Instead they are computed using the gwfstrgss and gwfstrgsy rates.

mst:
 $f_{storage} = \frac{ \partial \left(S_w \theta C\right)}{\partial t}$

There are several reasons to switch to using the saturation directly:
- The code becomes clearer. We can now recognize the equation as is stated in the documentation
- The computation from rate to cell saturation works fine for confined cases, but when we are working with unconfined cases their is an additional requirement that needs to be fulfilled i.e. Sy = theta
- My main motive for this change is In preparation of the higher order time integration. For that i need to store saturation for multiple older timesteps (or compute it using gwfstrgss  and gwfstrgsy but that means that i have to store those for multiple time steps). This change makes it possible to do.

**Note**
After this change we can almost entirely get rid of gwfstrgss and gwfstrgsy in the FMI. It only being used in 1 place in the prt-fmi file.

- [ ] Closed issue #xxxx

### DIFF
--- a/src/Model/Connection/GwtInterfaceModel.f90
+++ b/src/Model/Connection/GwtInterfaceModel.f90
@@ -118,6 +118,8 @@ contains
                       this%fmi%memoryPath)
     call mem_allocate(this%fmi%gwfsat, this%neq, 'GWFSAT', &
                       this%fmi%memoryPath)
+    call mem_allocate(this%fmi%gwfsat_old, this%neq, 'GWFSAT_OLD', &
+                      this%fmi%memoryPath)
     call mem_allocate(this%fmi%gwfspdis, 3, this%neq, 'GWFSPDIS', &
                       this%fmi%memoryPath)
   end subroutine allocate_fmi
--- a/src/Model/GroundWaterEnergy/gwe-est.f90
+++ b/src/Model/GroundWaterEnergy/gwe-est.f90
@@ -227,11 +227,10 @@ contains
       !
       ! -- calculate new and old water volumes and solid volume
       vcell = this%dis%area(n) * (this%dis%top(n) - this%dis%bot(n))
-      vnew = vcell * this%fmi%gwfsat(n) * this%porosity(n)
-      vold = vnew
-      if (this%fmi%igwfstrgss /= 0) vold = vold + this%fmi%gwfstrgss(n) * delt
-      if (this%fmi%igwfstrgsy /= 0) vold = vold + this%fmi%gwfstrgsy(n) * delt
       vsolid = vcell * (DONE - this%porosity(n))
+
+      vnew = vcell * this%fmi%gwfsat(n) * this%porosity(n)
+      vold = vcell * this%fmi%gwfsat_old(n) * this%porosity(n)
       !
       ! -- add terms to diagonal and rhs accumulators
       term = (this%rhos(n) * this%cps(n)) * vsolid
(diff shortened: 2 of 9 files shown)


src/Model/ModelUtilities/FlowModelInterface.f90
```diff
@@ -698,6 +702,7 @@ subroutine advance_bfr(this)
               nr = this%dis%get_nodenumber(nu, 0)
               if (nr <= 0) cycle
               this%gwfsat(nr) = this%bfr%auxvar(1, i)
+              this%gwfsat_old(nr) = this%bfr%auxvar(1, i)
```
mjr-deltares, line 705: If you want to do it this way, you also have to write sat_old to the budget file, see the sav_sat subroutine in gwf-npf.f90. But technically, sat_old is already in the file though...

src/Model/Connection/GwtInterfaceModel.f90
```diff
@@ -118,6 +118,8 @@ subroutine allocate_fmi(this)
                       this%fmi%memoryPath)
     call mem_allocate(this%fmi%gwfsat, this%neq, 'GWFSAT', &
                       this%fmi%memoryPath)
+    call mem_allocate(this%fmi%gwfsat_old, this%neq, 'GWFSAT_OLD', &
```
mjr-deltares, line 121: Note that the data also has to be copied in to the interface model. This is done in GwtGwtConnection.f90, for example, for gwfsat you see there:

`call this%cfg_dv('GWFSAT', 'FMI', SYNC_NDS, (/STG_BFR_EXG_AD/))`

which copies gwfsat at the stage STG_BFR_EXG_AD. 

And, for parallel, this data might not be there. So we also have to add it to the MPI sync. It should be simple, just follow the whereabouts of this variable in VirtualGwtModel:

`type(VirtualDbl1dType), pointer :: fmi_gwfsat => null()`

and the same for VirtualGweModel. That, and adding a parallel test of course, should get you there.

mjr-deltares (changes requested):
It looks clean, generally OK I'd say, but those 2 comments below need some follow up.

src/Model/GroundWaterEnergy/gwe-est.f90
```diff
@@ -227,11 +227,10 @@ subroutine est_fc_sto(this, nodes, cold, nja, matrix_sln, idxglo, rhs)
       !
       ! -- calculate new and old water volumes and solid volume
       vcell = this%dis%area(n) * (this%dis%top(n) - this%dis%bot(n))
-      vnew = vcell * this%fmi%gwfsat(n) * this%porosity(n)
-      vold = vnew
-      if (this%fmi%igwfstrgss /= 0) vold = vold + this%fmi%gwfstrgss(n) * delt
-      if (this%fmi%igwfstrgsy /= 0) vold = vold + this%fmi%gwfstrgsy(n) * delt
       vsolid = vcell * (DONE - this%porosity(n))
+
+      vnew = vcell * this%fmi%gwfsat(n) * this%porosity(n)
```
christianlangevin, line 232: In this recalculation of `vnew` and `vold` does it give the same results as the calculations that you replaced?  The previous equations calculate based on compressible and pore volume changes.  I feel like this new approach does not account for changes in compressible water volume.  For a confined aquifer, it looks like vnew and vold will always be the same.

christianlangevin (changes requested):
I'm not sure this approach is set up to handle changes in compressible storage.  Of course, this is often quite small if reasonable values are used for specific storage.  Also, as I'm sure you know, a similar approach is used for the advanced transport packages.

Manangka:
@christianlangevin @mjr-deltares I'm putting this PR on a hold for now till our meeting next week

PR #2773, opened 2026-04-15
fix(disu): use nodesuser for CELLID range check in get_nodenumber_idx1

### LINKED ISSUE #2772: HFB CELLID range/connectivity validation changed between 6.4.4 and 6.6.3 for DISU + IDOMAIN
## Summary

For a DISU model with `IDOMAIN` inactive cells, the HFB package behaves
inconsistently across MF6 versions. The same DIS + HFB input files that run
cleanly in 6.4.4 fail in 6.6.3 and 6.7.0. `mf6io.pdf` documents HFB `CELLID`
for DISU as a user node number, which matches 6.4.4 behavior.

## Tested versions

- MF6 6.4.4 (02/13/2024) — works
- MF6 6.6.3 (09/29/2025) — fails (see Issue 1)
- MF6 6.7.0 (02/05/2026) — fails (see Issues 1 and 2)

## Model

Quadtree DISU, `NODES 1575000`, `NJA 10981930`, 1,332,698 active cells
(242,302 with `IDOMAIN=0`). HFB contains 9,354 barrier pairs written with
full-grid user node IDs (1..1,575,000). Generated by Groundwater Vistas.

## Issue 1 — Range validation uses active-cell count, not declared NODES

6.6.3 and 6.7.0 reject any HFB `CELLID` whose user-node value exceeds the
active-cell count:

```
ERROR REPORT:
  1. Node number (1464588) is less than 1 or greater than nodes (1332698).
```

The "nodes" bound in the error equals the active-cell count, not the DIS-
declared \`NODES 1575000\`. 6.4.4 accepts the same file. Either the validator
regressed or the docs/IO spec for HFB CELLIDs with IDOMAIN need to clarify
that users must now supply reduced node IDs.

## Issue 2 — Reduced CELLIDs flagged as "unconnected" in 6.7.0

Manually reducing CELLIDs (pack active cells into 1..n_active in user-index
order) gets past the Issue 1 range check in both 6.6.3 and 6.7.0. In 6.7.0,
~1,920 of the 9,354 GWV pairs then fail with:

```
ERROR REPORT:
  1.  HFB no. 1 is between two unconnected cells: (1227971) and (1227963)
  ...
```

Both endpoints of each failing pair are active, and the pair appears in the
DIS JA for both nodes (verified via flopy / the input DIS file). Something
in the reduced-JA construction for quadtree DISU + IDOMAIN appears to drop
connections it should retain.

aacovski:
## Summary

Fixes #2772.

`Disu%get_nodenumber_idx1` checked `nodeu > this%nodes` (reduced/active count) instead of `nodeu > this%nodesuser`. The equivalent checks in `Disv.f90:937` and `Disv2d.f90:878` already use `nodesuser`, and the function indexes `this%nodereduced` (sized `nodesuser`) on the next line, so the original bound was inconsistent with both the function's contract and the sibling implementations.

The bug was latent until a caller passed `icheck=1`. The HFB rewrite (now using `dis%get_nodenumber(nodeu, 1)` directly rather than going through `noder_from_cellid`, which uses `icheck=0`) exposed it: HFB `CELLID`s using user node numbers above the active-cell count are rejected with "Node number (N) is less than 1 or greater than nodes (n_active)", even when the cell is active and the user node number is valid per the MF6 IO guide.

## Changes

- `src/Model/Discretization/Disu.f90`: change the bound from `this%nodes` to `this%nodesuser` (one place, two lines).
- `autotest/test_gwf_disu_hfb_idomain.py`: regression test. Builds a 3x3x3 DISU via `get_disu_kwargs`, deactivates one cell so `n_active = 26 < n_user = 27`, then adds an HFB barrier between user cells 26 and 27 (both active, laterally adjacent in the bottom layer). Pre-fix: errors with `Node number (27) is less than 1 or greater than nodes (26)`. Post-fix: runs cleanly.

## Test plan

- [x] Regression test reproduces the failure on stock 6.7.0 (verified locally with the same model setup).
- [ ] CI runs the full autotest suite.

### DIFF
--- a/autotest/test_gwf_disu_hfb_idomain.py
+++ b/autotest/test_gwf_disu_hfb_idomain.py
@@ -1,0 +1,86 @@
+"""
+Regression test for issue #2772.
+
+Build a 3x3x3 DISU model, deactivate one cell via IDOMAIN, then add an
+HFB barrier whose CELLIDs use user (full-grid) node numbers that exceed
+the active-cell count. Both endpoints of the barrier are active and
+adjacent. The HFB IO guide documents CELLID for DISU as a user node
+number, so this should run cleanly.
+
+Prior to the fix in `Disu.f90:get_nodenumber_idx1`, the range check used
+`this%nodes` (reduced/active count) rather than `this%nodesuser`,
+inconsistent with the equivalent checks in DISV/DISV2D. Reading the HFB
+file then failed with:
+
+    Node number (N) is less than 1 or greater than nodes (n_active).
+"""
+
+from pathlib import Path
+
+import flopy
+import numpy as np
+import pytest
+from flopy.utils.gridutil import get_disu_kwargs
+from framework import TestFramework
+
+cases = ["disuhfb"]
+
+
+def build_models(idx, test):
+    name = cases[idx]
+    ws = test.workspace
+    nlay, nrow, ncol = 3, 3, 3
+    delr = 10.0 * np.ones(ncol)
+    delc = 10.0 * np.ones(nrow)
+    top = 0.0
+    botm = [-10.0, -20.0, -30.0]
+    disukwargs = get_disu_kwargs(nlay, nrow, ncol, delr, delc, top, botm)
+    # Deactivate user node 2 so n_active = 26 < n_user = 27
+    idomain = np.ones((nlay, nrow * ncol), dtype=int)
+    idomain[0, 1] = 0
+    disukwargs["idomain"] = idomain
+
+    sim = flopy.mf6.MFSimulation(
+        sim_name=name, version="mf6", exe_name="mf6", sim_ws=ws
+    )
+    flopy.mf6.ModflowTdis(sim)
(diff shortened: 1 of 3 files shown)


christianlangevin:
Really nice fix @aacovski.  Thanks for this.  Looks like the PR is working its way through the system.  Can you also add a bug fix description?  Maybe something like the following?  Thanks!

```
[[items]]
section = "fixes"
subsection = "internal"
description = "When a horizontal flow barrier is placed bewteen two cells, MODFLOW checks to ensure that the cells are adjacent.  This check did not function correctly for DISU grids, in some cases, if the IDOMAIN functionality was used to remove cells.  The check for two adjacent DISU cells was corrected to work for IDOMAIN functionality."
```

aacovski:
Thanks! Added the release note in 964b432 (using your suggested wording verbatim).

christianlangevin (approved):

PR #2776, opened 2026-04-17
refactor: add hash to development version string

wpbonelli:
Use meson [`vcs_tag()`](https://mesonbuild.com/Reference-manual_functions_vcs_tag.html) to embed the latest commit hash into the version string emitted by `mf6 -v` for development builds, e.g.

```bash
$ mf6 -v
mf6: 6.8.0.dev0+4e4fdebc
```

On simulation runs the header output will be like

```bash
                                   MODFLOW 6
                U.S. GEOLOGICAL SURVEY MODULAR HYDROLOGIC MODEL
              VERSION 6.8.0.dev0+a7dcc8ba (preliminary) 02/06/2026
                               ***DEVELOP MODE***
```

For release versions, the version string remains simply `x.y.z`. This works by trying `git describe --exact-match --tags HEAD`, which yields the tag name if HEAD is on a tag (as it is on release branches), otherwise using `git rev-parse --short HEAD` to get a short SHA suffix for development builds.

This makes it easy to see at a glance which exact version of the source a development build is built from. (Worth noting however that the hash will not reflect any uncommitted changes to files going into a build.) This facilitates `flopy.mf6` module introspection; if flopy stores the version the module was generated from, it can check if it's compatible with a given binary by checking its version string.

The build qualifier/date i.e. "(preliminary) MM/DD/YYYY" remain hardcoded, only to be updated by `distribution/update_version.py`. It could be confusing that the hash might be much more recent than the date following it. The date/qualifier could be done automatically too, but I figure they serve a somewhat different purpose, marking when we cut a release or start a new development cycle, while the addition of the hash is just a clean way to match binaries to the source they are built from.

The static `version.f90` is preserved for compatibility with MSVS and make. ~~This PR sets up pre-build hooks for MSVS files to accomplish the same thing as meson `vcs_tag()`, assuming Python is on the PATH. If not, the static version file is used. Note that MSVS developers will need to omit the automatic changes to the static `version.f90` when checking files in.~~


wpbonelli:
Decided after discussion not to try the MSVS pre-build hooks, just use the static `version.f90` for MSVS, like for make.

wpbonelli:
@mwtoews if you have thoughts on this, would be interested to hear. We mentioned possibly doing this for flopy and the other python projects a while ago, via setuptools_scm or versioneer or similar. But I think it buys us more immediate value here in mf6 hence trying it here first.

wpbonelli:
After some discussion with @mjr-deltares, it seems ideal to drop the qualifier/date" (preliminary) MM/DD/YYYY", at least for development builds, if not entirely. There is already a build timestamp immediately below the header section, like

```
MODFLOW 6 compiled Apr 18 2026 05:21:36 with GCC version 13.4.0
```

This almost always disagrees with the hardcoded date, since we update the latter only for releases and when a development cycle begins. IMO, this is confusing. And even for release builds, where the two dates match, it's redundant.

And "(preliminary)" is redundant information, since we already have multiple indications that a build is not an official version: "DEVELOP MODE", the ".devN" suffix, plus the hash, after this PR. While these are not identical (we could have a "release mode" build that still has ".devN+hash") they all indicate that the build is not an official release.

With the suggested simplifications, the header section could look like this for dev builds

```
                                   MODFLOW 6
                U.S. GEOLOGICAL SURVEY MODULAR HYDROLOGIC MODEL
                          VERSION 6.8.0.dev0+a7dcc8ba
                               ***DEVELOP MODE***

        MODFLOW 6 compiled Apr 18 2026 05:21:36 with GCC version 13.4.0
```

And for release builds

```
                                   MODFLOW 6
                U.S. GEOLOGICAL SURVEY MODULAR HYDROLOGIC MODEL
                                 VERSION 6.8.0

        MODFLOW 6 compiled Apr 18 2026 05:21:36 with GCC version 13.4.0
```

christianlangevin:
Hey @wpbonelli, I hear you on the confusion with the dates in the header.  The one thing I would add is that these "official" dates are important to MODFLOW users for releases.  I see them as more of title that goes with the official version number.  And while there is redundancy with the build date, that can sometimes be a good thing.  If they are not in sync, then that indicates a possible mismatch between the official version and when it was built.  My 2 cents!

image

wpbonelli:
Thanks @christianlangevin I see the value of the separate release date now.

Martijn suggested the date could become `--/--/----` or `xx/xx/xxxx` in development builds, and filled only for releases. Making it clearer that it's a release date, part of a label applied to the release, distinct from the build timestamp, which as you say can be checked against the release date.

Or we could just omit the release date for development builds and include it only for releases.

The variable names in `version.f90` could in any case reflect the distinction between a normal version string, that just contains the number and an optional VCS tag, and a "full" version string with a title section as well.

```fortran
  character(len=*), parameter :: VERSIONNUMBER = '6.8.0.dev0'
  character(len=*), parameter :: VERSIONVCSTAG = "@VCS_TAG@"
  character(len=*), parameter :: VERSIONTITLE = ''
  character(len=*), parameter :: VERSION = VERSIONNUMBER//VERSIONVCSTAG
  character(len=*), parameter :: FULLVERSION = &
                                 VERSIONNUMBER// &
                                 VERSIONVCSTAG// &
                                 VERSIONTITLE
```

Should we keep "preliminary" in the title for dev builds?

mwtoews:
Embedding commit SHA info from `vcs_tag()` into some visible version info is a great way to go, as it could help identify and compare revisions between releases.

Further to this is another idea to create a (currently non-existing) [mf6 container](https://github.com/MODFLOW-ORG/pkgs/mf6) to run a single version commit in [Docker](https://www.docker.com/) containers hosted by [ghcr.io](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry). Maybe I can work-up a demo on my personal repo, but what it would look like is:

    docker run -it --rm -v /local/path/to/sim:$(pwd) -w $(pwd) ghcr.io/MODFLOW-ORG/mf6:4d7f41a

which would run commit version 4d7f41ab902c7cb6b0b75fa265f32912e5487b24 on a simulation in `/local/path/to/sim`. If there was a bug or issue with that version of mf6, then a different SHA would be used to run the simulation with a slightly different version of mf6. Having the SHA info visible in the program output would be indeed useful to retain.

A specific example where this is done is [this program](https://github.com/libgeos/geos/pkgs/container/geosop) automated with [this action](https://github.com/libgeos/geos/blob/main/.github/workflows/container.yml) and [Dockerfile](https://github.com/libgeos/geos/blob/main/tools/ci/geosop/Dockerfile).

Is there any interest? Does anyone else use Docker?

---

_Update:_ prototype done, and it didn't take too much effort. See https://github.com/mwtoews/modflow6/pkgs/container/modflow6

To run it, install Docker, you many need to configure some permission stuff. The default entrypoint is `mf6`, so you can see the version like this:
```
$ docker run -it --rm ghcr.io/mwtoews/modflow6:bb7483c508 -v

mf6: 6.8.0.dev0 (preliminary) 02/06/2026
```
Or with the last SHA in this PR commit (bb7483c508743b3317a2a21851aacc2d180ec25d):
```
$ docker run -it --rm ghcr.io/mwtoews/modflow6:bb7483c508 -v

mf6: 6.8.0.dev0+bb7483c508
```

wpbonelli:
After some more discussion, we concluded the following. The version string consists of three sections

  - version number
  - VCS tag (optional)
  - title (optional)

The version number will just be `x.y.z` for releases, plus `.devN` for develop builds.

The VCS tag section will be empty for releases, `+sha` (short hash) for develop builds, with optional `.dirty` suffix depending whether the build includes uncommitted changes (a common convention used by e.g. git itself).

The title will be empty for develop builds, and just contain the release date ` MM/DD/YYYY` for releases. It will no longer include "preliminary" for develop builds.

So, release version header looks like

```
                                   MODFLOW 6
                U.S. GEOLOGICAL SURVEY MODULAR HYDROLOGIC MODEL
                            VERSION 6.8.0 04/29/2026

        MODFLOW 6 compiled Apr 29 2026 06:06:11 with GCC version 13.4.0
```

Clean develop version, no uncommitted changes

```
                                   MODFLOW 6
                U.S. GEOLOGICAL SURVEY MODULAR HYDROLOGIC MODEL
                          VERSION 6.8.0.dev0+752420f1
                               ***DEVELOP MODE***

        MODFLOW 6 compiled Apr 29 2026 06:03:09 with GCC version 13.4.0
```

Develop version with uncommitted changes

```
                                   MODFLOW 6
                U.S. GEOLOGICAL SURVEY MODULAR HYDROLOGIC MODEL
                          VERSION 6.8.0.dev0+752420f1.dirty
                               ***DEVELOP MODE***

        MODFLOW 6 compiled Apr 29 2026 06:03:09 with GCC version 13.4.0
```

emorway-usgs (approved):
There's quite a bit here!  Thanks for working on this @wpbonelli, it will be nice to have this additional information on builds that are not from a specific release.  If everything is looking good in preliminary testing on a local machine, then I think it is OK to bring in.  If it looks like there are few hiccups the next time we do an official release, I think it will be OK to massage it further at that time (or maybe you've already mimicked a full release build locally and it appears to be working just fine).

jdhughes-dev (approved):

distribution/vcs_tag_suffix.py
```diff
@@ -0,0 +1,45 @@
...
+            .decode()
+            .strip()
+        )
+        dirty = (
+            subprocess.run(
+                ["git", "diff", "--quiet", "HEAD"], capture_output=True
+            ).returncode
+            != 0
+        )
+        return f"+{sha}.dirty" if dirty else f"+{sha}"
```
mwtoews, line 36: This method will fail if either `git` is not installed or this is not from a GIT repo.
For example, take this PR as a ZIP file: `wget -O head.zip https://api.github.com/repos/MODFLOW-ORG/modflow6/zipball/pull/2776/head`.

Perhaps the first part of this method should do a quick sanity check:
```python
try:
    subprocess.run(["git", "status"], capture_output=True, check=True)
except (subprocess.CalledProcessError, OSError):
    return ""
```
The fallback version for this scenario would then show `mf6: 6.8.0.dev0`

wpbonelli: True. I figured it was safe to assume git is available, didn't consider that. Thanks!

mwtoews (approved):

christianlangevin (approved):

mjr-deltares (approved):

PR #2780, opened 2026-04-20
doc(ems): clarify model/solution compatibility

### LINKED ISSUE #2778: Clarify which model types with which the iterative model solution can be applied
**Is your feature request related to a problem? Please describe.**
The description of the iterative model solution states "The model solution will solve all of the models that are added to it." However, none of the PRT models use IMS. If IMS can not be used with PRT it would be better to state that explicitly.

**Describe the solution you'd like**
A statement such as "The implicit model solution presently works with the Groundwater Flow (GWF), Groundwater Transport (GWT) and Groundwater Energy Transport (GWE) Models.

**Describe alternatives you've considered**
An alternative would be to state that IMS does not work with PRT models.

wpbonelli:
Close #2778


rbwinst-usgs:
It looks good to me.

emorway-usgs (approved):
Nice write-up.

aprovost-usgs (commented):
We'll need to remember to update the lists of compatible models for each solution as new models are added. An alternative approach might be to mention specific models here only as examples, not an exhaustive list, and to state in the description of each model what kind of solution(s) it admits. Pros and cons either way. I'm ok with it as is.

wpbonelli:
#2788 aims to address your comment @aprovost-usgs

PR #2788, opened 2026-04-23
docs: move model solution requirement to model sections

wpbonelli:
Followup on #2780. Use generic language in the IMS/EMS docs so we don't have to update these sections to include new model types in future, and specify which solution type is required in each specific model section.


aprovost-usgs (approved):
exactly what i had in mind, thanks

PR #2789, opened 2026-04-24
feat(prt): configurable event buffering

wpbonelli:
Previously PRT model particle events were written to output file(s) immediately. This was a deliberate choice motivated by MP7 suffering OOM errors when consolidating internal scratch files before writing the final output file.

With GWF models using adaptive time stepping (ATS), or with multiple Picard iterations when PRT and GWF are in the same solution group, output would be written for discarded solve attempts. Events should only end up in output files for the final, successful solve.

To gate events until a successful solve, buffer them until it's known whether they should be written. Introduce two buffering strategies, in memory and scratch file, with a new OC option `SCRATCH_BUFFER` enabling the latter. Memory buffering is used by default.

Since the buffer could conceivably grow large for models with lots of particles or lots of action per time step (e.g., the final time step when extended tracking is on), the scratch file approach is provided to let the modeler avoid OOM errors at the price of more file IO.

While this departs from informal MF6 convention to avoid scratch files in general, it's a well motivated option, not on by default.

This is a pure refactor but for the new option, which I figure merits calling it a feature.  It was split out of #2753. The retry misbehavior is fixed in that PR, not here.

___


src/Model/ParticleTracking/prt.f90
```diff
@@ -262,6 +262,9 @@ subroutine prt_ar(this)
     call this%oc%oc_ar(this%dis, DHNOFLO)
     call this%budget%set_ibudcsv(this%oc%ibudcsv)
 
+    ! Initialize the event buffer (memory or scratch file per OC option)
+    call this%tracks%init_buffer(this%oc%scratch_buffer)
```
Manangka, line 266: Would `use_scratch_buffer` be a better name?

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
```diff
@@ -210,99 +287,175 @@ logical function is_selected(this, event) result(selected)
...
+  end subroutine buffer_event
+
+  !> @brief Flush the event buffer to disk.
+  subroutine flush_buffer(this)
+    class(ParticleTracksType) :: this
+    call this%buffer%flush(this%files)
+  end subroutine flush_buffer
+
+  !> @brief Discard buffered events without writing.
+  subroutine discard_buffer(this)
```
Manangka, line 241: I might be mistaken but I don't see where this method is being called. Is it in use?

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
```diff
@@ -210,99 +287,175 @@ logical function is_selected(this, event) result(selected)
...
+    class(ParticleTracksType) :: this
+    call this%buffer%discard()
+  end subroutine discard_buffer
+
+  subroutine buffer_init(this)
+    class(ParticleTrackEventBufferType) :: this
+  end subroutine buffer_init
+
+  subroutine memory_append(this, rec)
```
Manangka: This seems to be working like a c++ vector. We have a STLVecInt class that does something similair but for int. Is it possible to reuse it/be consistent with it?

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
```diff
@@ -72,45 +83,100 @@ module ParticleTracksModule
...
+    procedure :: append => memory_append
+    procedure :: flush => memory_flush
+    procedure :: discard => memory_discard
+    procedure :: destroy => memory_destroy
+  end type MemoryBufferType
+
+  !> @brief Scratch-file particle event buffer. Records are written to
+  !! an unformatted sequential scratch file that is rewound on discard
+  !! and read back sequentially on flush.
+  type, extends(ParticleTrackEventBufferType) :: ScratchFileBufferType
```
Manangka: I would suggest moving the MemoryBufferType and ScratchFileBufferType to separate files. If you would also create a factory then this file can be completely decoupled from these concrete classes and only depends on the interface `ParticleTrackEventBufferType`.  
Some of the benefits of decoupling the code:
- This file becomes smaller and easier to read
- Would you ever want to add or remove  a concrete ParticleTrackEventBufferType then this file doesn't have to change

I did something like this recently for the isotherms. You can have a look at `IsothermFactoryModule`

Manangka:
Hi @wpbonelli, I had a look at your PR and left some of my thoughts. Feel free to ignore any of my suggestions

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
wpbonelli: no not yet, it will be used in #2753. figured the buffering deserved a separate PR.

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
wpbonelli: Right, the dynamic expansion is vector-like. I guess it could be a type with vectors, not an array of derived types (SoA instead of AoS) but it has reals and strings so we'd need `STLVecReal` etc right? that may be the right way ultimately. probably something to do separately though.

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
wpbonelli: good idea.. will do! thanks.

src/Model/ParticleTracking/prt.f90
wpbonelli: maybe so. we haven't used `use_...` for other variables that we probably should have, by the same convention, e.g. `prt-prp.options.local_z` -> `use_local_z`

wpbonelli:
thanks @Manangka good thoughts all.

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
Manangka: I agree. That sounds out of scope for this story. When we would pick that up we might also look into templating in FORTRAN

src/Model/ParticleTracking/prt.f90
Manangka: I had a look at our Fortran style guide. There isn't any convention about how to name variables. We can create a separate issue to sort that out (if we want to formalize that)

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
```diff
@@ -120,21 +111,32 @@ subroutine init_file(this, iun, csv, iprp)
...
+    end if
+    call this%buffer%init()
+  end subroutine init_buffer
+
   !> @brief Destroy the particle track manager.
   subroutine destroy(this)
     class(ParticleTracksType) :: this
+    call this%buffer%destroy()
+    if (allocated(this%buffer)) deallocate (this%buffer)
     if (allocated(this%files)) deallocate (this%files)
```
Manangka: I believe these 2 deallocations are not needed. As far as understand class variables that are declared as allocatable get deallocated automatically when the class gets deallocated. If it they would be declared as pointers then you do have to manually deallocate them.
You can verify it using valgrind

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
Manangka: That cleans up nicely!

src/Solution/ParticleTracker/Particle/ParticleTracks.f90
wpbonelli: Yes I think you're right.

doc/mf6io/mf6ivar/dfn/prt-oc.dfn
```diff
@@ -292,6 +292,14 @@ mf6internal dev_dump_evtrace
 longname print particle tracking events
 description print a verbose particle tracking event trace to standard output
 
+block options
+name scratch_buffer
+type keyword
+reader urword
+optional true
+longname buffer track events in scratch file instead of memory
+description keyword to stage track events in a scratch file instead of an in-memory buffer, which prevents out-of-memory errors for simulations with very many particles or events
```
aprovost-usgs, line 301: The description in develop.toml has some additional information that I think would be of help to users here, too.  Consider something like this:

description keyword to stage track events in a temporary (scratch) file instead of an in-memory buffer. Particle tracking results calculated during a time step are buffered, and output is deferred, until successful completion of the time step. By default, results are buffered in memory. However, for simulations in which a very large number of track events is generated during a time step, memory limits can be exceeded. In such cases, the SCRATCH_BUFFER option can be invoked to buffer results in a scratch file until successful completion of the time step. Buffering in a scratch file is typically slower than buffering in memory.

wpbonelli: updated the description as suggested.

PR #2795, opened 2026-05-05
build: add pixi gcc-build env

jdhughes-dev:
Add gcc-build pixi environment for building extended version of MODFLOW 6 on macOS and Linux operating systems.


jdhughes-dev:
@wpbonelli I put the script for updating the netcdf pc files in the distribution directory since this was the only directory with non-autotest scripts. I add a trap for the case where the netcdf libraries are not installed but maybe the netcdf update script should go somewhere else.

pixi.toml
```diff
@@ -128,6 +140,37 @@ make-release-notes = { cmd = "python mk_releasenotes.py", cwd = "doc/ReleaseNote
...
+    "test-drive-test",
+    "test-drive-install",
+] }
+gcc-build-update = { depends-on = ["update-pc-files", "test-drive-rebuild"] }
+
+[feature.gcc-build.dependencies]
+git = "*"
+meson = "*"
+ninja = "*"
+gcc = "*"
```
Manangka, line 167: Do we want to pin this to gcc v13.2 to be in line with the compiler being used in the CI-pipeline?
It would also also enable you to include test-drive

jdhughes-dev: @Manangka gcc-15 is the only version available for macos from conda-forge

jdhughes-dev: @Manangka test-drive is also rebuilt from source if the gcc-build environment is used. There is a commit in the test-drive repo that allows it to work with gcc-15. So when a new version of test-drive is available on conda-forge the gcc-build environment tasks can be simplified.

wpbonelli (approved):

pixi.toml
jdhughes-dev: @Manangka set linux-64 to use gcc v13.2 macos is still 15

wpbonelli (approved):

PR #2804, opened 2026-05-16
feat(ims): add amg preconditioner

jdhughes-dev:
Add a algebraic multi-grid preconditioner to IMS.


autotest/TestImsLinearMethods.f90
```diff
@@ -0,0 +1,748 @@
+!> @brief Unit tests for IMS linear solver preconditioners
```
mjr-deltares, line 1: This is great to have in our test coverage!

autotest/test_gwf_ims_amg.py
```diff
@@ -0,0 +1,595 @@
...
+    "ims_amg_cg",
+    "ims_amg_bicgstab",
+    "ims_amg_cg_adp",  # adaptive omega (RELAXATION_FACTOR 0); max 16 chars
+    "ims_amg_nsmooth1",  # NUMBER_SMOOTHING_ITERATIONS 1
+    "ims_amg_nsmooth4",  # NUMBER_SMOOTHING_ITERATIONS 4, CG
+    "ims_amg_nwt_ns1",  # Newton-Raphson + BiCGSTAB + AMG fixed omega + nsmooth=1
+    "ims_amg_nwt_fix",  # Newton-Raphson + BiCGSTAB + AMG fixed omega
+    "ims_amg_s25",  # CG + AMG + adaptive omega + sthresh=0.25
+    "ims_amg_s50",  # CG + AMG + adaptive omega + sthresh=0.50
+    "ims_amg_nwt_s25",  # Newton-Raphson + BiCGSTAB + AMG fixed omega + sthresh=0.25
```
mjr-deltares, line 54: Looking at the convergence behavior for this and some of the other cases, it looks like a single node is still delaying rapid convergence. Do we understand this? Maybe we can look at it in a bit more detail offline. The results are really impressive but it's almost as if we can even do better than this...

autotest/test_gwt_ims_amg.py
```diff
@@ -0,0 +1,352 @@
+"""
```
mjr-deltares, line 1: This collection of test cases is really great. I am curious to see how this will scale with parallel computations. We should get this out into develop and test on Snellius.

mjr-deltares (approved):

mjr-deltares (approved):

autotest/test_gwf_ims_amg.py
jdhughes-dev: Claude's analysis:

What's happening (base ims_amg_cg, from the inner-iteration listing): the first ~10 CG iterations converge the solution (max change → ~1e-10, ~10×/iter), then ~90 more iterations grind down the residual at a single interior cell hovering near the tight RCLOSE=1e-9. Total = 101 inner iterations for a 600-node problem.

The node: a high-contrast checkerboard cell — k_layer=[10,0.1,1] × checkerboard {10, 0.1} puts a K≈1 cell directly among K≈100 neighbors (100× contrast). It's weakly coupled → a near-null-space mode the pairwise aggregation + default smoother reduces slowly.

Which lever helps (inner iters, CG cases — measured):
- SMOOTHER_TYPE ILU0_ALL: 101 → 62 (biggest win)
- AMG_CYCLE_TYPE W: 101 → 68
- NUMBER_SMOOTHING_ITERATIONS 4: 101 → 83
- STRENGTH_THRESHOLD 0.25→0.50: ~101 (no effect) — so it's not an aggregation-strength issue
- GS smoother: 144 (worse)

Plan for the offline discussion with mjr:
1. Root cause is a localized coarse-grid mode at high-contrast cells; the default (ILU0-finest + GS-coarse) smoother handles it poorly at coarse levels.
2. Cheapest mitigations already available: recommend ILU0_ALL (or W-cycle / more sweeps) for high-contrast models.
3. Deeper fix (the "do even better"): improve aggregation so weakly-coupled/near-singleton cells aren't stranded (e.g., a merge pass), and/or a couple of extra finest-level sweeps targeting high-residual nodes.
4. Caveat: RCLOSE=1e-9 in the test exaggerates the tail; the practical cost depends on the target tolerance.

jdhughes-dev:
@mjr-deltares the PR now includes addition of the AMG solver as a PETSC option and an associated integration test that sweeps through AMG options for the parallel implementation. It looks like a reasonable implementation...but you are the final arbiter of all things parallel.

mjr-deltares:
looks good @jdhughes-dev , I will have a more careful look after the weekend and let you know.

jdhughes-dev:
### Design note: keep the preconditioner setup re-entrant (enables future per-stress-period switching)

Not a change request for this PR — flagging a small structural affordance now, while this code is being written, so a planned follow-up stays additive rather than reopening the preconditioner setup later.

**Future objective:** allow different linear/nonlinear settings per stress period, selected by steady-state vs transient — e.g. AMG for steady-state and an ILU type for transient (cheaper), plus different inner/outer iteration limits per period.

This PR already gets us most of the way there:

- preconditioner type is a settings enum (`IPC_ILU0` … `IPC_AMG`), so selection is a settings value, not hardcoded logic;
- on the PETSc path, `PcShellCtxType` already has a `pctx_create` / `pctx_destroy` pair that allocates/frees the type-specific storage (including the AMG hierarchy), and `pcshell_setup` already rebuilds the numeric factorization per solve.

The only gap is that the preconditioner **type + its type-specific storage are established once** (serial: at `*_ar`; PETSc: in `pctx_create`), with no path to change them mid-run.

**Ask:** make the preconditioner setup tear-down-and-rebuildable mid-run, driven by the settings enum, at both call sites:

- **serial `ImsLinearDataType`** — a `destroy()` + `setup(settings)` pair, i.e. `reconfigure(settings)`;
- **PETSc `PcShellCtxType`** — a `reconfigure(matrix, settings)` that composes the existing `pctx_destroy` / `pctx_create` when the re-derived `ipc` changes, reconfiguring the context **in place** so the `PCShellSetContext` wiring stays valid.

A couple of things that keep this cheap:

- the Krylov/KSP object doesn't need to change when only the PC changes, and the symbolic structure (IA/JA) is period-invariant (steady↔transient only changes the storage diagonal) — so a rebuild is just type-specific workspace + numeric factorization;
- the "AMG in a PETSc solution needs a fixed linear cycle" constraint you already enforce should simply be preserved on switch.

Per-period switching itself would land in a separate PR; the request here is only that the setup **can** be re-run so that follow-up doesn't have to rework this code. Happy to open that follow-up.

PR #2809, opened 2026-05-21
feat: tolerate unknown options

wpbonelli:
Add a keyword option `STRICT` to the simulation, defaulting true for consistency with previous behavior. In strict mode, unrecognized variables will stop the simulation with an error, otherwise the simulation will continue with a warning.

Alternative and/or complimentary solution to https://github.com/modflowpy/flopy/pull/2712.

___


src/Utilities/Idm/mf6blockfile/LoadMf6File.f90
```diff
@@ -315,12 +316,29 @@ recursive subroutine parse_block(this, iblk, recursive_call)
           if (endOfBlock) exit
           ! process line as tag(s)
           call this%parser%GetStringCaps(tag)
-          idt => get_param_definition_type( &
-                 this%mf6_input%param_dfns, &
-                 this%mf6_input%component_type, &
-                 this%mf6_input%subcomponent_type, &
-                 this%mf6_input%block_dfns(iblk)%blockname, &
-                 tag, this%filename)
+          if (itolerate_unknown == 1) then
```
mjreno: This isn't specific to the OPTIONS block you would also need to check the block here.

wpbonelli:
I'm undecided whether strict should be opt-in or on by default. Opt-in seems like nicer UX for anything driving multiple versions of MF6, like GUIs and flopy, with some added risk of not noticing if options go unused, because the simulation ran and the warning is buried/offscreen.

src/Utilities/Idm/mf6blockfile/LoadMf6File.f90
wpbonelli: I'm not sure if this should apply only to options or to any variable. new options are added much more often than new data variables so would probably cover most cases, but seems more consistent to apply to all variables. any thoughts?

PR #2813, opened 2026-05-25
fix: improve error message when dis package missing

wpbonelli:
<details>
<summary>First draft, superseded</summary>

If `mshape` is not associated when package data loads, it likely means the model doesn't have a discretization package. Calling `product(mshape)` is then undefined behavior,  on Windows / Intel Fortran as reported in #2806 it continues into an invalid state until the parser breaks, on Mac / gfortran I found it crashes immediately.

Stop with an informative error suggesting a discretization may be missing if `mshape` is not associated at package data load time. This seems a bit ad hoc, but I'm not sure where to trap for a missing dis otherwise. Maybe you have a better idea @mjreno ?
</details>

Stop with an informative error if grid discretization is missing after loading the model input file's `packages` block.

___

### DIFF
--- a/doc/ReleaseNotes/develop.toml
+++ b/doc/ReleaseNotes/develop.toml
@@ -63,3 +63,8 @@ description = "Avoids an unhandled and unnecessary exception in cases where a GW
 section = "fixes"
 subsection = "netcdf"
 description = "Fix an indexing issue in structured NetCDF export where READASARRAYS packages (RCHA, EVTA) with time-varying per-period arrays would crash with a NetCDF edge-exceeds-bound error when writing any stress period after the first."
+
+[[items]]
+section = "fixes"
+subsection = "basic"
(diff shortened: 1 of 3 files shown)


mjreno (commented):
I think this is a reasonable place to do this but I'm not sure that it would be complete- there are other uses of `mshape` that look like they might not be handled.  A broad check might be a better fit where the nam file `packages` block is processed.  In that context the model type is known, including supported packages, and `dis` could be verified as present when it should be.  That code is in `ModelPackageInputs.f90`.  We can look at that together if you want.

wpbonelli:
@mjreno how's it look now?

mjreno:
> @mjreno how's it look now?

Looks good, this does assume some DIS package is always model associated which seems okay.

mjreno (approved):

PR #2818, opened 2026-05-28
feat(sfr): SFR kinematic wave approximation Cr output and doc

jdhughes-dev:
Add a chapter on the SFR kinematic wave approximation (STORAGE) option to the Supplementary Technical Info document. Also add a table at the end of the Supplementary Technical Info document chapter that has a summary of the minimum, maximum, and mean Courant number to give users a sense of the validity of the SFR kinematic-wave results.


jdhughes-dev:
@emorway-usgs just rebased with develop...so this is ready to go once approved.

jdhughes-dev:
@emorway-usgs and @aprovost-usgs 

**TVD width-variation and dQ-vs-dA analysis**
Checked whether the TVD kinematic-wave scheme oscillates when reach width changes downstream, and whether an area-based (dA) limiter would beat the flow-based (dQ) limiter. Routed a sharp inflow pulse down a straight multi-reach channel with constant width, gradual widening and narrowing, and abrupt width steps (5↔25 m), at Courant numbers from ~0.2 to ~1, comparing against the standard kinematic scheme.

No oscillation for any width configuration when Cr ≤ 1. The van Leer limiter is TVD-bounded: at a width change the smoothness ratio just drops the correction to first-order upwind, which is monotone. Profiles stay clean — deep in narrow reaches, shallow in wide reaches, a single step at the width change.

The only oscillation is at Cr > 1, where the explicit update is unstable regardless of width. The direct volume balance still closes the budget to 0.00%, so it is easy to miss. ATS_COURANT keeps Cr ≤ 1 in practice

The dA limiter is worse, not better. Prototyped it (smoothness ratio and anti-diffusion from cross-sectional area, area gradient scaled to a flux by celerity). At constant width the depth overshoots the steady value by ~7% (peak 1.82 m vs the exact 1.75 m), while the dQ scheme peaks right at the steady value. A correct area-based limiter needs the downstream reach area for the outflow-face gradient, which the per-reach loop does not carry — a larger change for no demonstrated benefit. Kept the dQ limiter.

Two small hardening changes came out of this:

- Clamp the TVD outflow at zero before the volume balance, so the stored volume and the reported outflow use the same flux and the budget still closes exactly if the correction ever goes negative.
- Strengthened the "ATS_COURANT without ATS" warning to say the time step is not adapted and the solution may oscillate at Cr > 1 while the budget still closes.

Let me know if you have additional suggestions.

emorway-usgs:
> @emorway-usgs and @aprovost-usgs
> 
> **TVD width-variation and dQ-vs-dA analysis** Checked whether the TVD kinematic-wave scheme oscillates when reach width changes downstream, and whether an area-based (dA) limiter would beat the flow-based (dQ) limiter. Routed a sharp inflow pulse down a straight multi-reach channel with constant width, gradual widening and narrowing, and abrupt width steps (5↔25 m), at Courant numbers from ~0.2 to ~1, comparing against the standard kinematic scheme.

I also tried several combinations of:

- abrupt constriction ratios (1/2, 1/10, 1/100 as wide)
- flow rates (10, 100, 1000) 
- stream reach lengths (1, 10, 100)  

I ran everything in steady-state.  None of the testing turned up any oscillations in the calculated stream depths

aprovost-usgs:
> > @emorway-usgs and @aprovost-usgs
> > **TVD width-variation and dQ-vs-dA analysis** Checked whether the TVD kinematic-wave scheme oscillates when reach width changes downstream, and whether an area-based (dA) limiter would beat the flow-based (dQ) limiter. Routed a sharp inflow pulse down a straight multi-reach channel with constant width, gradual widening and narrowing, and abrupt width steps (5↔25 m), at Courant numbers from ~0.2 to ~1, comparing against the standard kinematic scheme.
> 
> I also tried several combinations of:
> 
> * abrupt constriction ratios (1/2, 1/10, 1/100 as wide)
> * flow rates (10, 100, 1000)
> * stream reach lengths (1, 10, 100)
> 
> I ran everything in steady-state. None of the testing turned up any oscillations in the calculated stream depths

@jdhughes-dev and @emorway-usgs - Thanks for looking into the alternative of using an area-based (dA) limiter.  Given that that the tests you ran collectively showed no unexpected issues with the current flow-based (dQ) limiter, and in some cases inferior performance with the dA limiter, I agree it makes sense to stick with the dQ limiter.

doc/SuppTechInfo/sfr-kinematic-wave.tex
```diff
@@ -0,0 +1,422 @@
...
+
+\begin{equation}
+	\label{eqn:sfr-continuity}
+	\frac{\partial Q}{\partial x} + \frac{\partial A}{\partial t} = q_L ,
+\end{equation}
+
+\noindent where $Q$ is the volumetric discharge ($L^3T^{-1}$); $A$ is the cross-sectional area of flow ($L^2$); $x$ is the longitudinal distance along the channel ($L$); $t$ is time ($T$); and $q_L$ is the lateral inflow rate per unit channel length ($L^2T^{-1}$). The kinematic-wave approximation is applicable when the Froude number is less than about 2 and the channel slope is sufficiently steep relative to the backwater effect \citep{McArthur1993, WoolhiserLiggett1967, PonceLiSimons1978, Ponce1991}.
```
emorway-usgs: There seems to be a small discrepancy where is states "...the kinematic-wave approximation is applicable when the Froude number is less than about 2..." and the 4th bullet under the "Applicability Conditions" section when it suggests users should not activate the kinematic-wave approach for a Froude number greater than 1.  Should this part be altered to instead suggest that the "kinematic-wave approximation is applicable when the Froude number is less than about 1"?

doc/SuppTechInfo/sfr-kinematic-wave.tex
```diff
@@ -0,0 +1,422 @@
...
+
+\subsubsection{Standard Kinematic-Wave Scheme}
+
+The standard kinematic-wave scheme solves for the unknown downstream discharge $Q_D^n$ at each reach by finding the root of the finite-difference residual using Newton-Raphson iteration. The lateral source term $q_L$ in the residual aggregates all per-reach inflow and loss contributions, and the temporal weighting parameter $\theta$ controls the balance between numerical stability and temporal accuracy. Once the Newton-Raphson iteration has converged, the volumetric storage change is computed explicitly from the converged cross-sectional areas to close the reach water budget.
+
+The net lateral source per unit length, $q_L$ ($L^2T^{-1}$), combines all per-reach inflow and loss terms
+
+\begin{equation}
+	\label{eqn:sfr-qlat}
+	q_L = \frac{q_r + q_{ro} - q_e}{\Delta x} - \frac{q_{gwf}}{\Delta x} ,
```
emorway-usgs, line 125: What about contributions from the MVR package?

doc/SuppTechInfo/sfr-kinematic-wave.tex
```diff
@@ -0,0 +1,422 @@
...
+
+The temporal weighting parameter $\theta$ defaults to 1.0 (fully implicit) and may be adjusted using the developer option \texttt{DEV\_STORAGE\_WEIGHT}. Values of $\theta$ between 0.5 and 1.0 are accepted; values less than 0.5 are not permitted because they can produce unstable solutions. A value of $\theta = 1.0$ provides maximum numerical stability at the cost of first-order temporal accuracy and is strongly recommended for all practical applications. When $\theta = 0.5$, equation~\ref{eqn:sfr-mf6-residual} becomes the Crank-Nicolson scheme \citep{CrankNicolson1947}, which is second-order accurate in time but may exhibit oscillations when $C_r < 1$ and the spatial discretization is coarse. The Courant number is evaluated at each reach and time step using the celerity estimated from the effective upstream inflow---the upstream discharge $Q_U^n$ adjusted for lateral contributions and groundwater exchange (eq.~\ref{eqn:sfr-courant}).
```
emorway-usgs: At release, will the parameter `DEV_STORAGE_WEIGHT` be truncated to just `STORAGE_WEIGHT` automagically?  I was thinking the `DEV_` prefix was reserved for development only and not used in an official version of the code?

doc/SuppTechInfo/sfr-kinematic-wave.tex
```diff
@@ -0,0 +1,422 @@
...
+\noindent The updated discharge is limited to a minimum of $10^{-30}$~m$^3$\,s$^{-1}$ to keep it positive. Iteration continues until
+
+\begin{equation}
+	\label{eqn:sfr-mf6-converge}
+	\left| Q_D^{n,k+1} - Q_D^{n,k} \right| < 2\delta
+	\quad \text{and} \quad
+	\left| f\!\left(Q_D^{n,k+1}\right) \right| < 2\delta ,
+\end{equation}
+
+\noindent or until the maximum of 100 Newton-Raphson iterations is reached. The perturbation $\delta$ is set to $0.999 \times \texttt{MAXIMUM\_DEPTH\_CHANGE}$ and is used for the numerical celerity estimate (eq.~\ref{eqn:sfr-celerity-num}), the numerical derivative (eq.~\ref{eqn:sfr-mf6-deriv}), and the Newton-Raphson convergence test. The Picard depth-change tolerance equals \texttt{MAXIMUM\_DEPTH\_CHANGE} directly. Both are governed by the same input parameter.
```
emorway-usgs: At the end of the sentence ending with "..., and the Newton-Raphson convergence test," would it make sense to include a reference to equation 16-16 for consistency with the other parts of the sentence that reference specific equations?  I think eqn 16-16 is what's in view here?

doc/SuppTechInfo/sfr-kinematic-wave.tex
```diff
@@ -0,0 +1,422 @@
...
+
+A dry reach that gains water from the aquifer---where the aquifer head lies above the streambed top---but receives little or no inflow from upstream cannot rewet with the standard iterative solution: at zero depth both the wetted cross-sectional area and the streambed conductance vanish, so a dry reach stays dry and receives no groundwater discharge. For such a reach the depth is instead found by bisection between a small wet depth and the aquifer depth above the streambed top, which lets the reach rewet and receive groundwater discharge for any time-step size. Reaches that receive meaningful inflow from upstream are solved with the standard iteration approach.
```
emorway-usgs: Instead of "For such a reach, ..." maybe for clarity it can be stated like this: "To overcome this issue, the depth in dry reaches that receive groundwater discharge is instead found by bisection..."

doc/SuppTechInfo/sfr-kinematic-wave.tex
```diff
@@ -0,0 +1,422 @@
...
+
+Figure~\ref{fig:sfr-kinwave-tvd} compares the standard and TVD schemes against the exact solution at $C_r = 0.5$ (panel~\textit{A}) and $C_r = 0.75$ (panel~\textit{B}). At both Courant numbers, the standard scheme diffuses the wave, reducing the peak amplitude and spreading the hydrograph. The TVD scheme, with the van Leer anti-diffusion correction active at the downstream reach, closely tracks the exact solution: the peak amplitude and wave shape are well-recovered even at $C_r = 0.5$, where the standard scheme produces marked attenuation. This demonstrates that the TVD option can provide accurate flood-wave routing at time steps significantly shorter than the optimal $C_r \approx 1$ value required by the standard scheme.
```
emorway-usgs: Something seems off in this paragraph.  If figure 16-3 is what is in view, as the start of the paragraph would suggest, the claim is that the standard finite difference approach diffuses the wave thereby reducing the peak amplitude and spreading the flow.  The next sentence ("The TVD scheme, with the van Leer...at Cr=0.5,...") then states that the TVD scheme "closely tracks the exact solution."  However, in panel "A" both the finite difference and TVD approaches appear to overlie one another?  The FD approach would appear to be doing just as well as the TVD approach for Cr=0.5 (panel A)?

doc/SuppTechInfo/sfr-kinematic-wave.tex
```diff
@@ -0,0 +1,422 @@
...
+\end{figure}
+
+\subsection{Practical Guidance}
+
+\subsubsection{Choosing a Suitable Time Step}
+\label{sec:sfr-time-step}
+
+The accuracy of the kinematic-wave scheme depends on the Courant number $C_r = c\,\Delta t / \Delta x$ (eq.~\ref{eqn:sfr-courant}), and figure~\ref{fig:sfr-kinwave} demonstrates that $C_r$ near unity provides the best balance between accuracy and computational efficiency. A practical starting point is therefore to target $C_r \approx 1$ at or near the expected base flow.
+
+For a rectangular channel, the kinematic-wave celerity can be estimated from Manning's equation. When the channel is hydraulically wide (width $B$ much larger than flow depth $d$), the hydraulic radius $R \approx d$, and the celerity simplifies to approximately five thirds of the mean flow velocity $\bar{V} = Q / A$
```
emorway-usgs, line 349: When the channel is hydraulically wide, wouldn't the hydraulic radius `R` be better approximated by B (width) as opposed to `d` (depth), as opposed to what is currently stated?

doc/SuppTechInfo/sfr-kinematic-wave.tex
```diff
@@ -0,0 +1,422 @@
...
+
+\noindent where $d_0$ is the base-flow depth and $F_0 = \bar{V}_0 / \sqrt{g\,d_0}$ is the Froude number at base flow. Values of $K \gtrsim 10$ indicate that the kinematic wave is a good approximation; values of $K < 5$ suggest that momentum storage, pressure gradients, or backwater effects are important and that the kinematic wave may produce significant errors \citep{Ponce1991}. The kinematic-wave option should not be used in the following situations:
+
+\begin{itemize}
+
+	\item \textit{Very mild slopes} ($S_0 \lesssim 10^{-4}$). Longitudinal pressure gradients and downstream backwater effects become important relative to friction, and the diffusive wave or full dynamic wave is more appropriate.
+
+	\item \textit{Regulated or ponded reaches}. Reaches immediately downstream of dams, weirs, or tidal boundaries are subject to imposed water-surface profiles that invalidate the kinematic assumption.
+
+	\item \textit{Reaches with strong stream-aquifer exchange}. Where large and rapidly varying groundwater seepage substantially alters the flow depth within a reach, the simple celerity estimate may be unreliable and convergence may be slow.
```
emorway-usgs: What about a large runoff event?  For example, if MVR is connecting UZF or DRN to SFR and there is a precip event contributing to 'large and rapidly' varying flows in the channel, does that compromise the usage of the `STORAGE` option?

doc/ReleaseNotes/develop.toml
```diff
@@ -24,6 +25,16 @@ section = "features"
...
+
+[[items]]
+section = "documentation"
+subsection = "advanced"
+description = "A new chapter describing the SFR Package kinematic-wave routing options has been added to the Supplemental Technical Information document. The chapter covers the kinematic wave approximation, the four-point implicit finite-difference discretization, the standard (STORAGE keyword) and TVD (ATS\\_COURANT keyword) routing schemes, example comparisons against exact kinematic-wave solutions for a two-reach MODFLOW 6 model at Courant numbers ranging from 0.5 to 2.0, and practical guidance for selecting a suitable time step."
```
emorway-usgs: Suggest changing "The chapter covers..." to "The chapter describes..."

emorway-usgs (approved):
Great job on the write up in the Supp Tech Info doc.  It read well and I only note a couple of minor discrepancies for your consideration (provided as comments in the PR).  My overall sense is that the `STORAGE` option seems limited to a narrow set of use cases (don't use in mildly slopes area, nor stream networks with lots of confluences, nor managed systems), but for other cases where storage changes in the channel are important, offers significant benefit.

doc/SuppTechInfo/sfr-kinematic-wave.tex
jdhughes-dev: Good catch that these two statements sit oddly together, though I think they are describing different criteria rather than contradicting each other.

The "less than about 2" figure relates to roll-wave (Vedernikov) instability, the point at which the kinematic wave ceases to approximate the dynamic wave. For a wide channel that threshold is near Fr = 2 with Chezy friction and closer to Fr = 1.5 with Manning. The `F_0 >= 1` bullet is a narrower, practical caution about this implementation: normal depth and the simple celerity estimate both degrade as flow approaches critical.

Rather than change 2 to 1, I would make the distinction explicit and say what each threshold marks, since both numbers are defensible for what they describe. I have not gone back to the four cited sources to confirm which criterion each supports, so let me know if you would rather I do that before wording it.

doc/SuppTechInfo/sfr-kinematic-wave.tex
jdhughes-dev: I think the text is right as written here. The hydraulic radius is the flow area divided by the wetted perimeter, so for a rectangular channel

R = A / P = Bd / (B + 2d)

As B/d becomes large the 2d term becomes negligible and R tends to Bd/B = d. So a hydraulically wide channel gives R ~ d, the flow depth, not the width. R has to carry dimensions of length in Manning's equation, and it is the depth that reduces the conveyance as the channel gets shallow.

doc/SuppTechInfo/sfr-kinematic-wave.tex
jdhughes-dev: That is a fair gap. The bullet as written covers groundwater seepage, but a large runoff event routed through MVR from UZF or DRN is a surface inflow and is not addressed.

The mechanism is different: MVR water is added to the reach inflow rather than distributed along the reach, so it raises the discharge and therefore the celerity, which raises the Courant number. It does not invalidate the STORAGE option, but a large and rapid inflow can push the Courant number above the range where the standard scheme is accurate, and ATS_COURANT would be the better choice in that case.

I would extend the bullet to cover rapidly varying surface inflow alongside groundwater seepage. Worth a test case if you think it needs demonstrating rather than stating.

jdhughes-dev:
@emorway-usgs thanks for the review. The comments on the equation reference, the wording of the bisection paragraph, the lateral inflow description, the figure 16-3 paragraph, and the release note have been addressed, and I have replied to the remaining three inline.

The SuppTechInfo document is only built in release mode, so there is no compiled chapter in the docs CI artifacts. Let me know if a compiled PDF of the chapter would help you finish the review and I will build one and attach it.

doc/SuppTechInfo/sfr-kinematic-wave.tex
emorway-usgs: Thanks for clarifying, I had the wrong equn in my mind when I was thinking of the hydraulic radius.

doc/SuppTechInfo/sfr-kinematic-wave.tex
emorway-usgs: I think it is fine if you word it yourself, seems like you have a good handle on the nuances.  No need to go back to the original source material.

doc/ReleaseNotes/develop.toml
emorway-usgs: Resolved with 0699825.

doc/SuppTechInfo/sfr-kinematic-wave.tex
emorway-usgs: Resolved with 0699825

doc/SuppTechInfo/sfr-kinematic-wave.tex
emorway-usgs: Resolved with 0699825

doc/SuppTechInfo/sfr-kinematic-wave.tex
emorway-usgs: Resolved with 0699825

doc/SuppTechInfo/sfr-kinematic-wave.tex
jdhughes-dev: Worded in bfb117b. The introduction now says that a Froude number near 2 is the onset of roll waves, where the kinematic wave stops approximating the dynamic wave, and the applicability bullet says that the caution near 1 is a property of this implementation, whose celerity estimate is derived from Manning's equation and becomes less reliable as the flow approaches critical.

I also dropped the claim in that bullet that the kinematic-wave approximation assumes subcritical flow. It does not, and the limitation there is the celerity estimate rather than the approximation itself.

doc/SuppTechInfo/sfr-kinematic-wave.tex
jdhughes-dev: The prefix is not truncated at release. There are sixteen DEV_ options in the definition files, several of them long lived, so the prefix marks an option as unsupported rather than temporary.

What prompted the question was a mistake in the definition file: the option is named dev_storage_weight but its description referred to STORAGE_WEIGHT twice, which reads as though the prefix is dropped. Fixed in 1866502.

The chapter no longer names the keyword at all. Input syntax belongs in the input and output guide, so the chapter now discusses the temporal weighting parameter only as a property of the discretization and keeps the caution against reducing it below 1.

doc/SuppTechInfo/sfr-kinematic-wave.tex
jdhughes-dev: Added in 0699825. The sentence now references the convergence criterion alongside the celerity estimate and the numerical derivative.

doc/SuppTechInfo/sfr-kinematic-wave.tex
jdhughes-dev: Tested this in 2d6dbf9. A twenty-fold inflow surge is routed through a five-reach channel, which raises the Courant number from 0.7 at base flow to about 2. The answer is that it does compromise the STORAGE option: the per-step budget discrepancy rises from below 0.001 percent at base flow to 42 percent, and takes several time steps to decay. The TVD scheme closes the budget through the same surge.

The test uses a step change in specified inflow rather than a mover. Water provided by MVR is added to the reach inflow rather than distributed along the reach, so it drives the routing the same way, and the specified inflow is easier to control.

The applicability bullet now covers rapidly varying surface inflow alongside groundwater seepage and recommends the TVD scheme where such events are expected.

doc/SuppTechInfo/sfr-kinematic-wave.tex
jdhughes-dev: Following up on this one. You were right to ask. DEV_STORAGE_WEIGHT was not calling DevOpt, so unlike the other two development options in the package it was accepted in a release build. Gated in 2d6dbf9, and test_gwf_sfr_kinematic01 is now marked developmode so it is not run against a release build.

PR #2835, opened 2026-06-12
chore: update pixi envs, consolidate CI logic, fix macOS extended CI

wpbonelli:
macOS extended CI began failing due to release of `netcdf-fortran` 4.6.3 which was built for homebrew with a newer `gfortran` than 13. To fix this, switch Linux and macOS extended CI from a mix of dependency sources to the existing `gcc-extended-build` pixi env, which pulls a self-consistent toolchain from conda-forge. This allows removing the custom `test-extended` action that builds PETSc from source, and avoids homebrew version drift and `gfortran` incompatibilities.

Also:
* Include `default` feature in `gcc-extended-build` pixi env, and drop newly redundant dependencies from the latter
* Use `setup-extended` task in CI and docs
* Set `UCX_TLS=sm,self` on extended test steps


wpbonelli:
trying to do too many things at once here

PR #2861, opened 2026-06-24
chore: split release notes into separate files

wpbonelli:
Instead of keeping release notes in a single file `doc/ReleaseNotes/develop.toml`, store them as separate files in a directory `doc/ReleaseNotes/items/`, to be merged when generating LaTeX for the docs. This avoids the possibility of merge conflicts due to PRs all editing a single file.

Instead of adding lines to the same `develop.toml`, the new way to add a release note would be to create a TOML file in `items/`, named anything (ideally something descriptive), with the same three attributes as before, `section`, `subsection`, and `description`, e.g. `fix-bmi-get-grid-model-type.toml`

```toml
section = "fixes"
subsection = "basic"
description = "The BMI interface's get\\_grid\\_model\\_type function worked only for numerical model types such as GWF, GWT, and GWE, but would crash for PRT models. Fix the function's implementation to work with all model types."
```


mjr-deltares (approved):
I had no real problems merging, but no objections against this method either.

jdhughes-dev (approved):

aprovost-usgs (approved):
Seems like a good idea if merge conflicts have been a problem in the past, or are likely to be in the future

emorway-usgs (approved):
I haven't had too much difficulty dealing with merge conflicts when they arise, but this is likely a nice improvement for those blocks of time right before a release when merge conflicts are not uncommon.  One minor question that comes to mind is will this approach increase some of the churn right after a release?  Will all the added toml files need to be cleaned out and if so, will it be automated or does it require manual attention?

wpbonelli:
> Will this approach increase some of the churn right after a release? Will all the added toml files need to be cleaned out and if so, will it be automated or does it require manual attention?

After #2977 it will automatically remove the files in the post-release PR, which we do as a manual step right now.

Manangka (approved):

PR #2880, opened 2026-07-06
refactor(ims): extract nonlinear outer-loop kernels into ImsNonlinearBase

jdhughes-dev:
## Description

First refactor step toward exposing the IMS solver through the MODFLOW API: extract the stateless bodies of the IMS **nonlinear (outer-iteration)** routines out of `NumericalSolutionType` into free subroutines in a new `ImsNonlinearBaseModule`, mirroring the existing `ImsLinearBase` pattern. `NumericalSolutionType` methods become thin delegating wrappers, so **behavior is unchanged** and the reusable kernels can be unit tested directly and reused outside the solution object.

This is a behavior-preserving, independently-mergeable move — no numerical results should change.

### Extracted kernels (verbatim transforms; `this%<member>` → explicit args)

| Free subroutine | Extracted from |
|---|---|
| `ims_nl_underrelax` | `sln_underrelax` (SIMPLE/COOLEY/DBD) |
| `ims_nl_maxval` | `sln_maxval` |
| `ims_nl_calcdx` | `sln_calcdx` |
| `ims_nl_dxmax` | `sln_get_dxmax` |
| `ims_nl_backtrack_flag` | `get_backtracking_flag` |
| `ims_nl_apply_backtrack` | `apply_backtracking` |
| `ims_nl_has_converged` | `sln_has_converged` |
| `ims_nl_nur_has_converged` | `sln_nur_has_converged` |
| `ims_nl_residual` | `sln_calc_residual` (r = A*x − b, zeroed for inactive) |
| `ims_nl_l2norm` | `sln_l2norm` (‖r‖₂) |

### Parallel seam

The overridable type-bound procedures keep their signatures, and `ParallelSolution`'s reduce-then-delegate overrides are preserved. The one override that did **not** delegate to the base — `par_has_converged`, which re-implemented the `dvclose` check — is now routed through the shared `ims_nl_has_converged` after its `MPI_Allreduce`, so the serial and parallel convergence criteria cannot diverge. Behavior is identical today (the reduced value is already non-negative, so the kernel's `abs()` is a no-op).

### Tests

Adds `autotest/TestImsNonlinearBase.f90` — 18 `test-drive` unit tests exercising the free kernels against hand-computed values (all three under-relaxation methods, inactive-cell skips, backtracking thresholds, convergence boundaries, signed dxmax, and a small diagonal system for residual / l2norm).

### Local verification

- Serial and extended (PETSc/MPI/NetCDF) builds link cleanly.
- All unit suites pass in both builds; `mf6` runs the outer loop to normal termination.
- `fprettify` clean on all changed/new files.

## Status

The matrix-coupled `sln_calc_residual` / `sln_l2norm` extraction (originally deferred) is now **included** in this PR. The full nonlinear outer-loop kernel is extracted. Still marked **draft** pending final review.

## Checklist

- [x] Updated meson files for new source files
- [ ] Updated `develop.toml` release notes — N/A (internal refactor, no user-facing change)

### DIFF
--- a/autotest/TestImsNonlinearBase.f90
+++ b/autotest/TestImsNonlinearBase.f90
@@ -1,0 +1,421 @@
+module TestImsNonlinearBase
+  use KindModule, only: I4B, DP
+  use ConstantsModule, only: DZERO, DONE, DEM20
+  use testdrive, only: check, error_type, new_unittest, test_failed, &
+                       to_string, unittest_type
+  use ImsNonlinearBaseModule, only: ims_nl_underrelax, ims_nl_maxval, &
+                                    ims_nl_calcdx, ims_nl_dxmax, &
+                                    ims_nl_backtrack_flag, &
+                                    ims_nl_apply_backtrack, &
+                                    ims_nl_has_converged, &
+                                    ims_nl_nur_has_converged
+  implicit none
+  private
+  public :: collect_imsnonlinearbase
+
+  real(DP), parameter :: tol = 1.0e-9_DP
+
+contains
+
+  subroutine collect_imsnonlinearbase(testsuite)
+    type(unittest_type), allocatable, intent(out) :: testsuite(:)
+    testsuite = [ &
+                new_unittest("underrelax_simple", test_underrelax_simple), &
+                new_unittest("underrelax_simple_inactive", &
+                             test_underrelax_simple_inactive), &
+                new_unittest("underrelax_cooley_kiter1", &
+                             test_underrelax_cooley_kiter1), &
+                new_unittest("underrelax_cooley_relax", &
+                             test_underrelax_cooley_relax), &
+                new_unittest("underrelax_dbd_kiter1", &
+                             test_underrelax_dbd_kiter1), &
+                new_unittest("underrelax_dbd_flipflop", &
+                             test_underrelax_dbd_flipflop), &
+                new_unittest("maxval_normalized", test_maxval_normalized), &
+                new_unittest("maxval_zero_denom", test_maxval_zero_denom), &
+                new_unittest("calcdx", test_calcdx), &
+                new_unittest("dxmax", test_dxmax), &
+                new_unittest("dxmax_signed", test_dxmax_signed), &
+                new_unittest("backtrack_flag_on", test_backtrack_flag_on), &
+                new_unittest("backtrack_flag_off", test_backtrack_flag_off), &
+                new_unittest("apply_backtrack", test_apply_backtrack), &
+                new_unittest("has_converged", test_has_converged), &
+                new_unittest("nur_has_converged", test_nur_has_converged) &
+                ]
+  end subroutine collect_imsnonlinearbase
+
+  !> @brief SIMPLE dampening (nonmeth == 1): x = xtemp + gamma * (x - xtemp)
+  !<
+  subroutine test_underrelax_simple(error)
+    type(error_type), allocatable, intent(out) :: error
+    integer(I4B), parameter :: neq = 2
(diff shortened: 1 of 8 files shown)


jdhughes-dev:
## Reviewer guide

**What this is:** the first refactor step toward exposing IMS through the MODFLOW API. It extracts the IMS **nonlinear (outer-iteration) kernel** out of `NumericalSolutionType` into stateless free subroutines in a new `ImsNonlinearBaseModule`, mirroring the existing `ImsLinearBase` pattern; the `NumericalSolutionType` methods become thin delegating wrappers. **Behavior-preserving — no numerical results should change.**

**Why:** untangling the reusable kernel from the model/exchange lists and file I/O makes the later API work additive instead of a hot-path rewrite, and lets the kernel be unit-tested directly.

### What changed
- New `src/Solution/LinearMethods/ImsNonlinearBase.f90` — 10 free subs: `ims_nl_underrelax` (SIMPLE/COOLEY/DBD), `ims_nl_maxval`, `ims_nl_calcdx`, `ims_nl_dxmax`, `ims_nl_backtrack_flag`, `ims_nl_apply_backtrack`, `ims_nl_has_converged`, `ims_nl_nur_has_converged`, `ims_nl_residual`, `ims_nl_l2norm`. Each is a verbatim move of the corresponding `sln_*` body with `this%<member>` replaced by explicit args.
- `NumericalSolution.f90` — those 10 routines are now thin wrappers delegating to the kernel.
- `ParallelSolution.f90` — one targeted change; see the seam note below.
- `autotest/TestImsNonlinearBase.f90` — 18 `testdrive` unit tests against hand-computed values (incl. a small diagonal `SparseMatrixType` system for `residual`/`l2norm`).

### Where to focus review
1. **The one subtle change — `par_has_converged` (`ParallelSolution.f90`).** Of the 9 parallel overrides, this is the only one that did **not** delegate to the base — it re-implemented the `dvclose` comparison inline after its `MPI_Allreduce`. Had we only rewired the base `sln_has_converged`, the parallel path would have silently kept the old inline logic. It now routes the globally-reduced value through the same shared `ims_nl_has_converged`. Behavior is identical today (the reduced value is already ≥ 0, so the kernel's `abs()` is a no-op), and the serial/parallel divergence trap is removed. The other 8 overrides delegate to their base method, so extracting the base body carries through automatically — no change needed there.
2. **Verbatim-ness.** Every wrapper should be a 1:1 move — worth a spot-check that no logic drifted, especially the DBD/COOLEY under-relaxation state handling and the residual inactive-cell zeroing.
3. **Scope of `residual`/`l2norm`.** These are matrix-coupled (they take the system matrix + x/rhs vectors); I folded them in here so the whole nonlinear kernel lands together. Flag if you'd prefer them split into a follow-up.

### Verification
- Serial + extended (PETSc/MPI/NetCDF) builds link clean.
- All 18 new unit tests + the full unit suite pass in both builds.
- `mf6` runs the outer loop to normal termination; `fprettify` clean.

### Not in this PR (intentional — later refactor steps)
- Descriptive nonlinear option enums + a `NonlinearSettingsType` (will mirror the `ImsLinearSettings` enum convention introduced in #2804, and land after that PR merges).
- An output opt-out flag.
- The API layer itself.

### CI note
An early red on `Pixi gcc-extended-build (ubuntu-22.04)` was an **intermittent SIGFPE flake** in the parallel `.mf6minsim` test — unrelated to this change: the serial run passes, the macOS variant passes, and the `residual`/`l2norm` path isn't even exercised by that model (no PTC, no `res_lim`). Cleared on rerun; CI is green.

jdhughes-dev:
Just brainstorming. Will try a different approach.

Issue #2917, opened 2026-08-11
GWT: gwfsatold omits the porosity when reconstructing the previous saturation


jdhughes-dev:
In MODFLOW 6 the saturation of a cell, `gwfsat`, is the fraction of the cell thickness that lies below the water table,

```
sat = (h - bot) / (top - bot)
```

smoothed near the top and the bottom of the cell and equal to one for a confined cell. It is a geometric quantity and carries no porosity; the volume of water in the mobile domain of a cell is `vcell * sat * thetam`.

`TspFmiType%gwfsatold` is used by its callers as the value of that same saturation at the end of the previous time step, but it is not calculated as one (`src/Model/TransportModel/tsp-fmi.f90:503-523`).

```fortran
vcell = this%dis%area(n) * (this%dis%top(n) - this%dis%bot(n))
vnew = vcell * this%gwfsat(n)                                    ! no porosity
vold = vnew
if (this%igwfstrgss /= 0) vold = vold + this%gwfstrgss(n) * delt  ! water volume
if (this%igwfstrgsy /= 0) vold = vold + this%gwfstrgsy(n) * delt  ! water volume
satold = vold / vcell
```

The first term is a cell volume scaled by a geometric fraction and has no porosity in it, and the two storage terms are volumes of water. Adding them together produces a quantity that is neither the saturation of the previous time step nor a water content, and dividing by the cell volume does not make it one.

`mst_fc_sto` performs the same reconstruction of the previous water volume and does include the porosity (`src/Model/GroundWaterTransport/gwt-mst.f90:244-249`), so the two disagree within the same package:

```fortran
vnew = this%dis%area(n) * (this%dis%top(n) - this%dis%bot(n)) * &
       this%fmi%gwfsat(n) * this%thetam(n)
vold = vnew
if (this%fmi%igwfstrgss /= 0) vold = vold + this%fmi%gwfstrgss(n) * delt
if (this%fmi%igwfstrgsy /= 0) vold = vold + this%fmi%gwfstrgsy(n) * delt
```

The value returned by `gwfsatold` is the current saturation plus the released water volume divided by the cell volume, so the change in saturation over a time step comes out as `sy * dh / thickness` rather than the change in the position of the water table, `dh / thickness`.

## Example

A cell with a top of 40, a bottom of 0, an area of 1e4, a porosity of 0.3 and a specific yield of 0.3, where the head falls from 36 to 32 over one time step:

| quantity | value |
| --- | --- |
[...]

PR #2918, opened 2026-08-12
feat(gwt): hold solute back when a cell drains

jdhughes-dev:
Adds a STRANDED_MASS option to the MST package. The mobile domain scales the solute mass of a cell by its saturation, so when the water table falls the solute the drained part held is returned to the water that remains and the simulated concentration rises. With this option that solute is instead held out of the mobile domain until the cell rewets, and the mass returned is the mass the same cell gave up, so no neighboring cell is consulted. Two quantities are held back: the solute in the water retained against drainage, which follows from POROSITY and the specific yield of the STO package rather than from new input, and the sorbed mass, because the solid aquifer material does not drain.

- **Results change for any unconfined model whose specific yield differs from its porosity.** The storage terms take the previous saturation from a stored value rather than reconstructing it from the flow to storage, which is correct only when the two are equal (#2921), and the sorbed term's reconstruction omits the porosity entirely (#2917). Nothing is held back where the specific yield equals the porosity; without sorption the peak mass held reaches 49 percent of the mass of a cell at half the porosity and 67 percent at one sixth, and sorbed mass adds to that. Results are bit-identical with the option off.
- **Conservation.** Over a drainage and rewetting cycle the mobile and stranded mass sum is conserved to 6 parts in 24 million, in both directions of the inequality, against 5.7 percent unaccounted at half the porosity and 16.3 percent above it with the option off.
- **Requires the STO-SY flow term when flows are read from a budget file**, which the flow model writes only if its STO package saves flows and uses specific yield. Without it the retained volume is unknowable and every drained pore volume would be treated as retained, so it is an error. A confined flow model is indistinguishable from one that did not save its storage flows and is the same error. Results are identical to a coupled simulation when the term is present.
- **Mass is stranded from the first time step.** The saturation a simulation starts from is not stored by the flow model, so the NPF package now evaluates the initial heads on request and the exchange hands the result to the transport model; a model draining from its first stress period previously left 9.2 percent of the initial mass unaccounted for. A budget file cannot supply that saturation, so a transport model reading flows from a file strands no mass during the first time step and warns if the water table moved.
- **An error with IST**, which occupies the same cells as the mobile domain, and **a warning with UZT**, which represents the same drained volume but often covers only part of a model.
- **Cells that drain completely** are deactivated for transport; their reservoirs are retained, continue to decay, and are returned on rewetting. The saturation of an inactive cell is recorded so the step that reactivates it reads as rewetting.
- **A warning** is written where the specific yield exceeds the porosity, in which case no water is retained and only sorbed mass is stranded.
- **Runs in parallel.** A reservoir belongs to a cell rather than a connection, so nothing crosses an exchange; a domain split into two models gives the same answer on two processes as on one, within the ordinary difference between the parallel and the serial solver, which is the same with the option on as with it off.

A chapter in the Supplemental Technical Information document gives the formulation, an oscillating water table example, the sensitivity to the specific yield and the distribution coefficient, an example of a cell that drains completely and is later rewet, an assessment of what the approach does and does not represent, and its relation to MT3D-USGS, which strands sorbed mass on the same geometric share but represents neither the solute in retained water nor decay while it is held.

- [x] Referenced issue or pull request #2917
- [x] Referenced issue or pull request #2921


emorway-usgs:
Years ago, @vivekbedekar drafted a manuscript that used MT3D-USGS to explore different approaches for handling sorbed mass that is either stranded when the water table falls or 'reanimated' when the water table rises.  The work is not mine to discuss at length, but if additional options are sought, @vivekbedekar is a great resource for this sort of thing.

vivekbedekar:
I have a manuscript that explores several different formulations for handling adsorbed mass in transient systems. Maybe it’s time to dust it off and revive it. Happy to share it with you, @emorway-usgs or @jdhughes-dev, just drop me an email if you’re interested.

jdhughes-dev:
Thanks Vivek @vivekbedekar could you attach the document here.

PR #2944, opened 2026-08-26
docs(dis): clarify vertical pass-through cell requirements

wpbonelli:
Pass-through connections only remove hydraulic resistance, they don't change top/bot, so pass-through cells must have negligible thickness to avoid a gap in the grid and a resulting error. I'm not sure on the wording, or if this is warranted in general, but maybe stating it more clearly somehow could address e.g. #2942


doc/mf6io/mf6ivar/dfn/olf-dis2d.dfn
emorway-usgs, line 1: @christianlangevin, I need to get up to speed on OLF, but I don't think the edited language applies and the original language needs some clarification.  There are no vertical connections in OLF, so the last two sentences don't make any sense.

Additionally, if an adjacent cell doesn't exist, can there be horizontal flow through connections?

wpbonelli: Good catch, thanks — reverted the OLF files (olf-dis2d.dfn, olf-disv2d.dfn) back to the original wording; the vertical-neighbor/TOP-BOTM language doesn't apply to a single-layer, horizontally-connected grid. Left the clarification on GWF/GWT/GWE/PRT where cells actually stack vertically. Leaving the horizontal-pass-through and OLF-specific wording questions open for you and @christianlangevin.

christianlangevin: Is there anything I need to look at here?  It is true that there are no vertical connections for the 2d grids used by OLF.

wpbonelli:
reverted OLF. whoops.

wpbonelli:
Tried to rephrase to make the DFN text clearer. Also figured we could improve the non-positive thickness error so it's more obvious what's going on when someone runs into this, by mentioning that

* for layers > 1, cell top is the bottom elevation of the cell above, and naming the overlying cell, and
* for inactive/VPT cells, cell geometry doesn't change

So if the cell above is not inactive/VPT, the error would now be

```
1. CELL (4,1,2) THICKNESS <= 0. TOP, BOT:     4.00000000000000
     5.00000000000000. The top of this cell is the bottom elevation specified
     for the cell directly above it, cell (3,1,2).
```

And for the situation in #2942 it would be

```
1. CELL (4,1,2) THICKNESS <= 0. TOP, BOT:     4.00000000000000
     5.00000000000000. The top of this cell is the bottom elevation specified
     for the cell directly above it, cell (3,1,2). That cell is inactive or a
     vertical pass-through cell (IDOMAIN <= 0); making a cell inactive does
     not change the geometry of its neighboring cells. Set its BOTM so this
     cell has positive thickness (for a pass-through cell, set its BOTM equal
     to the bottom elevation of the cell above it, giving it negligible
     thickness).
```

Is this too verbose?

Issue #2946, opened 2026-08-26
fix(maw): in-screen length of a non-vertical connection is not limited by the cell


jdhughes-dev:
The in-cell screen length of a non-vertical multi-aquifer well connection grows without bound as the tilt angle approaches horizontal, and nothing reports it. The length is

```
lw = (dz - 2 * radius * sin(omega)) / cos(omega)
```

computed in `maw_read_angledata` (`src/Model/GroundWaterFlow/gwf-maw.f90:1248`) and again in `maw_calc_lcorr` (`:1365`), where it scales the saturated conductance of the connection.

A connection length is required only when `cos(omega)` is within `1e-6` of zero (`:1237`), which is about 0.006 degrees of horizontal. Below that the length grows silently. For a cell 100 by 100 in plan and 10 thick, with a well radius of 0.1 and a screen spanning the cell:

| ANGLE | in-cell screen length | horizontal run | fits in the cell |
| --- | --- | --- | --- |
| 60 | 19.7 | 17.0 | yes |
| 75 | 37.9 | 36.6 | yes |
| 80 | 56.5 | 55.6 | yes |
| 85 | 112.5 | 112.0 | no |
| 89 | 561.5 | 561.4 | no |

At 85 degrees the screen runs 112 units through a cell 100 wide, and at 89 degrees it runs 561. Those lengths are not merely large; a screen cannot be longer than the cell it is connected to. The conductance is scaled by the same factor, so a connection at 89 degrees is given roughly 56 times the conductance of the vertical equivalent.

The user can set CONN\_LENGTH to limit the length, but has no way to know the angle at which the limit becomes necessary, and the angle depends on the shape of the cell.

## Proposed change

Limit the length to the diagonal of the cell in the plane of the tilt,

```
lmax = sqrt(area + dz * dz)
```

and write a warning when the limit is reached. The area is used rather than the side lengths of the cell because it is the only measure of horizontal extent that every discretization package provides; there is no `delr` or `delc` on the discretization base class, so a rule written in terms of cell sides would work only for DIS.

The limit does not change the length at or below about 80 degrees, so a model whose geometry is realizable is unaffected. It applies only where the screen has left the cell, and in that case the length is usually not the thing to correct: a well that runs 561 units horizontally passes through several cells and wants a connection to each of them. The warning should say so.

CONN\_LENGTH continues to override the calculated length.

## Also

[...]

PR #2949, opened 2026-08-27
feat(maw): warn when a non-vertical connection extends beyond its cell

jdhughes-dev:
A multi-aquifer well connection listed in the ANGLEDATA block spans a horizontal distance of LW times sin(ANGLE), which grows without bound as the tilt angle approaches horizontal and was not checked, so a screen extending well beyond the cell it is connected to was silently given a saturated conductance that was correspondingly too large. The horizontal distance is now compared to the maximum horizontal extent of the connected cell and a warning is written when it is exceeded, both for a length derived from the screen elevations and for one specified with CONN_LENGTH. The cell extent is the largest distance between two cell vertices for the DIS and DISV grids and the diagonal of a square with the same area for the DISU grid, which does not require vertices; the two are equal for a square cell.

Also adds a figure and text to the supplemental technical information describing the horizontal distance spanned by a non-vertical connection.

- [x] Referenced issue or pull request #2946


src/Model/GroundWaterFlow/gwf-maw.f90
```diff
@@ -1258,6 +1269,25 @@ subroutine maw_read_angledata(this)
...
+          if (hlen > extent) then
+            write (warnmsg, '(a,1x,i0,1x,a,1x,i0,1x,a,g0,a,g0,a)') &
+              'The horizontal distance spanned by maw well', n, 'connection', &
+              j, '(', hlen, &
+              ') is greater than the maximum horizontal extent of the '// &
+              'connected cell (', extent, &
+              '), so the screen extends beyond the cell it is connected '// &
+              'to and the calculated saturated conductance is too large. '// &
+              'Reduce ANGLE or CONN_LENGTH, or specify a separate '// &
+              'connection to each cell the well penetrates.'
```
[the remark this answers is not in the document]
jdhughes-dev: Good catch, fixed in 0426ff64.

The check is applied to SPECIFIED connections on purpose, because the screen is in the wrong cell whichever equation is used, but you are right that the program has no business calling a conductance it did not calculate too large. The conductance clause now depends on the equation:

```
so the screen extends beyond the cell it is connected to. The specified
saturated conductance is applied to a screen that is longer than the cell.
Reduce ANGLE or CONN_LENGTH, or specify a separate connection to each cell
the well penetrates.
```

The rest of the warning is unchanged. A fifth well using SPECIFIED was added to `test_gwf_maw_nonvertical_extent.py`, and the test now checks which clause each warning carries.

src/Utilities/GeomUtil.f90
```diff
@@ -403,6 +403,31 @@ function area(xv, yv, cw) result(a)
...
+    ! result
+    real(DP) :: e
+    ! local
+    integer(I4B) :: i
+    integer(I4B) :: j
+    real(DP) :: dx
+    real(DP) :: dy
+
+    e = DZERO
+    do i = 1, size(xv) - 1
```
aprovost-usgs, line 420: probably unlikely, but if you find in the future that this O{n^2) approach impacts the runtime in some cases, there's apparently a "rotating calipers" approach that's O(n) for a convex polygon

src/Model/GroundWaterFlow/gwf-maw.f90
```diff
@@ -1321,6 +1363,53 @@ subroutine maw_read_angledata(this)
...
+    real(DP), allocatable, dimension(:, :) :: polyverts
+    !
+    select case (this%dis%get_dis_enum())
+    case (DIS, DISV)
+      call this%dis%get_polyverts(node, polyverts)
+      extent = polygon_extent(polyverts(1, :), polyverts(2, :))
+      deallocate (polyverts)
+    case default
+      !
+      ! -- disu does not define cell vertices, so the extent is taken from the
```
aprovost-usgs: cell vertices can be defined for DISU (e.g., when they're needed for XT3D), so they're available if the user provided them

src/Model/GroundWaterFlow/gwf-maw.f90
```diff
@@ -1258,6 +1270,36 @@ subroutine maw_read_angledata(this)
...
+              cndmsg = 'The specified saturated conductance is applied to a '// &
+                       'screen that is longer than the cell.'
+            else
+              cndmsg = 'The calculated saturated conductance is '// &
+                       'correspondingly too large.'
+            end if
+            write (warnmsg, '(a,1x,i0,1x,a,1x,i0,1x,a,g0,a,g0,a)') &
+              'The horizontal distance spanned by maw well', n, 'connection', &
+              j, '(', hlen, &
+              ') is greater than the maximum horizontal extent of the '// &
```
aprovost-usgs: not sure it's worth splitting this hair, but could say "estimated horizontal extent" when it's estimated from the area

doc/SuppTechInfo/maw-nonvertical-connections.tex
```diff
@@ -73,6 +73,23 @@ \subsection{Non-Vertical Connection Geometry}
 
 \noindent and a connection length is required for horizontal connections.
 
+The horizontal distance spanned by a connection is $L_w \sin \omega$, which grows without bound as the tilt angle approaches horizontal (fig.~\ref{fig:mawnvextent}).
+For a cell 20 length units long and 10 length units thick with a screen that spans the cell thickness, the horizontal distance spanned by the connection is 17 length units at $\omega = 60^{\circ}$ but 112 length units at $\omega = 85^{\circ}$, which is nearly six times the length of the cell.
+A screen cannot be longer than the cell it is connected to, and a connection that spans more than the connected cell is given a saturated conductance that is correspondingly too large.
+The \mf~program compares the horizontal distance spanned by a connection to the maximum horizontal extent of the connected cell and writes a warning when it is exceeded.
+The maximum horizontal extent of a cell is calculated from the cell vertices for the DIS and DISV grids and from the cell area for the DISU grid, which does not require them.
```
aprovost-usgs: suggest: "... which does not require vertices."

doc/ReleaseNotes/develop.toml
```diff
@@ -148,7 +148,7 @@ description = "Reverts a 6.4.3 change that reset auxiliary variable arrays to ze
...
+description = "The Multi-Aquifer Well (MAW) Package now supports non-vertical (slanted or horizontal) well connections. When the NON\\_VERTICAL\\_WELLS option is specified, an ANGLEDATA block can be used to assign a tilt angle (deviation from vertical) to individual well connections. The saturated conductance for those connections is scaled by the in-cell screen length, which is calculated from the screen top, screen bottom, well radius, and tilt angle (or specified directly with an optional connection length), instead of the vertical screen thickness. A connection length must be specified for horizontal connections, and the MEAN conductance equation is recommended for horizontal connections. A warning is issued if the horizontal distance spanned by a connection is greater than the maximum horizontal extent of the connected cell, which indicates that the well should be specified with a separate connection to each cell that it penetrates."
```
aprovost-usgs: In the warning message it says, "Reduce ANGLE or CONN_LENGTH, or specify a separate connection to each cell the well penetrates."  Suggest modifying this description along those lines:
"... which indicates that the tilt angle or specified connection length should be reduced, or the well should be specified with a separate connection ..."

doc/mf6io/mf6ivar/dfn/gwf-maw.dfn
```diff
@@ -564,7 +564,7 @@ tagged false
...
+description value that defines the tilt angle of the multi-aquifer well connection, in degrees measured as a deviation from vertical.  ANGLE must be greater than or equal to 0.0 (a vertical connection) and less than or equal to 90.0 (a horizontal connection) degrees.  The in-cell screen length used to calculate the saturated conductance is computed from the screen top, screen bottom, well radius, and ANGLE, unless CONN\_LENGTH is specified.  The horizontal distance spanned by a connection grows rapidly as ANGLE approaches 90.0 degrees, and a warning is issued if it is greater than the maximum horizontal extent of the connected cell; a well that passes through more than one cell should be specified with a separate connection to each cell that it penetrates.
```
aprovost-usgs: maybe: "... a connection that is intended to pass through more than one cell should be ..."  And maybe add something like: "If the connection spans more than one cell unintentionally, the tilt angle or specified connection length should be reduced."

note that the above suggestions also use "connection" instead of "well," since users might think of the well as consisting of more than just the "connection".  could alternatively use "well screen"

doc/SuppTechInfo/Figures/MAWNonVerticalExtent.pdf
aprovost-usgs, line 1: Good idea to have this figure, and nicely done.  Might it be good to mention in the caption that while the connection is shown to be extending into cells to the right, that's just one possibility shown for the sake illustration?  The calculation done in MAW doesn't actually determine the direction(s) in which the connection spills over, and it's up to the user to decide that.

doc/ReleaseNotes/develop.toml
aprovost-usgs: please see my related comment in gwf-maw.dfn

aprovost-usgs (changes requested):
nice job, just minor suggestions

src/Utilities/GeomUtil.f90
jdhughes-dev: Noted, and worth recording here for whoever hits it. Rotating calipers would be O(n) after the convex hull, and cell polygons are convex, so the hull is free.

Leaving it as is for now: it runs once per ANGLEDATA entry at AR time, over cell polygons that are 4 to 8 vertices in practice, so n^2 is under 30 distance evaluations. The crossover with the calipers bookkeeping is well above any real cell.

src/Model/GroundWaterFlow/gwf-maw.f90
jdhughes-dev: You are right, and I had this wrong. `DisuType` carries `nvert`, `vertices`, `iavert`, and `javert` whenever the optional VERTICES and CELL2D blocks are given; what was missing was the accessor, since `Disu.f90` never overrode `get_polyverts` and the base version is a `terminate=.true.` programmer error. I read that as "no vertices" rather than "no accessor."

Fixed in 0d47e9bf. `get_polyverts`, `get_npolyverts`, and `get_max_npolyverts` are now implemented for DISU, mirroring the DISV versions; `iavert` and `javert` are built by the same `define_cellverts` code, indexed by user node rather than by cell2d index. `get_npolyverts` returns 0 when the blocks were not given, so `maw_cell_extent` asks whether the vertices are there instead of testing the grid type, and the area estimate is now only used when they are genuinely absent.

A fourth case was added to `test_gwf_maw_nonvertical_extent.py` using `get_disu_kwargs(..., return_vertices=True)`. It reports the same extent as the DIS and DISV grids of the same cells, and reports it as calculated rather than estimated.

src/Model/GroundWaterFlow/gwf-maw.f90
jdhughes-dev: Worth splitting, agreed, and it matters less now that DISU uses its vertices when they are there. The warning distinguishes the two:

```
) is greater than the maximum horizontal extent of the connected cell (
) is greater than the estimated maximum horizontal extent of the connected cell (
```

`maw_cell_extent` returns an `estimated` flag alongside the value, set when the extent came from the area rather than the vertices, which is now only a DISU grid without the VERTICES and CELL2D blocks. The test asserts which of the two wordings each grid produces.

doc/SuppTechInfo/maw-nonvertical-connections.tex
jdhughes-dev: Taken, though the sentence changed with the DISU vertices fix above:

> The maximum horizontal extent of a cell is calculated from the cell vertices where they are defined and is estimated as the diagonal of a square with the same area where they are not, which can happen for a DISU grid because the VERTICES and CELL2D blocks are optional.

doc/ReleaseNotes/develop.toml
jdhughes-dev: Done, in 0d47e9bf:

> A warning is issued if the horizontal distance spanned by a connection is greater than the maximum horizontal extent of the connected cell, which indicates that the tilt angle or specified connection length should be reduced, or that the connection should be specified as a separate connection to each cell that it passes through.

doc/mf6io/mf6ivar/dfn/gwf-maw.dfn
jdhughes-dev: Both taken. The ANGLE description now reads:

> The horizontal distance spanned by a connection grows rapidly as ANGLE approaches 90.0 degrees, and a warning is issued if it is greater than the maximum horizontal extent of the connected cell.  A connection that is intended to pass through more than one cell should be specified as a separate connection to each cell that it passes through.  If a connection passes through more than one cell unintentionally, ANGLE or CONN\_LENGTH should be reduced.

Your point about "well" versus "connection" is the better one, and I have carried it through the rest of the changes rather than just this description: the warning message, the release note, and the paragraph and figure caption in the supplemental technical information all say "connection" now. A MAW well is a set of connections, and every one of these statements is about a single connection.

doc/SuppTechInfo/Figures/MAWNonVerticalExtent.pdf
jdhughes-dev: Thank you, and yes, that is the right thing to say. The program compares a distance to a distance; nothing in MAW knows which way the screen leans, and ANGLE carries no azimuth. Drawing the cells to the right made a scalar look like a direction.

The caption now ends:

> The screen is drawn extending to the right for illustration only; the direction a screen is oriented in is not specified and is not determined by the program.

The chapter already made the consequence explicit a few lines above, which this now matches:

> A tilt angle is not accompanied by an azimuth, so the comparison cannot account for the direction the screen is oriented in and a screen can extend beyond a cell that is much longer in one horizontal direction than the other without a warning being written.

doc/ReleaseNotes/develop.toml
jdhughes-dev: Addressed together with the gwf-maw.dfn comment; the release note wording is quoted there.

aprovost-usgs (approved):
Thanks for the changes and responses - looks good

PR #2952, opened 2026-08-30
fix(ats): honor a submitted stable time step when DTADJ is zero or one

christianlangevin:
## The problem

`ats_submit_delt` (`src/Timing/ats.f90`) is how a package tells the ATS
Package that the next time step must be no longer than some value. It has two
kinds of caller:

| caller | `idir` | meaning of `dt` |
|---|---|---|
| `NumericalSolution` | present | the current step, to be *scaled* by `DTADJ` in response to the outer iteration count |
| GWT ADV, GWE ADV | absent | an **absolute** stability limit: the advective Courant limit from `ATS_PERCEL` |
| GWF SFR | absent | an **absolute** stability limit: the kinematic wave limit |

The entire body of the routine sat inside `if (tsfact > DONE)`, where
`tsfact` is the period's `DTADJ`. That test belongs to the scaled path only.
`utl-ats.dfn` documents `DTADJ` as "must be zero, one, or greater than one. If
dtadj is zero or one, then it has no effect on the simulation", and for the
scaled path that is exactly what the guard delivers. But it also swallowed the
absolute submissions, so with `DTADJ` set to zero or one:

* `ATS_PERCEL` was read, validated, and then ignored;
* SFR's kinematic wave limit was ignored;
* and in both directions — the step was neither lowered to the stable value
  nor raised toward it — because `ats_set_delt` assigns `dtstable` outright on
  every step after the first of a period.

Found while comparing transport schemes with `DTADJ = 0`, when a 31x31 problem
took 581,704 steps of a frozen `DT0` instead of 1,460.

## The change

Confine the `tsfact > DONE` test to the `present(idir)` branch. The scaled
path is unchanged; the absolute path (`dt_temp = dt`) now always runs, and
the shared tail — log the submission on `kstp > 1`, lower `dtstable` if the
submission is smaller — executes whatever `DTADJ` is. `ParallelAts` forwards
to the base routine and needs no change.

Behavior with `DTADJ > 1` is bit-for-bit what it was; only the `DTADJ = 0`
and `DTADJ = 1` cases change, and only in that they now do what the
documentation says.

## Test

`autotest/test_tsp_adv_ats_dtadj.py` reuses the 1D column of
`test_tsp_adv_ats.py` (a well injecting 1.0 into the first of 100 unit cells
at porosity 0.1, so water crosses a cell in 0.1 d and `ATS_PERCEL = 0.5`
limits the step to 0.05 d) with upstream weighting, and runs four cases
covering both transport models, both `DTADJ` values, and both directions:

| case | model | `DTADJ` | `DT0` | step must | expected steps | steps if ignored |
|---|---|---|---|---|---|---|
| `advdtadj0gwt` | GWT | 0 | 0.01 | grow | 101 | 500 |
| `advdtadj1gwt` | GWT | 1 | 1.0 | shrink | 81 | 5 |
| `advdtadj0gwe` | GWE | 0 | 1.0 | shrink | 81 | 5 |
| `advdtadj1gwe` | GWE | 1 | 0.01 | grow | 101 | 500 |

Each case asserts that the first step is `DT0` (ATS does not apply a
submission on the first step of a period), that every later step is at most
the ADV limit, and that the total step count is exact.

## Verification

Built on branch `fix-ats-submit-delt` (worktree `modflow6-atsfix`, off
`develop` at 80d04dc) and run on 2026-08-30:

* `test_tsp_adv_ats_dtadj.py`: 4 passed with the fix.
* Negative control — `ats.f90` reverted, rebuilt, same test: 4 failed exactly
  as predicted (`took 500 time steps, expected 101` for the grow cases;
  `steps exceed the ADV limit 0.05: max 1.0` for the shrink cases).
* `test_tsp_adv_ats.py` (`DTADJ = 2`, the previously working path): 2 passed,
  unchanged.
* `test_gwf_sfr_kinematic_ats.py` could not be run: the flopy in the pixi
  environment predates the `ATS_COURANT` option added to SFR in #2818 and
  rejects the kwarg before any model is built. Not related to this change.
* `check_format.py` / fprettify: no changes to `ats.f90`; `ruff` clean on
  the test.


autotest/test_tsp_adv_ats_dtadj.py
```diff
@@ -0,0 +1,236 @@
...
+        reordering_method="NONE",
+        relaxation_factor=relax,
+        filename=f"{name}.ims",
+    )
+
+
+def build_models(idx, test):
+    name = list(cases)[idx]
+    mtype, dtadj, dt0 = cases[name]
+
```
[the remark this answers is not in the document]
christianlangevin: Agreed, addressed in 1c06945: the test is now parametrized on the case name alone and the helpers index `cases[name]` directly.

wpbonelli (approved):
Looks good to me. Only thought is, maybe it'd be clearer to have two routines instead of one whose semantics change depending on parameters, like `ats_submit_delt` for the absolute and `ats_scale_delt` for the scaled case? Maybe overkill.

christianlangevin:
> Looks good to me. Only thought is, maybe it'd be clearer to have two routines instead of one whose semantics change depending on parameters, like `ats_submit_delt` for the absolute and `ats_scale_delt` for the scaled case? Maybe overkill.

Thanks, @wpbonelli.  Let me test that.

christianlangevin:
> Looks good to me. Only thought is, maybe it'd be clearer to have two routines instead of one whose semantics change depending on parameters, like `ats_submit_delt` for the absolute and `ats_scale_delt` for the scaled case? Maybe overkill.

This is a good suggestion, but I think it should wait for now as it also requires additional changes for the parallel case.  Once this is in, we can evaluate splitting into two routines.  The bug that this PR fixes probably would have been caught earlier if the routines were better named and `idir` was dropped.

doc/ReleaseNotes/develop.toml
```diff
@@ -60,6 +50,11 @@ section = "fixes"
...
...n = "A time step length submitted to the adaptive time stepping (ATS) Package by a package that needs to limit the step for its own stability was discarded whenever DTADJ for that period was zero or one.  DTADJ is the multiplier ATS applies when it lengthens or shortens the step in response to the solver iteration count, and the documentation states that a value of zero or one means it has no effect on the simulation.  It should have no bearing on a step length that a package submits directly, because ATS assigns such a length to the time step rather than scaling it.  Three packages submit a step this way: the ADV Package of the GWT and GWE models submits the advective Courant limit, and the SFR Package of the GWF model submits the kinematic wave limit.  With DTADJ set to zero or one those limits had no effect at all, in either direction, so the time step was neither raised nor lowered toward the stable value.  In particular the ATS\\_PERCEL option of the ADV Package, which limits the fraction of a cell that may be traversed in one time step, was accepted without complaint and then ignored.  The test that discards the submitted length now applies only where DTADJ is actually used."
```
aprovost-usgs: The sentence "It should have no bearing ..." seems like it could lead to misunderstanding.  Consider replacing with something like, "In that case, the step lengths submitted by packages should be honored."

aprovost-usgs (approved):
One minor comment for your consideration.  Please merge whenever you're ready

doc/ReleaseNotes/develop.toml
christianlangevin: Good suggestion.  Change was made.

Issue #2940, opened 2026-08-25
More clearly describe the Print and Save options in the PRT Output Control description


rbwinst-usgs:
**Is your feature request related to a problem? Please describe.**
It appears that the SAVE BUDGET option in the Period section of the PRT output control file refers to saving the budget in the binary budget file and thus requires that BUDGET FILEOUT be specified in the Options section. Similarly PRINT BUDGET requires that BUDGETCSV FILEOUT be specified in the Options section. This wasn't entirely clear to me.

**Describe the solution you'd like**
I suggest that the description of budgetfile and budgetcsvfile be changed as follows
budgetfile—name of the output file to write budget information when SAVE BUDGET <ocsetting> is specified in the PERIOD block.
budgetcsvfile—name of the comma-separated value (CSV) output file to write budget summary information.
A budget summary record will be written to this file for each time step of the simulation when PRINT BUDGET <ocsetting> is specified in the PERIOD block.

wpbonelli:
Updating the description as you suggest makes sense to me. I think this is true of all models, not just PRT, right?

wpbonelli:
Tried to clarify how this works in the linked PR. It is true that if `SAVE BUDGET` appears for any period, a binary budget output file must be set, or an error will be raised. But `PRINT BUDGET` hasn't ever gated the CSV budget output file for PRT or any other model, as far as I can see. The main behavioral difference between binary and CSV output files is the OC period settings don't control CSV file output, every period is written to CSV files regardless.

PR #2954, opened 2026-08-31
docs(oc): clarify binary and csv budget file options

wpbonelli:
Address #2940. Maybe we can follow up at some point with a new IO guide section describing how output controls work in a unified way.

### DIFF
--- a/doc/mf6io/mf6ivar/dfn/chf-oc.dfn
+++ b/doc/mf6io/mf6ivar/dfn/chf-oc.dfn
@@ -43,7 +43,7 @@ reader urword
 tagged false
 optional false
 longname file keyword
-description name of the output file to write budget information.
+description name of the output file to write budget information. Data are written to this file only for time steps in which SAVE BUDGET is specified in a PERIOD block. If SAVE BUDGET is specified in a PERIOD block and this file is not specified, MODFLOW 6 will terminate with an error.
 
 block options
 name budgetcsv_filerecord
@@ -77,7 +77,7 @@ reader urword
 tagged false
 optional false
 longname file keyword
-description name of the comma-separated value (CSV) output file to write budget summary information.  A budget summary record will be written to this file for each time step of the simulation.
+description name of the comma-separated value (CSV) output file to write budget summary information. If specified, a budget summary record will be written to this file for every time step of the simulation.
 
 block options
 name qoutflow_filerecord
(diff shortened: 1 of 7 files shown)


doc/mf6io/mf6ivar/dfn/chf-oc.dfn
```diff
@@ -43,7 +43,7 @@ reader urword
 tagged false
 optional false
 longname file keyword
-description name of the output file to write budget information.
+description name of the output file to write budget information. Data are written to this file only for time steps in which SAVE BUDGET is specified in a PERIOD block. If SAVE BUDGET is specified in a PERIOD block and this file is not specified, MODFLOW 6 will terminate with an error.
```
aprovost-usgs, line 46: Just want to confirm that MF6 will terminate with an error if the file isn't specified - not just  silently decline to save the budget

doc/mf6io/mf6ivar/dfn/chf-oc.dfn
```diff
@@ -77,7 +77,7 @@ reader urword
 tagged false
 optional false
 longname file keyword
-description name of the comma-separated value (CSV) output file to write budget summary information.  A budget summary record will be written to this file for each time step of the simulation.
+description name of the comma-separated value (CSV) output file to write budget summary information. If specified, a budget summary record will be written to this file for every time step of the simulation.
```
aprovost-usgs, line 80: this covers csv files for the models, and i think that's good start for now.  there are also csv files for packages, some of which are more nuanced because they involve internally determined conditions for when/whether output actually occurs.  can revisit that in the future to see if they need clarification, as well

aprovost-usgs (approved):
just a couple of comments for your consideration.  please go ahead and merge whenever you're ready

doc/mf6io/mf6ivar/dfn/chf-oc.dfn
wpbonelli: Yes, it will terminate with an error like

```
REQUESTING TO SAVE BUDGET BUT BUDGET SAVE FILE NOT SPECIFIED
```

Likewise for `HEAD`, `CONCENTRATION`, `TEMPERATURE`, `STAGE`, etc. Added a test case to back the guarantee. Should we add similar text to the DFNs for the other output variables too?

aprovost-usgs: @wpbonelli that's probably a good idea, but i propose we take a holistic look at how best to clarify these aspects of the documentation after the imminent release

PR #2958, opened 2026-09-02
fix(GeomUtil): fix inverse compose

wpbonelli:
Particle coordinate transformations were not correctly inverted, causing positions to be reported incorrectly on rotated quad-refined grids. This could produce discontinuities in the pathline without disturbing the overall path. Thanks to @rbwinst-usgs for reporting this.

prtbug5_compare_full

PRT maintains the cumulative particle tracking transform as a pair `(offset, rotation)`, e.g. $(t, \alpha)$. The hierarchical tracking methods each compose their own transform $(a, \varphi)$ on the cumulative transform $(t, \alpha)$ to produce $(t', \alpha')$. 

Forward and inverse were separate hand-derived formulas, split over conditional branches. The inverse dropped a rotation, causing a spurious term to be added to every coordinate written to output files until the next coordinate reset between cells. #2675 traded an unbounded form of this "dropped the R" error, where error could accumulate, for a bounded one, where the error is small compared to the size of the cell, but scales with cell size and with the rotation angle.

The overall pathline was/is not affected since the tracking method uses direct point transformations, which work(ed) correctly, while the particle event reporting logic accumulates transformations in the way described above.

Rewrite both directions of `compose` in a simpler way. Add unit tests for `transform` and `compose` and update the integration tests to check that a particle is never in two different lateral positions at a single time. This seems like a good invariant to check across all PRT tests as it would have caught this. (Different vertical positions at the same time are fine — a dry particle may drop instantly to the water table.)

___


aprovost-usgs (approved):
looks good

PR #2985, opened 2026-09-14
fix(prt): prevent duplicate releases

### LINKED ISSUE #2980: Budget error in PRT when particles attempt to be released after stop time
**Describe the bug**
If a PRT model has a stop time specified and also has a particle released after the stop time, the particle is not released which is good. However, there is a budget error reported for that time step which seems incorrect.

**To reproduce**
run the attached model and look at the budget for the first time step in stress period 3

**Expected behavior**
No budget error.

**Screenshots**

**Environment**
 - Operating system Windows 11
 - MODFLOW 6 version VERSION 6.9.0.dev0+073acdc.dirty

[bug6.zip](https://github.com/user-attachments/files/32144533/bug6.zip)

wpbonelli:
Fix #2980. PRT could erroneously release multiple particles from the same release point at the same time, causing an unbalanced mass budget, when a PRP package's period-block release setting coincided with an explicitly configured release time landing on a time boundary between two steps. The resulting pair of otherwise-identical particles were not distinguished in internal accounting, and could silently corrupt the model's mass budget.

PRT distinguishes particles on the basis of four properties (model index, package index, release point index, release time), and uses a "gumball dispenser" metaphor for release points, stipulating that multiple particles may not be released from a single point at the same time. Coincident releases are now consolidated into a single release as intended, and reported under the earlier of the two time steps for times falling on a time step boundary.

Add a `TimeSelectType%contains_close()` routine that checks a period-block candidate against the full, static set of explicit release times, which is completely known before the time step loop starts.

Also gate `prt_solve`'s release check on the particle's status rather than `trelease >= totimc` alone, since that condition could otherwise be satisfied a second time by an already-released particle whose release time lands exactly on a later step's start.

In the tests, make `check_budget_data()` check that the budget actually balances at every time step.

___

- [x] Closed issue #2980
- [x] Formatted new and modified Fortran source files with fprettify
- [x] Updated definition files
- [x] Updated develop.toml with a plain-language description of the bug fix


wpbonelli:
@aprovost-usgs I think we had a case of conflicting design principles here. We explicitly designed particle identity and release logic around the idea that you can only release one particle per point per instant. But at some point we also decided releases could coincide, distinguished just by kper/kstp, if a release time fell at the very end of the time step, and the next time step also had an immediate release. That would work if we included release kper/kstp in particle identity, or had some kind of explicit particle ID, but with the composite key approach (imdl, iprp, irpt, trelease) we can't distinguish particles released at the same time, even if it happens in separate time steps.

I figured better to stick with the existing release metaphor and particle identity and sacrifice a bit of intuitiveness. The consequence is, a particle release may fall on the left or right side of a time step boundary, depending whether it's configured in the stress period settings (right), or with an explicit time (left), and when both coincide, the release will fall on the left. Which may be surprising but is just a consequence of the algorithm encountering explicit times that coincide with a time step boundary in the preceding time step.

src/Model/ParticleTracking/prt.f90
```diff
@@ -1090,22 +1090,24 @@ subroutine prt_solve(this, isuppress_output)
...
-                &stop time ', particle%tstop, '; particle will not &
-                &be released.'
-              call store_warning(warnmsg)
-              call this%method%terminate(particle, status=TERM_UNRELEASED)
-            else
-              ! The particle was released this time step; emit a
-              ! release event.
-              call this%method%release(particle)
+          if (particle%istatus == 0) then
+            particle%istatus = ACTIVE ! Set active status in case of release
```
aprovost-usgs, line 1094: this is ok as long as istatus is always reset as needed subsequently.  but could it lead to a problem if the code is modified in the future, and if so, might there be a "safer" default?

src/Model/ModelUtilities/TimeSelect.f90
```diff
@@ -289,4 +291,28 @@ subroutine extend(this, a)
...
+    class(TimeSelectType) :: this
+    real(DP), intent(in) :: t
+    real(DP), intent(in) :: tolerance
+    logical(LGP) :: found
+    integer(I4B) :: i
+
+    found = .false.
+    if (.not. allocated(this%times)) return
+    do i = 1, size(this%times)
+      if (is_close(this%times(i), t, atol=tolerance)) then
```
aprovost-usgs, line 311: since rtol isn't specified, it'll default to DSAME.  just want to confirm that that's what we want in this application.  it does seem like a relative tolerance might be desirable, since the magnitudes of the times can be vastly different in different models/units.  on the other hand, are we simply interested in whether they're "close" relative to the precision to which we can represent them?

aprovost-usgs (commented):
I concur with sticking with the existing release metaphor and particle identity

src/Model/ModelUtilities/TimeSelect.f90
wpbonelli: Good point. I wonder if we can just set `rtol = DZERO` and leave it to the user to adjust `RELEASE_TIME_TOLERANCE` as appropriate for their model.

In this PR it's probably best to keep this `is_close()` call as is, identical with the other one, so that the deduplication is consistent within and between time steps. This PR essentially just extends that deduplication check to the between time step case.

I'll come back to this separately.

src/Model/ParticleTracking/prt.f90
wpbonelli: I think we could make the `release()` routine set `istatus = ACTIVE` instead of doing it here. This is symmetric with the `terminate()` routine and I think it is "safer" in the sense you're getting at, maybe we can talk about it and confirm. I'll make a separate PR as we have done this active status "preassignment" for a while, it isn't introduced here.

PR #2988, opened 2026-09-16
fix(bnd): warn when an observation matches no boundary

### LINKED ISSUE #1940: Issue warning if nothing is observed at an observation location
**Is your feature request related to a problem? Please describe.**
A no data value (3e30) is written if a specific boundary condition (location or boundname) does not exist in a stress period. Based on Issue #1939

**Describe the solution you'd like**
Write a warning message if nothing is observed at an observation location.

**Describe alternatives you've considered**
None.

**Additional context**
None.

KR-Ravindra:
## Problem

An observation on a traditional stress package identifies the location it
observes either by a boundname or by a cellid. `bnd_rp_obs` in
`src/Model/ModelUtilities/BoundaryPackage.f90` searches the package's boundary
list for a match at the start of every stress period, records whether it found
one in `jfound`, and then never tests `jfound`. The procedure only checks
`count_errors()` before returning, so an observation that matches nothing is
accepted silently and `bnd_bd_obs` goes on to report the `DNODATA` value for
it.

Nothing tells the user about it. A misspelled or unused boundname, or a cellid
for a cell that contains no boundary, produces an observation file full of
`DNODATA` that is easy to mistake for a simulated value. Every advanced stress
package already reports this condition -- SFR, MAW and LAK in `sfr_rp_obs`,
`maw_rp_obs` and `lak_rp_obs`, and APT in `tsp-apt.f90` -- so the base class
was the only place where an unmatched observation passed without comment.

Closes https://github.com/MODFLOW-ORG/modflow6/issues/1940

## Fix

Test `jfound` and store a warning naming the observation and its package and
reporting the boundname or the cell that was not matched:

```
No boundary matching boundname "NO_SUCH_WELL" was found for observation
"MISSING_BY_NAME" in package "WEL". The DNODATA value will be returned for
this observation.
```

Two details are worth flagging for review:

* **A warning, not an error.** The advanced packages raise an error here, but
  they can: their features are fixed for the whole simulation. The boundary
  list of a traditional stress package is allowed to change between stress
  periods, so an observation may legitimately match nothing during some of
  them and the run should not be stopped.
* **Reported once, not once per stress period.** Because `bnd_rp_obs` runs
  every stress period, a persistently unmatched observation would otherwise
  produce one warning per period. The message is keyed on the observation and
  package names through the optional `substring` argument of `store_warning`,
  which suppresses a message whose key already appears in a stored warning, so
  each unmatched observation is reported exactly once.

The change is confined to `bnd_rp_obs`. No advanced package is touched, no
input keyword is added, and no definition or project file needs regenerating.

## How tested

New regression test `autotest/test_gwf_obs_not_found.py`. It runs a one-layer
model with constant heads at both ends and a single named well, over two
stress periods, carrying four WEL observations: the well's boundname, a
boundname no well uses, the cell containing the well, and a cell with no
boundary in it. It asserts that

* exactly the two unmatched observations are warned about and the two matched
  ones are not,
* each warning reports the right kind of location (boundname vs. cell) and the
  right location and package,
* each unmatched observation is warned about exactly once despite the two
  stress periods, which is what pins the deduplication, and
* the observation output really does contain `DNODATA` for the unmatched
  observations and the well's rate for the matched ones, so the message's claim
  is checked rather than assumed.

The model terminates normally throughout; the test asserts on warnings, not on
failure.

Verified on macOS (arm64, GNU Fortran 16.2.0, meson):

* The new test passes with the change and fails with the change reverted and
  the binary rebuilt, reporting `warnings were reported for [], expected
  ['missing_by_cell', 'missing_by_name']`.
* `pytest -n 10 -m "not slow" test_gwf_*.py test_gwt_obs01.py test_gwe_obs.py`
  passes, so the new warning does not disturb existing models.
* The modified file compiles clean under `-Dwerror=true`.
* `fprettify -c .fprettify.yaml` leaves the Fortran file unchanged,
  `ruff check` and `ruff format --check` are clean on the new test, and
  `codespell` is clean on all three files.

- [x] Closed issue #1940

This change was written with an AI agent operated by KR-Ravindra, who reviewed
and tested it.


KR-Ravindra:
Round 1 self-review.

Checked against `develop`:

1. `bnd_rp_obs` sets `jfound` in both search branches and never reads it; only `count_errors()` gates the return, so an unmatched observation currently reaches `bnd_bd_obs` with `BndFound = .false.` and is written as DNODATA. The change tests `jfound` right after the search and leaves the loop otherwise untouched.
2. `store_warning(msg, substring)` exists in `SimModule` with the optional dedupe key; `warnmsg` is `MAXCHARLEN` in `SimVariablesModule` and `IDstring` is 200 chars on `ObserveType`, so the writes cannot truncate for legal boundnames.
3. The test asserts the warning count across two stress periods (pins the dedupe), the location kind and value, the package name, and the DNODATA / rate values in the observation CSV. It fails with the Fortran change reverted.
4. Warning rather than error is deliberate: the boundary list of a traditional stress package may change per stress period, while the advanced packages that error here have fixed features.

Workflows are in `action_required` (first contribution to this repo), so CI has not run yet. Marking ready so a maintainer can approve the runs.

PR #2989, opened 2026-09-16
fix(hfb): error on multiple barriers on one cell connection

### LINKED ISSUE #1215: Multiple HFBs between same two cells
MODFLOW 6 does not appear to be capable of handling multiple HFBs between the same two connected cells.  Suggest checking to see if there is a straightforward and sensible way to support this.  If not, then MODFLOW should terminate with an error if duplicates are encountered.

KR-Ravindra:
## Problem

MODFLOW 6 accepted two or more HFBs on the same cell connection without complaint, and the HFB package assumes there is at most one. Reported in #1215.

`check_data` resolves each barrier's cell pair to a position in the model `ja` array and stores it in `idxloc(ihfb)`; it errors only when the two cells are not connected. Two barriers on the same pair therefore resolve to the same connection and nothing notices. Three things then go wrong:

- `condsat_modify` reads `condsat(jas(ipos))`, saves it into `csatsav(ihfb)`, and writes the barrier-corrected value back. For the second barrier on a connection, the value it saves as "the original" is the one the first barrier already modified.
- `condsat_reset` restores in `ihfb` order, so the last duplicate wins and `condsat` is never returned to its unmodified value. It drifts further with every stress period that re-reads HFB data.
- The non-Newton branch of `hfb_fc` reads the matrix entry it is about to overwrite (`aterm`) and adjusts the diagonal by `aterm - cond`, so each duplicate subtracts the barrier correction from the diagonal again.

Heads are silently wrong in these models, with no message pointing at the cause.

## Fix

`check_data` now compares each barrier's resolved connection against those of the preceding barriers, and stores an error naming both barrier numbers and both cells when they match, so the simulation terminates during input processing:

```
  1.  HFB no. 1 and HFB no. 2 are both between cells (1,3,2) and (1,3,3).
     Only one HFB can be assigned to a cell connection.
```

The comparison is on `jas(idxloc(...))` rather than on `idxloc` itself. `jas` maps a connection and its transpose to the same symmetric entry, so `(n, m)` and `(m, n)` are recognised as one connection — which is what the downstream code sees, since `condsat_modify`, `condsat_reset` and `hfb_fc` all index the symmetric arrays through `jas`. Comparing raw `idxloc` values would have missed a duplicate entered with the cells reversed. The lookup is an inner loop over the preceding barriers rather than a scratch array indexed by connection position: `nhfb` is small, it matches the loop already in this procedure, and it needs no allocation sized to `nja`.

### Why an error rather than support for multiple barriers

The issue asks for either. Supporting duplicates means deciding how several barriers on one face combine — presumably resistances in series — and that changes the conductance a model computes, affects the XT3D and Newton paths as well as the standard one, and needs its own documentation and test coverage. That is a numerical-method decision that belongs with the maintainers, so this PR implements the error the issue names as the fallback and does not change results for any model that is valid today. If you would rather have the barriers combined in series, I am happy to close this in favour of that.

A duplicate is currently accepted, so this is a new terminating error for input that used to run. It only triggers on input whose results were already wrong in the ways described above.

## How tested

- New `autotest/test_gwf_errors.py::test_hfb_duplicate_connection_error`, which builds a minimal GWF model with a repeated barrier and a second repeat with the cells reversed, and asserts both are reported against `test.hfb`. It passes with the fix and fails without it (mf6 terminates normally instead of erroring).
- The full HFB test set was run to confirm nothing valid started failing: `test_gwf_disu_hfb_idomain.py`, `test_gwf_vhfb.py` (6 cases), `test_gwf_vsc05_hfb.py` (4 cases), `test_gwf_npf_thickstrt.py` (9 cases, 4 with HFB), `test_gwf_npf_vcv.py`, plus `test_gwf_errors.py` — 37 passed. No existing test defines duplicate barriers.
- Broad sweep `pytest test_gwf_*.py -m "not slow"`: 406 passed, 35 skipped. The 6 failures and 1 collection error are all missing optional Python packages in my local environment (`shapely`, `scipy`) and are present without this change.
- `gwf-hfb.f90` compiles clean under a `-Dwerror=true` configuration (Homebrew GCC 16.2.0). The two `-Werror=character-truncation` failures that configuration reports are pre-existing and in other files.
- `fprettify -c .fprettify.yaml`, `ruff check`, `ruff format --check` and `codespell` are clean on the files touched.

## Links

- Closes #1215

- [x] Closed issue #1215

This change was written with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.

Checked against `develop`:

1. `check_data` is the only writer of `idxloc`, and `condsat_reset`, `condsat_modify` and the non-Newton branch of `hfb_fc` all reach `condsat` and the matrix through `jas(idxloc(ihfb))`, so `jas(idxloc(jhfb)) == jas(idxloc(ihfb))` is the right key: it catches the reversed-cell entry and cannot flag two barriers on different connections.
2. `idxloc(ihfb) = 0` at the top of the loop plus the `cycle` after the unconnected-cells error means a barrier that failed the connection check neither carries a stale position into the duplicate loop nor is compared against later (`if (idxloc(jhfb) == 0) cycle`).
3. `hfb_rp` re-runs `source_data` -> `check_data` for every stress period that provides HFB data, so a duplicate introduced in a later period is caught too.
4. The error goes through `store_error_filename(this%input_fname)`, the same path the existing chd/drn tests in `test_gwf_errors.py` assert on, so the `ERROR OCCURRED WHILE READING FILE 'test.hfb'` assertion is stable.

Staying a draft until CI finishes (nothing has failed so far); I will mark it ready then.

KR-Ravindra:
Round 2 self-review. CI is green on all 20 jobs (the two skips are the pages/RTD deploy steps that only run on develop). Nothing changed since round 1, so marking this ready for review.

Issue #2275, opened 2025-04-09
LKBD-COND observation type


emorway-usgs:
In the SFE package, there is an observation type for the amount of heat conductively exchanged with the streambed material (`strmbd-cond`). At the moment, there is no equivalent observation type for LKE.  In the future, it may be useful to add a `lakbed-cond` observation type to LKE.

PR #2990, opened 2026-09-16
feat(lke): add lakebed-cond observation type

KR-Ravindra:
## Problem

The SFE Package has a `strmbd-cond` observation type that reports the amount of
heat conductively exchanged with the streambed material. LKE computes the
equivalent quantity for the lakebed -- it is already a budget term
(`LAKEBED-COND`) and is already written to the LKE budget file -- but it cannot
be observed, so getting it per lake or per lakebed connection means
post-processing the binary budget file. Issue
https://github.com/MODFLOW-ORG/modflow6/issues/2275 suggests adding the missing
observation type.

The issue is worded as a future nice-to-have rather than a request for work
now, so this is offered for whenever you want it. The change is purely
additive: no existing input is interpreted differently and no existing model
result changes.

## Fix

`src/Model/GroundWaterEnergy/gwe-lke.f90`:

- New private term routine `lke_lbcd_term`, with the same argument signature and
  the same budget-term lookup pattern as `sfe_sbcd_term` in `gwe-sfe.f90`: the
  wetted area comes from the auxiliary variable of the `GWF` budget term, the
  thermal conductance is `ktf * wa / s`, and the rate is that conductance times
  the gradient between the groundwater cell and the lake.
- `lakebed-cond` is registered in `lke_df_obs` with the same `ProcessIdPtr`
  (`apt_process_obsID`) as the sibling per-feature observation types, with a
  matching case in `lke_rp_obs` and in `lke_bd_obs`.

The two existing places that compute this arithmetic inline (`lke_fc_expanded`
and `lke_fill_budobj`) are deliberately **not** refactored to call the new
routine. Leaving them untouched keeps the change bit-for-bit identical for every
existing model and confines the diff to new code. Happy to fold them into the
new routine in a follow-up if you would prefer the duplication removed.

Two points worth a reviewer's eye:

1. **Name.** The issue title says `LAKBED-COND`, but the budget-term text in
   `gwe-lke.f90` is `LAKEBED-COND`, the budget index is described as "lakebed
   conduction", and `strmbd-cond` is exactly the lowercased SFE budget-term text
   `STRMBD-COND`. I therefore implemented `lakebed-cond`, which is the spelling
   the code already supports and which reads consistently against
   `strmbd-cond`. Say the word and I will rename it to `lakbed-cond` to match
   the issue.
2. **Observation indexing.** `lke_rp_obs` uses `rp_obs_budterm` on the lakebed
   conduction budget term rather than `rp_obs_byfeature` as SFE does. SFR has
   exactly one groundwater connection per reach, so for SFE the feature number
   and the `GWF` budget-term entry coincide; a lake has many lakebed
   connections, so indexing by feature number would silently return only the
   first connection of the lake and could return a connection belonging to a
   different lake. Using `rp_obs_budterm` makes the observation behave exactly
   like the sibling `lke` observation type in the same package: an integer
   `lakeno` observes one lakebed connection and a boundname sums the conductive
   exchange over all of that lake's lakebed connections. `lke_bd_obs` guards on
   `iboundpak(id1(jj))` for the same reason, matching `lke_fill_budobj` and the
   parent `apt_bd_obs`.

No definition-file or `src/Idm/` work: observation types are registered in code
only, and `strmbd-cond` appears nowhere under `doc/mf6io/mf6ivar/dfn/` or
`src/Idm/`. No source files were added, renamed or removed, so the meson and
Visual Studio project files are unchanged.

Documentation: a row for `lakebed-cond` in `doc/Common/gwe-lkeobs.tex`, mirroring
how `strmbd-cond` is documented in `doc/Common/gwe-sfeobs.tex`, and a `features`
/ `advanced` entry in `doc/ReleaseNotes/develop.toml`.

## How tested

`autotest/test_gwe_lke_conduction.py` is extended to request two of the new
observations -- one by the lake boundname, which sums all of the lake's lakebed
connections, and one by lake number, which returns a single connection -- and to
check them against independent sources rather than merely asserting they are
populated:

- the summed observation is compared with the `LAKEBED-COND` rate in the LKE
  package budget written to the listing file, including the in/out direction;
- the single-connection observation is compared with a value computed in the
  test from the wetted lakebed area written to the LAK Package budget file, the
  thermal conductivity and thickness of the lakebed material, and the simulated
  lake and groundwater temperatures.

Results on macOS with Homebrew GCC 16.2.0, meson build, flopy from git:

- `pytest test_gwe_lke_conduction.py` -> 3 passed (all three scenarios). The
  externally computed single-connection value agrees with the observation to the
  full precision of the observation output.
- With the `gwe-lke.f90` change reverted and the binary rebuilt, all three
  scenarios fail with `Observation type not found: LAKEBED-COND`. With the
  conductance in the new routine deliberately doubled instead, the value
  assertions fail, confirming they check values and not just presence.
- All 28 `autotest/test_gwe*.py` files, `-m "not slow"` -> 51 passed, so the new
  observation type does not disturb existing GWE models.
- `fprettify -c .fprettify.yaml` reports no changes for `gwe-lke.f90`; every
  added line is within the 82-column limit. `ruff check` and
  `ruff format --check` pass on the test. `codespell` passes on all four files.
- The file compiles clean under `-Dwerror=true`. The only `-Werror` failures in
  that configuration are the two pre-existing `-Werror=character-truncation`
  errors in `gwf-mvr.f90` and `utils/zonebudget/src/zoneoutput.f90`, which are
  unrelated to this change.
- `doc/ReleaseNotes/mk_releasenotes.py` renders the new entry without warnings.
  I could not build the LaTeX documents (no TeX distribution available here), so
  the new table row was verified by inspection against the surrounding rows: it
  has the same five columns, the previously-last row now ends with `\\`, and the
  new last row omits it, matching `gwe-sfeobs.tex` and the `\input{...} \\` in
  `doc/Common/gwe-obstypetable.tex`.

- [x] Closed issue #2275

This change was written with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.

Checked against `develop`:

1. `lke_find_lak_package` sets `idxbudlbcd` to the same index as `idxbudgwf` (the `GWF` case), so the `LAKEBED-COND` list and the `GWF` list are one list. Indexing the `GWF` aux variable with `ientry` in `lke_lbcd_term` is therefore consistent with what `lke_fill_budobj` already does for this term.
2. `rp_obs_budterm` on that term together with `apt_process_obsID` (which fixes `NodeNumber2` to 1) is exactly how the existing `lke` observation type is wired (`StoreObsType('lke', ...)` in `lke_df_obs` and the `LKE` case of `apt_rp_obs`), so an integer ID selects the first connection of the lake and a boundname sums all of its connections, as the new doc row states.
3. `iboundpak` is dimensioned by lake count while `jj` is a connection index, so the `iboundpak(id1(jj))` guard is required; the `LKE` case of the parent `apt_bd_obs` guards the same way.
4. Moving the trailing `\\` from the previous last row to the new one matches `gwe-obstypetable.tex`, which appends `\\` after the `\input`.
5. On the name: `lakebed-cond` matches the budget text `LAKEBED-COND`, as `strmbd-cond` matches `STRMBD-COND`; I read the issue's `lakbed-cond` as a typo but will rename if preferred.

Staying a draft until the workflows are approved and CI has run.

